PR #17, opened 2017-06-08
Add socket commands to reconfigure haproxy without reloading

aiharos:
This adds reconfiguring haproxy backend servers through stats socket without needing a reload.

The server "slots" for adding new endpoints are pre-allocated in the configuration file. Running out of slots or having too many empty ones does trigger a reload, and the increment by which the slots are added or removed is configurable through configmap.

Any other change in configuration behaves as before and triggers a reload.


jcmoraisjr:
Thanks for supporting this ingress controller!

A few tweaks to your PR:

* People might want the old behaviour. Add a bool option (default to false) eg `backend-slots` to turn on and off this strategy
* Leave OnUpdate small and move the code and dependencies to a new file. Package controller is fine
* Do you need a parameter to the stats socket? If so leave the default as /var/run/haproxy-stats.sock, instead this could be the hardcoded value.
* Use Server instead of Srv in order to avoid ambiguity with Service
* Make golint happy and remove the `_` here: `for k, _ := range toRemoveEndpoints {`

aiharos:
Thanks for the comments, glad to be able to help! Reducing reloads should be a good quality-of-life improvement.

I've added an option "dynamic-scaling" that can be used to get back to previous behavior of reloading after every OnUpdate() (still through the added code-path, just without issuing socket commands and addition of empty server slots).

The stats socket location doesn't need to be a configurable config-map parameter, but it makes sense to have it defined in a single place, to prevent template and code going out of sync.

pkg/controller/stats_socket.go
```diff
@@ -0,0 +1,189 @@
...
+		updatedConfig.BackendSlots = map[string]types.HAProxyBackendSlots{}
+
+		for _, backendName := range updKeys {
+			newBackend := types.HAProxyBackendSlots{}
+			newBackend.FullSlots = map[string]types.HAProxyBackendSlot{}
+			for i, endpoint := range updBackendsMap[backendName].Endpoints {
+				newBackend.FullSlots[fmt.Sprintf("%s:%s", endpoint.Address, endpoint.Port)] = types.HAProxyBackendSlot{
+					fmt.Sprintf("server%04d", i),
+					&endpoint,
+				}
```
aiharos: It is possible to use `addr:port` if `DynamicScaling` is false, I'll change to that. You are correct that it's not feasible if `DynamicScaling` is enabled.

pkg/controller/stats_socket.go
```diff
@@ -0,0 +1,189 @@
...
+		_, ok := s2[k]
+		if !ok {
+			s3[k] = v
+		}
+	}
+	return s3
+}
+
+// populate template variables and optionally issue socket commands to reconfigure haproxy without reloading
+func reconfigureBackends(haproxy *HAProxyController, updatedConfig *types.ControllerConfig) {
```
aiharos: I'll give it a go. ;)

pkg/controller/controller.go
```diff
@@ -127,30 +128,47 @@ func (haproxy *HAProxyController) BackendDefaults() defaults.Backend {
...
+// Reload the backend if the configuration has changed
 func (haproxy *HAProxyController) Reload(data []byte) ([]byte, bool, error) {
-	if !haproxy.configChanged(data) {
-		return nil, false, nil
-	}
+	/*
+		if !haproxy.configChanged(data) {
+			return nil, false, nil
+		}
+	*/
```
aiharos: Indeed, I'll work in a `DeepEqual` even when `DynamicScaling` is false.

pkg/controller/controller.go
```diff
@@ -127,30 +128,47 @@ func (haproxy *HAProxyController) BackendDefaults() defaults.Backend {
 
 // OnUpdate regenerate the configuration file of the backend
 func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) ([]byte, error) {
-	data, err := haproxy.template.execute(newControllerConfig(&cfg, haproxy))
+	updatedConfig := newControllerConfig(&cfg, haproxy)
+
+	reconfigureBackends(haproxy, updatedConfig)
+
+	data, err := haproxy.template.execute(updatedConfig)
```
aiharos: There is a benefit to generating the configuration file even if `reloadRequired` is false - having the configuration in sync with in-memory configuration makes it easy to check what is currently active by just inspecting haproxy.cfg - so I would advise to keep this.

pkg/controller/controller.go
```diff
@@ -127,30 +128,47 @@ func (haproxy *HAProxyController) BackendDefaults() defaults.Backend {
 
 // OnUpdate regenerate the configuration file of the backend
 func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) ([]byte, error) {
-	data, err := haproxy.template.execute(newControllerConfig(&cfg, haproxy))
+	updatedConfig := newControllerConfig(&cfg, haproxy)
+
+	reconfigureBackends(haproxy, updatedConfig)
```
aiharos: I was on the fence about this too - I'll switch to the less coupled variant.

pkg/controller/config.go
```diff
@@ -75,25 +75,28 @@ func newHAProxyConfig(haproxyController *HAProxyController) *types.HAProxyConfig
 			DefaultMaxSize: 1024,
```
aiharos: Absolutely!

pkg/controller/config.go
```diff
@@ -75,25 +75,28 @@ func newHAProxyConfig(haproxyController *HAProxyController) *types.HAProxyConfig
...
+		BackendCheckInterval:        "2s",
+		Forwardfor:                  "add",
+		MaxConn:                     2000,
+		HSTS:                        true,
+		HSTSMaxAge:                  "15768000",
+		HSTSIncludeSubdomains:       false,
+		HSTSPreload:                 false,
+		StatsPort:                   1936,
+		StatsAuth:                   "",
+		DynamicScaling:              true,
```
aiharos: Alright

pkg/controller/controller.go
```diff
@@ -127,30 +128,47 @@ func (haproxy *HAProxyController) BackendDefaults() defaults.Backend {
...
 	err := ioutil.WriteFile(haproxy.configFile, data, 0644)
 	if err != nil {
 		return nil, false, err
 	}
+
+	if !haproxy.ReloadRequired {
+		glog.Infoln("reload not required")
+		return nil, false, nil
+	}
+	glog.Infoln("reload is required")
```
aiharos: Noted

pkg/controller/controller.go
```diff
@@ -127,30 +128,47 @@ func (haproxy *HAProxyController) BackendDefaults() defaults.Backend {
...
 	// TODO missing HAProxy validation before overwrite and try to reload
 	err := ioutil.WriteFile(haproxy.configFile, data, 0644)
 	if err != nil {
 		return nil, false, err
 	}
+
+	if !haproxy.ReloadRequired {
+		glog.Infoln("reload not required")
+		return nil, false, nil
+	}
```
aiharos: I'll see about making this more informative

pkg/controller/controller.go
jcmoraisjr: Just to let you know Ingress now implement it's own equals before call `OnUpdate` and `Reload`  doesn't exist anymore. I can update Ingress right now (you'd need to merge or rebase) or I can update Ingress on top of your code and remove the `configChanged` part afterwards. I'd follow the latter but this definitely is your choice: update ingress now or on top of your changes?

aiharos:
The addition of builtin Ingress equality before invoking `OnUpdate` seems like it makes things a bit easier in that we no longer need to worry about reloading when nothing changed and DynamicScaling is disabled, so the commits don't add `DeepEqual` for that case. 

In removing `Reload` we can move the code from it into `OnUpdate`, change `haproxy.reloadRequired` into just a local variable within `OnUpdate` and keep generating haproxy.cfg (since due to the change it should only be generated when the Ingress has changed).

Hopefully I've addressed the rest of the comments and haven't forgotten something.

pkg/controller/controller.go
aiharos: I agree with the latter, hope that the set of updates based on your comments makes it straightforward.

jcmoraisjr:
Thanks!

jcmoraisjr:
@aiharos FYI just updated the readme file with [dynamic-scaling](https://github.com/jcmoraisjr/haproxy-ingress#dynamic-scaling) config. Please send a new PR if you want to change something.

aiharos:
I forgot about the docs, thanks for updating - looks good!

PR #19, opened 2017-06-15
Replace bash-only substring expansion

aiharos:
/bin/sh points to bash in the alpine base image, but is usually a symlink to dash on other distros. This replaces bash-only substring expansion with a dash-compatible version.

### DIFF
--- a/rootfs/start.sh
+++ b/rootfs/start.sh
@@ -17,7 +17,7 @@
 set -e
 
 init() {
-    if [ $# -gt 0 ] && [ "$(echo $1 | awk '{s=substr($0, 1, 2); print s}')" != "--" ]; then
+    if [ $# -gt 0 ] && [ "$(echo $1 | cut -b1-2)" != "--" ]; then
         exec "$@"
         exit 0
     fi
--- a/rootfs/start.sh
+++ b/rootfs/start.sh
@@ -17,7 +17,7 @@
 set -e
 
 init() {
-    if [ $# -gt 0 ] && [ "${1:0:2}" != "--" ]; then
+    if [ $# -gt 0 ] && [ "$(echo $1 | awk '{s=substr($0, 1, 2); print s}')" != "--" ]; then
         exec "$@"
         exit 0
     fi


jcmoraisjr:
Alpine's sh symlink to BusyBox, which iirc implements ash. The default sh apparently does implement this type of parameter expansion. I prefer supported expansions in order to avoid forks. I'd also use eg `sed -nr 's/.../\1/p' <<< "$*"` instead of `echo "$*" | sed -nr 's/.../\1/p'` but ash apparently doesn't support this type of redirection. Bash btw isn't installed.

Did you have any problem with the actual implementation?

aiharos:
The problem arises on a base distro image not using busybox, i.e. Debian/Ubuntu, RedHat, Arch, etc. Since `/bin/sh` is usually dash or bash in Bourne shell compatibility mode for most distros, `#!/bin/sh` scripts ordinarily don't contain supersets of the base POSIX (Bourne shell) features, and this code results in a following surprise 
`./start.sh: 20: ./start.sh: Bad substitution`

A way to retain the original expansion code in a more portable way could be to swap the `#!/bin/sh` with an explicit shell that supports running the code. `#!/bin/bash` would likely be better suited than ash, as ash is usually not found outside of busybox.

Since as you mentioned bash isn't installed in Alpine by default and since the Bourne feature set is POSIX (reliably consistent everywhere) I proposed to change the expansion code instead.

jcmoraisjr:
This is the way to go if this is a generic script the user run on it's own environment. But this `start.sh` script is tightly coupled with the Alpine container because I wouldn't taking into account run it elsewhere.

How did you find this issue? How do you use this script in your devel environment?

aiharos:
In many cases debugging is more comfortable for me in a Debian environment so I reached for a non-Alpine base image while testing some ideas and different haproxy versions in conjunction with the controller. I can also imagine cases of people preferring a different base image for policy reasons.

It is not a difficult patch to keep stashed for such occasions but I believe the change to upstream is little or no impact - and it also makes for a more harmonious reading of code with that `#!/bin/sh` up top.

I've had bash constructs inadvertently seep into my sh scripts myself at times so I thought this was such an occasion, but in the end it's your discretion if this warrants merging.

aiharos:
This should do it then.

jcmoraisjr:
Thanks!

PR #22, opened 2017-06-28
verify new configuration is valid before replacing

aiharos:
This works for native reload strategy. For multibinder reload strategy all configuration files are now kept in sync but validation is not performed as it cannot pass in a standalone haproxy process due to file-descriptor use.

As always, open for discussion.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -78,8 +78,9 @@ func checkValidity(configFile string) error {
 
 // multibinderERBOnly generates a config file from ERB template by invoking multibinder-haproxy-erb
 func multibinderERBOnly(configFile string) (string, error) {
-	_, err := exec.Command("multibinder-haproxy-erb", "/usr/local/sbin/haproxy", "-f", configFile, "--erb-write-only").CombinedOutput()
+	out, err := exec.Command("multibinder-haproxy-erb", "/usr/local/sbin/haproxy", "-f", configFile, "-c", "-q").CombinedOutput()
 	if err != nil {
+		glog.Warningf("Error validating config file:\n%v", string(out))
 		return "", err
 	}
 	return configFile[:strings.LastIndex(configFile, ".erb")], nil
@@ -96,13 +97,11 @@ func RewriteConfigFiles(data []byte, reloadStrategy, configFile string) error {
 	}
 
 	if reloadStrategy == "multibinder" {
-		unverified, err := multibinderERBOnly(tmpf)
+		generated, err := multibinderERBOnly(tmpf)
 		if err != nil {
-			glog.Warningln("Error generating file contents from ERB template")
 			return err
 		}
-		// multibinder haproxy configuration file will not pass validation due to file descriptors so it is skipped
-		err = os.Rename(unverified, "/etc/haproxy/haproxy.cfg")
+		err = os.Rename(generated, "/etc/haproxy/haproxy.cfg")
 		if err != nil {
 			glog.Warningln("Error updating config file")
 			return err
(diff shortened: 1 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -61,3 +65,59 @@ func SendToSocket(socket string, command string) error {
...
+	if err != nil {
+		glog.Warningf("Error validating config file:\n%v", string(out))
+		return err
+	}
+	return nil
+}
+
+// multibinderERBOnly generates a config file from ERB template by invoking multibinder-haproxy-erb
+func multibinderERBOnly(configFile string) (string, error) {
+	_, err := exec.Command("multibinder-haproxy-erb", "/usr/local/sbin/haproxy", "-f", configFile, "--erb-write-only").CombinedOutput()
```
jcmoraisjr: It's safe use `-c` here and remove `--erb-write-only`. This will call HAProxy with a valid .cfg file. Changing this and the output will be useful. Just follow the same approach used in `checkValidity`.

pkg/utils/utils.go
```diff
@@ -61,3 +65,59 @@ func SendToSocket(socket string, command string) error {
...
+		return err
+	}
+
+	if reloadStrategy == "multibinder" {
+		unverified, err := multibinderERBOnly(tmpf)
+		if err != nil {
+			glog.Warningln("Error generating file contents from ERB template")
+			return err
+		}
+		// multibinder haproxy configuration file will not pass validation due to file descriptors so it is skipped
```
jcmoraisjr: If using `-c`, this comment won't be true anymore.

pkg/utils/utils.go
```diff
@@ -61,3 +65,59 @@ func SendToSocket(socket string, command string) error {
...
+	_, err := exec.Command("multibinder-haproxy-erb", "/usr/local/sbin/haproxy", "-f", configFile, "--erb-write-only").CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	return configFile[:strings.LastIndex(configFile, ".erb")], nil
+}
+
+// RewriteConfigFiles safely replaces configuration files with new contents after validation
+func RewriteConfigFiles(data []byte, reloadStrategy, configFile string) error {
+	tmpf := "/etc/haproxy/new_cfg.erb"
```
jcmoraisjr: I'd suggest using `os.TempDir() + ...` instead.

aiharos: I was considering this, but placing the temporary file in the same directory as the target file ensures the `os.Rename` will perform on the same filesystem and be atomic so that seemed like a safer bet.

I also considered using `ioutil.TempFile` as a safest option, but that seems a hassle as multibinder expects a filename ending with `.erb` and this is not a feature of `ioutil.TempFile` yet. 

Actually we could use `ioutil.TempDir` created within `/etc/haproxy/` to ensure the same filesystem and no interference with the new file even when the temporary filename itself is hardcoded, at the cost of having to clean up the temporary directory.

Let me know it that sounds good or you prefer something else.

pkg/utils/utils.go
aiharos: I was positive I tested this and it didn't work as expected, but now after retesting it indeed seems fine. Fixed.

pkg/utils/utils.go
jcmoraisjr: I was taking into account that, at this moment, `/` is the only filesystem. But we cannot say that this will stay the same about the tmp subdir. Since this is more a convention than an issue the actual implementation is fine.

jcmoraisjr:
Thanks!

Yantrio:
@jcmoraisjr Hey, can we get a docker image pushed of this to the canary tag?

jcmoraisjr:
@Yantrio `canary` tag can only be updated when starting the `v0.4` release cycle, where this implementation belongs to. I changed the rules a bit and introduced the `snapshot` tag which can be updated during the development cycle. I've just updated this tag with the current implementation.

Issue #21, opened 2017-06-20
Implement health check


jcmoraisjr:
Implement an URL used for health check

woopstar:
Wouldn't it be as simple, as to implement this in the haproxy.cfg

```
# Use /healthz to report haproxy's status
frontend healthz
    mode http
    monitor-uri /healthz
```

According to the documentation at http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4-monitor-uri:

`is the exact URI which we want to intercept to return HAProxy's health status instead of forwarding the request.`

jcmoraisjr:
Sure, haproxy.cfg is probably the way to go. Implement here means configure, perhaps listening another port and taking care with port conflict on deployments where hostNetwork is true, and also document and update the deployment instructions.

woopstar:
Well, you cannot use the health check if you run with hostNetwork to true or am I wrong?

jcmoraisjr:
> Well, you cannot use the health check if you run with hostNetwork to true or am I wrong?

It works, the difference is that the IP of the pod - which is the default value of `probe/httpGet/host` - is the IP of the host instead of an IP of the overlay network.

jcmoraisjr:
Implemented on 9260339 .

@westse FYI `accept-proxy` was added on stats page. The HAProxy's health check URI defaults to `:10253/healthz`.

westse:
@jcmoraisjr, I just realized `/healthz` works for health checking, but the stats page still needs `accept-proxy` to be omitted for internal metrics gathering. For example, I plan to use [haproxy_exporter](https://github.com/prometheus/haproxy_exporter) in the same pod as haproxy-ingress in order to get metrics into Prometheus.

Options I see:
1. Remove `accept-proxy` from stats page
1. In addition to `add-proxy-protocol`, add another property (e.g. `use-proxy-protoco-for-stats`) for independently setting `accept-proxy` on the stats page

Which do you prefer, or do you see another option? I can produce a PR if you would like. Also, do you want an issue created?

Thanks

jcmoraisjr:
@westse thanks for pointing this out, I'll also use HAProxy with Prometheus and will probably have the same issue.

A PR would be handy. Use `stats-proxy-protocol` property.

PR #23, opened 2017-07-10
Add use-proxy-protocol option

westse:
This enables external load balancers, e.g. AWS ELBs, to use the proxy protocol on requests to haproxy-ingress.

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -71,7 +71,7 @@ backend {{ $backend.Name }}
 ###### HTTP frontend
 ######
 frontend httpfront
-    bind *:80
+    bind *:80{{ if $cfg.UseProxyProtocol }} accept-proxy{{ end }}
     mode http
 {{ if ne $cfg.Syslog "" }}
     option httplog
@@ -127,7 +127,7 @@ frontend httpfront
 ###### HTTPS frontend (tcp mode)
 ######
 frontend httpsfront
-    bind *:443
+    bind *:443{{ if $cfg.UseProxyProtocol }} accept-proxy{{ end }}
     mode tcp
     tcp-request inspect-delay 5s
     tcp-request content accept if { req.ssl_hello_type 1 }
--- a/rootfs/start.sh
+++ b/rootfs/start.sh
@@ -59,7 +59,9 @@ listen main
 EOF
 
     # Add erb code to a new template file
-    sed "/^    bind \+\*\?:/s/\*\?:\(.*\)/<%= bind_tcp('0.0.0.0', \1) %>/" \
+    # - [0-9]* will match actual ports, for example 443 in bind *:443
+    # - {{[^}]*}} will match when the port is templatized, for example {{ $cfg.StatsPort }} in bind *:{{ $cfg.StatsPort }}
+    sed "/^    bind \+\*\?:/s/\*\?:\(\([0-9]*\)\|\({{[^}]*}}\)\)/<%= bind_tcp('0.0.0.0', \1) %>/" \
         "$TEMPLATE" > "${CONFIG}.tmpl"
 }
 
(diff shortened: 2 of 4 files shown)


rootfs/start.sh
```diff
@@ -59,7 +59,7 @@ listen main
 EOF
 
     # Add erb code to a new template file
-    sed "/^    bind \+\*\?:/s/\*\?:\(.*\)/<%= bind_tcp('0.0.0.0', \1) %>/" \
+    sed "/^    bind \+\*\?:/s/\*\?:\(\([0-9]*\)\|\({{[^}]*}}\)\)/<%= bind_tcp('0.0.0.0', \1) %>/" \
```
westse: This refactor was needed ensure that only the port was included in the `\1` regex capture group . Notice the two cases:
- `[0-9]*` will match actual ports, for example `443` in `bind *:443`
- `{{[^}]}}` will match when the port is templatized, for example `{{ $cfg.StatsPort }}` in `bind *:{{ $cfg.StatsPort }}`

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -127,7 +127,7 @@ frontend httpfront
 ###### HTTPS frontend (tcp mode)
 ######
 frontend httpsfront
-    bind *:443
+    bind *:443{{ if $cfg.UseProxyProtocol }} accept-proxy{{ end }}
```
jcmoraisjr: What about use this on the stats port?

rootfs/start.sh
jcmoraisjr: Add these two items (only the items is fine) above the `sed`.

rootfs/etc/haproxy/template/haproxy.tmpl
westse: Actually omitting from stats port was intentional because I plan to perhaps use it for readiness and liveness checks, which would break because Kubernetes doesn't send the proxy protocol. What are your thoughts?

rootfs/start.sh
westse: Do you mean I should break it out into a separate shell variable like the following?
```
bindport='\(\([0-9]*\)\|\({{[^}]*}}\)\)'
sed "/^    bind \+\*\?:/s/\*\?:$bindport/<%= bind_tcp('0.0.0.0', \1) %>/" \
```

jcmoraisjr: No, add the explanation as a comment above the `sed`! Something like:

```
# Add erb code to a new template file
# - [0-9]* will match actual ports, for example 443 in bind *:443
# - {{[^}]*}} will match when the port is templatized, for example {{ $cfg.StatsPort }} in bind *:{{ $cfg.StatsPort }}
sed ...
```

rootfs/etc/haproxy/template/haproxy.tmpl
jcmoraisjr: I'd like to see the stats page working if the user for any reason want to expose it. Planning to include a proper URL to be used for health-check as described in #21

westse: Would it makes sense to have an two stats rules (one for external use, and one for internal use) so they can be configured independently? That way the external one can use `accept-proxy`, while the internal one can omit it for Kubernetes readiness/liveness checks.

Anyway, it sounds like that will be part of issue #21. Are you comfortable with the changes in this PR? (The comments to start.sh have been added)

Thanks for the feedback!

jcmoraisjr: Do you see any problem on your deployment if `accept-proxy` is added to the stats binding after the implementation of a proper health-check url as described on #21 ? I have only bare metal deployments, won't use this feature myself but want to see a consistent configuration.

westse: No there should be no problem with adding `accept-proxy` to the stats binding as long as it is in addition to the proposed monitor-uri binding from #21 and the latter does not have `accept-proxy` (so Kubernetes can call it without the proxy protocol).

Thanks for the merge!

PR #24, opened 2017-07-11
Prevent 404s if Host header includes port 80

westse:
When ingress hostname does not include a port, ensure proxy config exists for when Host header includes either "{host}"" or ""{host}:80". Needed because while some clients will strip ":80", some clients do not. Without this change requests with "Host: {host}:80" would not match, causing HAProxy to return 404s.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -17,11 +17,13 @@ limitations under the License.
 package types
 
 import (
+	"regexp"
 	"k8s.io/ingress/core/pkg/ingress"
 	"k8s.io/ingress/core/pkg/ingress/annotations/authtls"
 	"k8s.io/ingress/core/pkg/ingress/annotations/proxy"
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Nice catch, thanks. Some thoughts about the changes:

* `sni` doesn't have this problem so only the `HTTP` block need to be changed
* A new `range $ing.HTTPServers` with ACLs sounds a better approach, avoiding the slice of hostnames and the loops - what about that?

westse:
Sure I'll take a look at both of those - thanks for the suggestions (will be a few days while I am out).

I am new to haproxy config so I'm not clear on either of those, so any pointers or code snippets you can provide will be helpful.

jcmoraisjr:
I made a small PoC in order to check the syntax and so... it's done and pushed to master. I'm closing this PR but feel free to add new comments if you find something that differs from what you were expecting.

westse:
@jcmoraisjr , looks to be working. Thanks for taking care of it!

PR #26, opened 2017-07-18
stats page proxy protocol config separate from general proxy protocol…

westse:
… config

@jcmoraisjr


jcmoraisjr:
Thanks! I'll update `snapshot` shortly.

PR #27, opened 2017-07-19
fix reloadRequired potentially not set correctly

aiharos:
This fixes a corner-case when if only some but not all socket commands fail, the successful ones would overwrite the previous value of `reloadRequired`.


jcmoraisjr:
Thanks!

PR #31, opened 2017-07-31
add support for wildcard hostnames

from-nibly:
I added support for using a regex to allow *. wildcard hostnames.  This also escapes characters that are not allowed in ACL rules to an `_`. 

An example ishost rule will look like this: `ishost-_.example.com hdr_reg(host) ^([^\.]+)\.example.com`

### DIFF
--- a/pkg/controller/template.go
+++ b/pkg/controller/template.go
@@ -23,6 +23,7 @@ import (
 	"github.com/jcmoraisjr/haproxy-ingress/pkg/utils"
 	"os/exec"
 	gotemplate "text/template"
+	"regexp"
 )
 
 type template struct {
@@ -38,6 +39,15 @@ var funcMap = gotemplate.FuncMap{
 		}
 		glog.Error("invalid type conversion on backendHash template function")
 		return ""
+	},
+	"hostnameRegex": func(hostname string) string {
+		rtn := regexp.MustCompile(`\.`).ReplaceAllLiteralString(hostname, "\\.")
+		rtn = regexp.MustCompile(`\*`).ReplaceAllLiteralString(rtn, "([^\\.]+)")
+		return "^" + rtn
+	},
+	"labelize": func (identifier string) string {
+		re := regexp.MustCompile(`[^a-zA-Z0-9:_\-.]`)
+		return re.ReplaceAllLiteralString(identifier, "_")
 	},
 }
 
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -79,24 +79,24 @@ frontend httpfront
     mode http
 {{ if ne $cfg.Syslog "" }}
     option httplog
-    {{ if ne $cfg.HTTPLogFormat "" }}
+{{ if ne $cfg.HTTPLogFormat "" }}
     log-format {{ $cfg.HTTPLogFormat }}
-    {{ end }}
+{{ end }}
 {{ end }}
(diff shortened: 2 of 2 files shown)


from-nibly:
@jcmoraisjr @westse

from-nibly:
not sure how to add the missing dependency k8s.io/kubernetes/pkg/util/validation

jcmoraisjr:
Thanks for this addition. Some general tips:

* Perhaps there is no need to use a `SafeHostname` field/func, just populate the `Hostname` itself with `Hostname[1:]` if it's a wildcard domain. What you think?
* No need to use a `EscapedHostname` - the updated `Hostname` will only have valid chars to use as identifier name and also to match domains
* I think you should use `-m reg` before `-i` on `req.ssl_sni` fetch
* Unindent all `{{commands}}` in the template
* Import from  `"k8s.io/apimachinery/pkg/util/validation"` if you want to use the k8s validation
* Move `IsWildcardHostname` to template/funcMap or even as another `HAProxyServer` field - leave the types package without rules
* Run `go vet ./pkg/...` and `golint ./pkg/...` in order to see and fix some minor issues in the code

from-nibly:
we need to keep the *. on the hostname because we need to support both *.example.com and example.com as two separate routing rules, with two different destinations, at the same time. 
The SafeHostname/EscapedHostname allows us to do that.  Unless I'm not understanding what you are saying.
I can create all of those functions as additional template functions instead of functions inside types.go

jcmoraisjr:
I meant use `.example.com` as the hostname, which is: a valid HAProxy identifier, a valid matching suffix and an invalid domain name which will avoid ambiguity.

from-nibly:
I'm not sure how to unmarshal the Hostname field into .example.com from *.example.com

Also, I just realized that .example.com is not a valid suffix because we need to escape the dots to prevent `example-com` from matching it.
I propose two methods

hostnameRegex(hostname string) which takes a hostname and returns the regex that will match on that hostname correctly.
This can be used in place of special casing wildcard hostnames in the template. We would convert all hostnames wildcard or not to a regex string. `*.example.com -> ^([^\.]+)\.example\.com` and `example.com -> ^example\.com` so it will work in all cases and with all hostnames.

and labelize(identifier string) which takes a string and makes it a valid haproxy label.

from-nibly:
This solution also allows us to implement wildcard hostnames in requests that will be terminated at the service instead of haproxy.

from-nibly:
I have updated the pr with the proposed changes.

jcmoraisjr:
Nice approach. Only worried about the small, although real, performance penalty of matching regex when a simple string match would be enough. I'm aware that the template will become (even more) messy, otoh this is on behalf of a better cfg file.

pkg/controller/template.go
```diff
@@ -39,6 +40,15 @@ var funcMap = gotemplate.FuncMap{
 		glog.Error("invalid type conversion on backendHash template function")
 		return ""
 	},
+	"hostnameRegex": func(hostname string) string {
+		rtn := regexp.MustCompile(`\.`).ReplaceAllLiteralString(hostname, "\\.")
+		rtn = regexp.MustCompile(`\*`).ReplaceAllLiteralString(rtn, "([^\\.]+)")
+		return "^" + rtn
```
jcmoraisjr: @from-nibly I'd change this to `"^" + rtn + "$"` so this doesn't match other hostnames that starts with this regex. Eg `*.xpto.com` would match `sub.xpto.com.br`. Problem with this change?

from-nibly:
No that sounds great!

PR #32, opened 2017-08-02
Wildcard conditionals

from-nibly:
@jcmoraisjr Here is an alternative solution to https://github.com/jcmoraisjr/haproxy-ingress/pull/31 that uses the conditionals.

### DIFF
--- a/pkg/controller/template.go
+++ b/pkg/controller/template.go
@@ -49,6 +49,9 @@ var funcMap = gotemplate.FuncMap{
 		re := regexp.MustCompile(`[^a-zA-Z0-9:_\-.]`)
 		return re.ReplaceAllLiteralString(identifier, "_")
 	},
+	"isWildcardHostname": func(identifier string) bool {
+		return regexp.MustCompile(`^\*\.`).MatchString(identifier)
+	},
 }
 
 func newTemplate(name string, file string) *template {
--- a/pkg/controller/template.go
+++ b/pkg/controller/template.go
@@ -23,6 +23,7 @@ import (
 	"github.com/jcmoraisjr/haproxy-ingress/pkg/utils"
 	"os/exec"
 	gotemplate "text/template"
+	"regexp"
 )
 
 type template struct {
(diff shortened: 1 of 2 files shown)


pkg/controller/template.go
```diff
@@ -39,6 +40,18 @@ var funcMap = gotemplate.FuncMap{
...
+		rtn := regexp.MustCompile(`\.`).ReplaceAllLiteralString(hostname, "\\.")
+		rtn = regexp.MustCompile(`\*`).ReplaceAllLiteralString(rtn, "([^\\.]+)")
+		return "^" + rtn
+	},
+	"labelize": func (identifier string) string {
+		re := regexp.MustCompile(`[^a-zA-Z0-9:_\-.]`)
+		return re.ReplaceAllLiteralString(identifier, "_")
+	},
+	"isWildcardHostname": func(identifier string) bool {
+		return !!regexp.MustCompile(`^\*\.`).MatchString(identifier)
```
jcmoraisjr: Remove the double `not` (`!!`)

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -79,24 +79,28 @@ frontend httpfront
...
 {{ if ne $cfg.Syslog "" }}
     option httplog
-    {{ if ne $cfg.HTTPLogFormat "" }}
+{{ if ne $cfg.HTTPLogFormat "" }}
     log-format {{ $cfg.HTTPLogFormat }}
-    {{ end }}
 {{ end }}
-{{ range $server := $ing.Servers }}
+{{ end }}
+{{ range $server := $ing.HTTPServers }}
```
jcmoraisjr: This must be `$ing.Servers` otherwise ssl redirect will fail.

pkg/controller/template.go
from-nibly: oh huh for some reason i thought it wasn't coercing without that.

from-nibly:
@jcmoraisjr those changes have been made

jcmoraisjr:
Thanks!

PR #33, opened 2017-08-04
503 server to respond to 127.0.0.1:8181 backends

from-nibly:
create a 503 frontend/backend to respond to the default backends that are sent by the ingress library.

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -263,6 +263,11 @@ backend error413
     errorfile 400 /usr/local/etc/haproxy/errors/413.http
     http-request deny deny_status 400
 
+listen error503noendpoints
+    bind *:8181
+    mode http
+    errorfile 503 /usr/local/etc/haproxy/errors/503noendpoints.http
+
 ######
 ###### Stats page
 ######
--- a/rootfs/usr/local/etc/haproxy/errors/503noendpoints.http
+++ b/rootfs/usr/local/etc/haproxy/errors/503noendpoints.http
@@ -1,0 +1,8 @@
+HTTP/1.0 503 Service Unavailable
+Cache-Control: no-cache
+Connection: close
+Content-Type: text/html
+
+<html><body><h1>503 Service Unavailable</h1>
+No endpoints found
+</body></html>


jcmoraisjr:
Sorry the delay. What's the purpose of this PR?

from-nibly:
The ingress library returns a bunch of `Backend`s that have a remote host and port of `127.0.0.1:8181`. This comes from `nginx-ingress-controllers` implementation of having a server that listens on that port and only returns 503s. So when there is a kubernetes service with no `address`s in the `Endpoints` object it sets the upstream to the 503 server and nginx returns a 503. Currently (without this pr) the `haproxy-ingress` project still sends requests to `127.0.0.1:8181` but gets a connection refused. This actually ends up with a response with status 503 but when logged shows that it got a connection refused. Adding this server will define a distinction between not having any `address`s on the `Endpoints` and getting a connection refused for some other reason.

jcmoraisjr:
Sure, logging, makes sense. A few suggestions:

* Create just a `listen` section below the `error413`
* What about use a `http-request deny deny_status 503` instead of using an external file?

from-nibly:
The advantage of having the file is that it allows us to post a reason in the html. Is there a way to do that with the inline solution?

jcmoraisjr:
I wasn't taking into account a reason to the end user. Do you have such requirement? Internal logs is another matter and should be fixed.

from-nibly:
we are using this for testing so 503s because a service isn't up yet can happen all the time. it would be nice if we could let the user know that is the reason.

jcmoraisjr:
Lets give it a try. Another few changes:

* Move to `/usr/local/etc/haproxy/errors/` where all HAProxy stuff (except template and cfg) is located
* Rename the `listen` section to `error503noendpoints` in order to mimic the http filename
* Add a line break at the end of the http file and convert it to CRLF as suggested in the HAProxy [docs](http://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4-errorfile)

from-nibly:
@jcmoraisjr changes have been made

jcmoraisjr:
Missing change to the new path in the template.

from-nibly:
whoops. right you are. Changed and tested

jcmoraisjr:
Thanks

PR #39, opened 2017-08-30
Set graceful docker stopsignal

bcorijn:
Kubernetes respects the STOPSIGNAL from the dockerfile when terminating pods. Setting the stopsignal to the gracefulstop instead of hard stop, allows for rolling updates to the HAProxy pods themselves with zero downtime (when not using Daemonsets or Hostports).


bcorijn:
Nvm, realized this is not needed with how the application works. Quite the brainfart on my side...

Issue #38, opened 2017-08-30
ELB SSL termination with HSTS and HTTPS redirect?


idubrov:
Essentially, the same situation as in https://github.com/kubernetes/ingress/issues/314 but with HAProxy Ingress controller.

If I'm terminating SSL on ELB and use HTTP (or PROXY), is there any way to configure this controller to do redirect to HTTPS and return HSTS header?

Seems like, it's not directly supported and the only option I have is to replace the haproxy.tmpl with one that sets up a separate port just for the HTTPS redirect (+ change http frontend to return HSTS). Am I right?

jcmoraisjr:
Please let me know if 74783ec can help.

idubrov:
It looks exactly what I need (though I haven't tried it myself).

Two minor things:

1) It looks to me that `acl is-https-to-http var(txn.hdr_proto) eq https` could be outside of `hasHTTPStoHTTP` condition since it can recognize request as `HTTPS` if it came with `X-Forwarded-Proto: https` no matter if `HTTPStoHTTPPort` is set or not.

2) I find readme a little bit confusing, especially this part:
> this configuration will not work and will either loop because of https redirect or HSTS header will not be added

ColDog:
I tried this out, findings are below: 

I didn't have any issues with HSTS and all that worked well. I actually just ended up setting the `https-to-http` port to `80` which meant it triggered the acl rule on all' traffic for me. I'm not sure if this is how this is meant to be used, but it triggered the desired behaviour for me which is to set `HSTS` headers on both secure and insecure protocols.

HTTPS redirect did not work as I expected. I think this is because the redirect statements are nested under the `UseHTTPS` check: (https://github.com/jcmoraisjr/haproxy-ingress/commit/74783eccd5e7b7a1043cd332c03c31171597cfd1#diff-0aee42afd25a32b87c1c27ad8c705c40R134). I removed these checks on a local branch and everything works as expected.

jcmoraisjr:
@idubrov 

> It looks to me that `acl is-https-to-http var(txn.hdr_proto) eq https` could be outside of `hasHTTPStoHTTP` condition since it can recognize request as HTTPS if it came with `X-Forwarded-Proto: https` no matter if `HTTPStoHTTPPort` is set or not.

The idea here is to declare only what is really needed. One that does ssl offload on HAProxy shouldn't have to know what this strange and unused acl means.

> I find readme a little bit confusing, especially this part:

[...]

PR #40, opened 2017-08-30
HSTS and HTTPS redirect for case when SSL is terminated at the load balancer

idubrov:
Follow-up pull request for #38 

I don't like that I had to copy the whole httpfront section.

Also, this new https2httpfront only loop over non-HTTPS servers -- not sure if this is correct (although in this mode HTTPS servers are sort of useless anyway as we terminate SSL before getting to HAProxy).

Any comments/suggestions? Makes sense/does not?

### DIFF
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -104,6 +104,7 @@ func newHAProxyConfig(haproxyController *HAProxyController) *types.HAProxyConfig
 		StatsProxyProtocol:          false,
 		HTTPLogFormat:               "",
 		TCPLogFormat:                "",
+		ForceSSLRedirect:            false,
 	}
 	if haproxyController.configMap != nil {
 		utils.MergeMap(haproxyController.configMap.Data, &conf)
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -74,6 +74,7 @@ type (
 		StatsProxyProtocol          bool   `json:"stats-proxy-protocol"`
 		HTTPLogFormat               string `json:"http-log-format"`
 		TCPLogFormat                string `json:"tcp-log-format"`
+		ForceSSLRedirect            bool   `json:"force-ssl-redirect"`
 	}
 	// Userlist list of users for basic authentication
 	Userlist struct {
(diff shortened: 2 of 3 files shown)


jcmoraisjr:
This does make sense, yes. I'm fine with the proposal but I wouldn't like to support another duplication on the template. Let me think a few more time about this (for 1..2? include/import? ...?). New ideas are very welcome.

idubrov:
Another option I was considering is to

 * Always send HSTS on port 80 (browser would ignore it if received via HTTP)
 * Make :81 to be "redirect" port that will always send traffic to https

In this case, the configuration would be inverse, HTTP would go to 81 to get redirect and HTTPS would go to 80.

However, I don't like two things here:
 * Sending HSTS over HTTP is sort of untidy. You can check X-Forwarded-Proto, but with TCP backend AWS ELB does not send X-Forwarded-Proto.
 * No control over which HTTP traffic gets redirected (it's all or nothing). In the implementation I submitted it (should, anyway :) ) work in a way that it redirects only if a) global force-ssl-redirect is set to true b) per-ingress annotation ingress.kubernetes.io/force-ssl-redirect is set to true

jcmoraisjr:
Is there any difference in headers between HTTP and HTTPS requests? If so the same listener could handle both requests, changing the behavior with ACL - and inside an annotation saying "I'm behind a LB".

idubrov:
If load balancer uses HTTP protocol to the backend, it sets X-Forwarded-Proto header to either http/https.

However, with PROXY protocol (TCP backend in AWS ELB), there is no difference -- requests are exactly the same.

idubrov:
If AWS load balancer uses HTTP protocol to the backend, it sets X-Forwarded-Proto to indicate original scheme (it's either "http" or "https"). 

However, if ELB uses TCP protocol to the backend (PROXY protocol), there is no difference, requests are exactly the same.

ColDog:
Any updates on this PR? I'm very interested in the solution to this issue, if you need a hand testing or writing I'm interested in helping.

idubrov:
@ColDog check #38. It seems like it would solve that issue (I haven't tried it myself, though).

PR #42, opened 2017-09-01
Drain Endpoint on Removal

ColDog:
HAProxy has both a maintenance state and a drain state that a backend can be put into through the socket. http://cbonte.github.io/haproxy-dconv/1.8/management.html#9.3-set%20server.

I think that the ingress controller should put the backend into the drain state to allow currently active connections to drain but remove this backend from load balancing. One consequence of this is checks will still be performed, as a result it may be necessary to put the backend into maintenance state after a specific interval.

### DIFF
--- a/pkg/controller/stats_socket.go
+++ b/pkg/controller/stats_socket.go
@@ -62,7 +62,7 @@ func endpointsSubtract(s1, s2 map[string]*ingress.Endpoint) map[string]*ingress.
 // remove Ingress Endpoint from a backend by disabling a specific server slot
 func removeEndpoint(statsSocket, backendName, backendServerName string) bool {
 	err := utils.SendToSocket(statsSocket,
-		fmt.Sprintf("set server %s/%s state maint\n", backendName, backendServerName))
+		fmt.Sprintf("set server %s/%s state drain\n", backendName, backendServerName))
 	if err != nil {
 		glog.Warningln("failed socket command srv remove")
 		return false


jcmoraisjr:
Couldn't find this behaviour in the docs and couldn't find the time to test this myself. Did you reproduce this in your local tests, ie, active connections are dropped when backends are put in maintenance mode?

aiharos:
Currently active connections shouldn't be dropped after a server enters maint mode, only new ones will not be accepted. Entering drain mode will allow a server to continue accepting even new connections but only if the persistence cookie is set on them.

For this use-case I don't think the drain mode will be applicable, since if the controller receives the updated backend list from ingress after the pods are already destroyed there will be nothing to drain from.

There seems to be a grace period between when endpoints are removed and when the pods are actually killed, but I don't know if there is a mechanism available to ensure the ordering of these operations. Due to same reason it seems to me that even after a server is set to maint (or drain) the pod be killed with connections still active when the grace period expires.

ColDog:
I was not able to reproduce anything in testing. I think now that @aiharos is correct that currently active connections shouldn't be dropped after a server enters maintenance mode, the docs are ambiguous on this but I think it's implied. So I think this PR can be closed.

The original intent was to find a way to drain connections before a pod is killed. My strategy now is:
Use a `preStop` hook that signals the application to begin returning 500 on it's healthcheck endpoint and to start draining it's own connections. Once all connections are drained it should exit.

This works but it would be nice to have a solution implemented in the load balancer layer. I can't see how to do this at the moment.

aiharos:
You can also try the following:
- increase the terminationGracePeriodSeconds to something generous in the deployment spec
- catch the TERM signal sent to the app and instead of it exiting immediately have it exit after the last connection is served

As soon as the deployment is scaled down extra pods should be set into maint mode by the controller, but with them still alive (due to handling the TERM signal) they will have terminationGracePeriodSeconds time to complete serving active connections and shut down on their own, before they get the KILL signal and are removed.

That or the preStop hook should have the app handle the shutdown better (even without returning 500 on healthchek endpoint, once set for the targeted pods maint mode will not send new connections to them). Connections can still be dropped when the pods are killed depending on the grace period, but you can tune that towards your app for minimum hassle.

ColDog:
Right, I think handling the sigterm makes the most sense. I think sending 500 on the healthcheck endpoint is unnecessary even with my use of the preStop hook because Kubernetes should remove the app from the endpoints resource before sending the sigterm or preStop hook.

This doc here is a good resource: https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.

Thanks for your help! This gave a great overview of how to handle graceful termination.

PR #44, opened 2017-09-09
Add support for server-alias annotation

### LINKED ISSUE #43: Allow RegExp for hostnames
I need to have a possibility to set a hostname as a regular expression, e.g. `[a-z0-9]{10}.domain.com`

AFAIK, Ingress rules don't support it because it wasn't expected to use it that way:
```
The Ingress "test-frontend" is invalid: spec.rules[0].host: Invalid value: "[a-z0-9]{10}.domain.com": a wildcard DNS-1123 subdomain must start with '*.', followed by a valid DNS subdomain, which must consist of lower case alphanumeric characters, '-' or '.' and end with an alphanumeric character (e.g. '*.example.com', regex used for validation is '\*\.[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

So I believe that it could be implemented via custom annotations.

Scukerman:
Fixes #43.
Needs review.

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -98,6 +98,9 @@ frontend httpfront
 {{ else }}
     acl ishost-{{ $server.HostnameLabel }} hdr(host) {{ $server.Hostname }} {{ $server.Hostname }}:80
 {{ end }}
+{{ if ne $server.Alias "" }}
+    acl ishost-{{ $server.HostnameLabel }}-alias {{ if isRegexHostname $server.Alias }}hdr_reg(host) '{{ aliasRegex $server.Alias }}'{{ else }}hdr(host) {{ $server.Alias }} {{ $server.Alias }}:80{{ end }}
+{{ end }}
 {{ end }}
 {{ range $server := $ing.HAServers }}
 {{ if $server.UseHTTP }}
@@ -143,6 +146,22 @@ frontend httpfront
 {{ end }}
 {{ end }}
 {{ end }}
+{{/* Aliases loop */}}
+{{ range $server := $ing.HAServers }}
+{{ if ne $server.Alias "" }}
+{{ if $server.UseHTTPS }}
+{{ if $server.SSLRedirect }}
+    redirect scheme https if ishost-{{ $server.HostnameLabel }}-alias{{ if $hasHTTPStoHTTP }} !is-https-to-http{{ end }}
+{{ else }}
+{{ range $location := $server.Locations }}
+{{ if $location.Rewrite.SSLRedirect }}
+    redirect scheme https if ishost-{{ $server.HostnameLabel }}-alias{{ $location.HAMatchPath }}{{ if $hasHTTPStoHTTP }} !is-https-to-http{{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
 {{ range $server := $ing.HAServers }}
 {{ if or $server.UseHTTP $hasHTTPStoHTTP }}
 {{ range $location := $server.Locations }}
@@ -152,6 +171,16 @@ frontend httpfront
 {{ end }}
 {{ range $location := $server.Locations }}
     use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }} { path_beg {{ $location.Path }} }
+{{ end }}
+{{ end }}
+{{ end }}
+{{/* Aliases loop */}}
+{{ range $server := $ing.HAServers }}
+{{ if or $server.UseHTTP $hasHTTPStoHTTP }}
+{{ range $location := $server.Locations }}
+{{ if ne $server.Alias "" }}
+    use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }}-alias { path_beg {{ $location.Path }} }
+{{ end }}
 {{ end }}
 {{ end }}
 {{ end }}
@@ -178,6 +207,14 @@ frontend httpsfront
     use_backend httpsback-{{ $server.HostnameLabel }} if { req.ssl_sni -m reg -i {{ hostnameRegex $server.Hostname }} }
 {{ else }}
     use_backend httpsback-{{ $server.HostnameLabel }} if { req.ssl_sni -i {{ $server.Hostname }} }
+{{ end }}
+{{ end }}
+{{ end }}
+{{/* Aliases loop */}}
+{{ range $server := $ing.HAServers }}
+{{ if $server.UseHTTPS }}
+{{ if ne $server.Alias "" }}
+    use_backend httpsback-{{ $server.HostnameLabel }} if { req.ssl_sni -i {{ if isRegexHostname $server.Alias }}-m reg '{{ aliasRegex $server.Alias }}'{{ else }}{{ $server.Alias }}{{ end }} }
 {{ end }}
 {{ end }}
 {{ end }}
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -167,6 +167,7 @@ func (cfg *haConfig) createHAProxyServers() {
 			Locations:       haLocations,
 			SSLRedirect:     sslRedirect,
 			CertificateAuth: server.CertificateAuth,
+			Alias:           server.Alias,
 		}
 		// Ingress uses `_` hostname as default server
 		if server.Hostname == "_" {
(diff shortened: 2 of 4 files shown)


Scukerman:
Tested using this manifest:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: echoheaders
  annotations:
    kubernetes.io/ingress.class: haproxy
    ingress.kubernetes.io/server-alias: "[a-z0-9]{10}.domain.com"
spec:
  tls:
  - hosts:
    - 123456789012345.domain.com
    secretName: tls-domain-com
  rules:
  - host: 123456789012345.domain.com
    http:
      paths:
      - path: /
        backend:
          serviceName: echoheaders
          servicePort: 80
```

The real domain name has been replaced by domain.com.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -87,8 +87,14 @@ frontend httpfront
...
 {{ if isWildcardHostname $server.Hostname }}
     acl ishost-{{ labelize $server.Hostname }} hdr_reg(host) {{ hostnameRegex $server.Hostname }} {{ hostnameRegex $server.Hostname }}:80
+{{ if ne $server.Alias "" }}
+    acl ishost-{{ labelize $server.Hostname }}-alias {{ if isRegexpHostname $server.Alias }}hdr_reg(host) {{ aliasRegex $server.Alias }}{{ else }}hdr(host) {{ $server.Alias }}{{ end }} {{ hostnameRegex $server.Hostname }}:80
+{{ end }}
 {{ else }}
     acl ishost-{{ labelize $server.Hostname }} hdr(host) {{ $server.Hostname }} {{ $server.Hostname }}:80
+{{ if ne $server.Alias "" }}
+    acl ishost-{{ labelize $server.Hostname }}-alias {{ if isRegexpHostname $server.Alias }}hdr_reg(host) {{ aliasRegex $server.Alias }}{{ else }}hdr(host) {{ $server.Alias }}{{ end }} {{ $server.Hostname }}:80
+{{ end }}
```
jcmoraisjr: These two `if ne $serverAlias ""` could be outside `if isWildcard` condition. The `:80` should also use the `$serverAlias`.

Note also that you shoud quote your regex or escape spaces, otherwise HAProxy will read two broken regex.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -135,6 +141,9 @@ frontend httpfront
 {{ end }}
 {{ range $location := $server.Locations }}
     use_backend {{ $location.Backend }} if ishost-{{ labelize $server.Hostname }} { path_beg {{ $location.Path }} }
+{{ if ne $server.Alias "" }}
+    use_backend {{ $location.Backend }} if ishost-{{ labelize $server.Hostname }}-alias { path_beg {{ $location.Path }} }
+{{ end }}
```
jcmoraisjr: Create a new iteration in order to move all alias to the bottom. This will ensure that hostnames will match instead of a regex.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -160,6 +169,13 @@ frontend httpsfront
 {{ else }}
     use_backend httpsback-{{ labelize $server.Hostname }} if { req.ssl_sni -i {{ $server.Hostname }} }
 {{ end }}
+{{ if ne $server.Alias "" }}
+{{ if isRegexpHostname $server.Alias }}
+    use_backend httpsback-{{ labelize $server.Hostname }} if { req.ssl_sni -m reg -i {{ aliasRegex $server.Alias }} }
+{{ else }}
+    use_backend httpsback-{{ labelize $server.Hostname }} if { req.ssl_sni -i {{ $server.Alias }} }
+{{ end }}
+{{ end }}
```
jcmoraisjr: Move this check to a new loop as well in order to match hostnames first.

jcmoraisjr: About `aliasRegex`, what about remove it altogether? This would avoid issues like trying to use the `.` as a regex instead of as a literal. If so you could also simplify the expression above and use `isRegexpHostname` inline - check in a single line if `-m reg` will/won't be used.

rootfs/etc/haproxy/template/haproxy.tmpl
Scukerman: I haven't decided yet what I should do with `:80` at the end of ACL. If alias is used as a regex then `:80` is supposed to be a part of regex. But what if I have a dollar sign is placed at the end of the regex?

Scukerman: oh, I've found what you made in 917bb1032df09d0b7607e7f040161a595510ac8b. Ok then.

rootfs/etc/haproxy/template/haproxy.tmpl
Scukerman: Actually I was too sleepy and I thought for some reason that it is a good idea to have this function.

We have two ways of using this annotation:
1. Hostname as a regex.
2. Hostname as a plain domain.

Both ways could be checked by `isRegexpHostname`.

So if we single-quote all the regexes in the template, then we can get rid of `aliasRegex` from all the places where it is used.

What do you think?

jcmoraisjr:
> I haven't decided yet what I should do with :80 at the end of ACL.

Some clients add the port number in the `Host:` header, which is what HAProxy will match to choose a backend. Because of that that `:80` is there.

> We have two ways of using this annotation:
>
>     Hostname as a regex.
>     Hostname as a plain domain.
>
> Both ways could be checked by `isRegexpHostname`.
>
> So if we single-quote all the regexes in the template, then we can get rid of `aliasRegex` from all the places where it is used.
>
> What do you think?

The regex should match the whole hostname so it should start with `^` and end with `$`. Note also that, as you have already pointed out, we cannot concat `^alias-regex$` and `:80`. The current strategy used in `hostnameRegex` could be used as a starting point.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -98,6 +98,10 @@ frontend httpfront
 {{ else }}
     acl ishost-{{ $server.HostnameLabel }} hdr(host) {{ $server.Hostname }} {{ $server.Hostname }}:80
 {{ end }}
+{{ if ne $server.Alias "" }}
+    acl ishost-{{ $server.HostnameLabel }}-alias {{ if isRegexpHostname $server.Alias }}hdr_reg(host) '{{ aliasRegex $server.Alias }}' '{{ aliasRegex $server.Alias }}:80'{{ else }}hdr(host) {{ $server.Alias }} {{ $server.Alias }}:80{{ end }}
+{{ end }}
+{{ end }}
```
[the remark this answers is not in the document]
Scukerman: Still can't figure out what should I do.
Simplify it better than it is? I can get rid of using `aliasRegex`

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -181,6 +190,11 @@ frontend httpsfront
 {{ end }}
 {{ end }}
 {{ end }}
+{{ range $server := $ing.HTTPSServers }}
+{{ if ne $server.Alias "" }}
+    use_backend httpsback-{{ $server.HostnameLabel }} if { req.ssl_sni {{ if isRegexpHostname $server.Alias }}-m reg {{ end }}-i '{{ $server.Alias }}' }
+{{ end }}
+{{ end }}
```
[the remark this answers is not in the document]
Scukerman: It's my bad :)

rootfs/etc/haproxy/template/haproxy.tmpl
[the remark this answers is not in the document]
Scukerman: Oh, got it.

Scukerman:
> the annotation should be as simple as eg [a-z]+\.mydomain\.com.
Actually, you can't use backslashes in annotation.

rootfs/etc/haproxy/template/haproxy.tmpl
Scukerman: No, we can't. Annotation doesn't accept backslashes so I had to keep this function.

jcmoraisjr:
Please try to rebase all these commits into a single one. In case of any problem, just ping back, I'll merge it like it is.

Scukerman:
Couldn't squash correctly because of merge. So I have made a brand-new commit with all changes and force-pushed here.

And also I added https redirect for aliases.
Just in case, check it once again please if that's all right.

jcmoraisjr:
Sounds great, thanks.

Issue #52, opened 2017-09-21
Can't rely on ssl_fc_has_crt


Scukerman:
I ran into an issue when `ssl_fc_has_crt` is not always `true` when user provides client certificate, therefore user gets error page which is telling that no certificate was received.
Haproxy docs suggest to use `ssl_c_used` instead of `ssl_fc_has_crt`.

```
ssl_fc_has_crt : boolean
Returns true if a client certificate is present in an incoming connection over
SSL/TLS transport layer. Useful if 'verify' statement is set to 'optional'.

***
Note: on SSL session resumption with Session ID or TLS ticket, client
certificate is not present in the current connection but may be retrieved
from the cache or the ticket. So prefer "ssl_c_used" if you want to check if
current SSL session uses a client certificate.
***
```

So we can't rely on `ssl_fc_has_crt`

PR #53, opened 2017-09-22
Fix handling error496

Scukerman:
Fixes: #52
Replace variable that detects usage of client certificate from `ssl_fc_has_crt` to `ssl_c_used`.


jcmoraisjr:
Sounds good, thanks!

PR #56, opened 2017-10-03
Add an X-Forwarded-Proto header in https frontends

cbonte:
Some applications requires a X-Forwarded-Proto header when the reverse
proxy offloads SSL.
By adding "http-request set-header X-Forwarded-Proto https", the header
will be added if none exist, or it will replace previous occurences if
it was already present.


jcmoraisjr:
Thanks!

Issue #55, opened 2017-10-02
Annotation support: rewrite-target


ausmith:
[Kubernetes ingress controllers](https://github.com/kubernetes/ingress) supports the annotation [`ingress.kubernetes.io/rewrite-target`](https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md#rewrite). This is a valuable annotation to support in microservice architectures. 

Example: I have 2 services, an API and a telemetry service, both on myapi.domain.com. The `path` setup in the kubernetes ingress definition for the API is `/api`, while `/telemetry` is used for the telemetry service. We then might re-use that same telemetry service with a different domain but for some reason the path has to be `/stats` instead of `/telemetry`. As such, the telemetry service has been coded to expect inbound traffic on the path `/` and the `rewrite-target` is used to overwrite `/telemetry` and `/stats` to `/`.

We believe (and are testing) that the resulting haproxy configuration should likely be something like:
```
reqrep ^([^\ :]*)\ /telemetry/(.*)     \1\ /\2  if { path_beg /telemetry }
use_backend out-telemetry-80 if { path_beg /telemetry }
```

We have already eliminated `regsub` from the running for this config as it does not seem to support capture groups (does not allow closing parentheses in the regex). Thoughts?

ausmith:
The following rewrite target options are available in master as of #57 getting merged.

|rewrite target|ingress path|request path|output|
|-------------|-------------|-------------|------|
|/|/abc|/abc|/|
|/|/abc|/abc/|/|
|/|/abc|/abc/x|/x|
|/y|/abc|/abc|/y|
|/y|/abc|/abc/|/y/|
|/y|/abc|/abc/x|/y/x|
|/|/abc/|/abc|**404**|
|/|/abc/|/abc/|/|
|/|/abc/|/abc/x|/x|

That meets my business needs for this annotation as described in the original issue. @jcmoraisjr should we leave this issue open for handling annotations with a trailing slash or open a separate issue for that?

jcmoraisjr:
No need to create a new one, this very issue is enough.

elideveloper:
There is a problem if a requested path contains some repeated value. Rewrite target "/" just cuts it down double time. 
For instance: 
request is  "/person/person/find"
we need to leave "/person/find" but get just "/find"
the same with "/p/person/find" -> "/erson/find".
So, is it a valid behaviour and how to deal with it?
our config:
apiVersion: extensions/v1beta1
[...]

PR #57, opened 2017-10-03
Add support for rewrite-target annotation

ausmith:
#55 

Adds support for ingress annotation: `ingress.kubernetes.io/rewrite-target`.

Our test case for this work was `ingress.kubernetes.io/rewrite-target: /` as that has been the most common use case for us internally. This implementation SEEMS to be feature similar to the current implementation in the [kubernetes/ingress](https://github.com/kubernetes/ingress) nginx ingress controller, but our test cases were rather limited so unsure.

Used https://gist.github.com/kushmansingh/47644ef4db96f11b1ea450ae19072552 to do our testing but could have used most any other container that echoes the headers.

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -105,6 +105,8 @@ frontend httpfront
     acl ishost-{{ $server.HostnameLabel }}-alias {{ if isRegexHostname $server.Alias }}hdr_reg(host) '{{ aliasRegex $server.Alias }}'{{ else }}hdr(host) {{ $server.Alias }} {{ $server.Alias }}:80{{ end }}
 {{ end }}
 {{ end }}
+
+    http-request set-var(req.path) path
 {{ range $server := $ing.HAServers }}
 {{ if $server.UseHTTP }}
 {{ range $location := $server.Locations }}
@@ -119,6 +121,19 @@ frontend httpfront
 {{ end }}
 {{ end }}
 {{ end }}
+
+{{ range $server := $ing.HAServers }}
+{{ if $server.UseHTTP }}
+{{ range $location := $server.Locations }}
+{{ $rewriteTarget := $location.Rewrite.Target }}
+{{ if ne $rewriteTarget "" }}
+    reqrep ^([^\ :]*)\ {{ $location.Path }}/(.*$) \1\ {{ if eq $rewriteTarget "/" }}/{{ else }}{{ $rewriteTarget }}/{{ end }}\2 if { var(req.path) -m beg {{ $location.Path }} } {{ if ne $server.Alias "" }} ishost-{{ $server.HostnameLabel }} or ishost-{{ $server.HostnameLabel }}-alias {{ else }} ishost-{{ $server.HostnameLabel }} {{ end }}
+    reqrep ^([^\ :]*)\ {{ $location.Path }}(.*$) \1\ {{ $rewriteTarget }}\2 if { var(req.path) -m beg {{ $location.Path }} } {{ if ne $server.Alias "" }} ishost-{{ $server.HostnameLabel }} or ishost-{{ $server.HostnameLabel }}-alias {{ else }} ishost-{{ $server.HostnameLabel }} {{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
+{{ end }}
+
 {{ if eq $cfg.Forwardfor "add" }}
     reqidel ^X-Forwarded-For:.*
     option forwardfor
@@ -129,7 +144,7 @@ frontend httpfront
 {{ if $server.UseHTTP }}
 {{ $appRoot := $server.RootLocation.Rewrite.AppRoot }}
 {{ if ne $appRoot "" }}
-    redirect location {{ $appRoot }} if ishost-{{ $server.HostnameLabel }} { path / }
+    redirect location {{ $appRoot }} if ishost-{{ $server.HostnameLabel }} { var(req.path) -m str / }
 {{ end }}
 {{ end }}
 {{ end }}
@@ -169,11 +184,11 @@ frontend httpfront
 {{ if or $server.UseHTTP $hasHTTPStoHTTP }}
 {{ range $location := $server.Locations }}
 {{ if ne $location.Proxy.BodySize "" }}
-    use_backend error413 if ishost-{{ $server.HostnameLabel }} { path_beg {{ $location.Path }} } { req.body_size gt {{ sizeSuffix $location.Proxy.BodySize }} }
+    use_backend error413 if ishost-{{ $server.HostnameLabel }} { var(req.path) -m beg {{ $location.Path }} } { req.body_size gt {{ sizeSuffix $location.Proxy.BodySize }} }
 {{ end }}
 {{ end }}
 {{ range $location := $server.Locations }}
-    use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }} { path_beg {{ $location.Path }} }
+    use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }} { var(req.path) -m beg {{ $location.Path }} }
 {{ end }}
 {{ end }}
 {{ end }}
@@ -182,7 +197,7 @@ frontend httpfront
 {{ if or $server.UseHTTP $hasHTTPStoHTTP }}
 {{ range $location := $server.Locations }}
 {{ if ne $server.Alias "" }}
-    use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }}-alias { path_beg {{ $location.Path }} }
+    use_backend {{ $location.Backend }} if ishost-{{ $server.HostnameLabel }}-alias { var(req.path) -m beg {{ $location.Path }} }
 {{ end }}
 {{ end }}
 {{ end }}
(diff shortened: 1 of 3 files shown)


ausmith:
Have now also tested `ingress.kubernetes.io/rewrite-target: /yourserver` and doing a curl against the kube cluster with this gives the following results:

```
curl http://test-ingress.com/myserver
# output: 2017/10/03 15:57:54  /yourserver

curl http://test-ingress.com/myserver/xyz
# output: 2017/10/03 15:57:54  /yourserver
```

With `ingress.kubernetes.io/rewrite-target: /yourserver` we get:

```
curl http://test-ingress.com/myserver
# output: 2017/10/03 16:00:51  /yourserver/

curl http://test-ingress.com/myserver/xyz
# output: 2017/10/03 16:00:25  /yourserver/xyz
```

So it is the responsibility of the ingress definer to include the trailing slash in the annotation or they might get unexpected responses.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -284,14 +296,25 @@ frontend httpsfront-{{ $host }}
 {{ if ne $appRoot "" }}
     redirect location {{ $appRoot }} if { path / }
 {{ end }}
+
 {{ range $location := $server.Locations }}
 {{ if ne $location.Proxy.BodySize "" }}
-    use_backend error413 if { path_beg {{ $location.Path }} } { req.body_size gt {{ sizeSuffix $location.Proxy.BodySize }} }
+    use_backend error413 if {{ $location.Backend }}_path_match { req.body_size gt {{ sizeSuffix $location.Proxy.BodySize }} }
```
jcmoraisjr: acl should be declared before being used.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -108,6 +108,18 @@ frontend httpfront
...
 {{ if $server.UseHTTP }}
 {{ range $location := $server.Locations }}
+    http-request set-var(req.ingress_orig_path) path
+    acl {{ $location.Backend }}_path_match var(req.ingress_orig_path) -m beg {{ $location.Path }} if or ishost-{{ $server.HostnameLabel }} ishost-{{ $server.HostnameLabel }}-alias
+{{ $rewriteTarget := $location.Rewrite.Target }}
+{{ if ne $rewriteTarget "" }}
+{{ if ne $server.Alias "" }}
+    reqrep ^([^\ :]*)\ {{ $location.Path }}[/]?(.*) \1\ {{ $rewriteTarget }}\2 if {{ $location.Backend }}_path_match ishost-{{ $server.HostnameLabel }} or ishost-{{ $server.HostnameLabel }}-alias
+{{ else }}
+    reqrep ^([^\ :]*)\ {{ $location.Path }}[/]?(.*) \1\ {{ $rewriteTarget }}\2 if {{ $location.Backend }}_path_match ishost-{{ $server.HostnameLabel }}
```
jcmoraisjr: These lines are almost the same, use an inline `if` instead.

Prefer http-request if possible, otherwise declare them after all http-requests in order to avoid warning from haproxy cfg parser.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -108,6 +108,18 @@ frontend httpfront
 {{ range $server := $ing.HAServers }}
 {{ if $server.UseHTTP }}
 {{ range $location := $server.Locations }}
+    http-request set-var(req.ingress_orig_path) path
+    acl {{ $location.Backend }}_path_match var(req.ingress_orig_path) -m beg {{ $location.Path }} if or ishost-{{ $server.HostnameLabel }} ishost-{{ $server.HostnameLabel }}-alias
```
jcmoraisjr: Add `ishost-...-alias` inside a condition - ne Alias "".

Prefer loop acls+setvar in a separate loop of reqrep. Worst from the point of view of the template but better to the final cfg.

rootfs/etc/haproxy/template/haproxy.tmpl
```diff
@@ -108,6 +108,18 @@ frontend httpfront
 {{ range $server := $ing.HAServers }}
 {{ if $server.UseHTTP }}
 {{ range $location := $server.Locations }}
+    http-request set-var(req.ingress_orig_path) path
```
jcmoraisjr: my assumption is that you are using a var because the path is being changed before match on use_backend, is it right? Note that you are using the same var name to different hosts and paths.

ausmith: Right, I was finding that after the `reqrep` we could not execute the same `acl` against the path when choosing the backend. OH! Probably do need to move the variable setting to outside the loop though, good call there.

rootfs/etc/haproxy/template/haproxy.tmpl
ausmith: Unfortunately, the documentation indicates `http-request` does not support capture groups, hence we had to fall back to the `reqrep` option instead.

ausmith:
I believe I've handled each of the comments and have done local testing to verify that these changes still behave properly. 👍

jcmoraisjr:
Some other changes:

* Why use `[/]?(.*)` instead of just `(.*)`?
* Why use an acl instead change `path_beg` to `var(req.path) -m beg`? Note that if you have more than one location in the same server you'll have name collision
* A shorter var name like `req.path` would be fine
* `app-root` annotation also match `path` and should be changed
* There is also an ugly workaround named `HAMatchPath` (`config.go`) which also match `path_beg` and should also be changed
* `http-request` is still after a `rsprep` on https proxy - move the `http-request` in order to avoid a haproxy warn
* Try to rebase your next changes and the old commits into a single new commit - or leave it as it is if thing goes wrong

ausmith:
@jcmoraisjr ty for the feedback! I did a rebase with all of it in one commit as requested. Notes from your other bullets:
- Why use `[/]?(.*)` instead of just `(.*)`?
  - I have tested without the `[/]?(.*)` and it causes a redirect instead of the expected behavior if the path is not an exact match (i.e. `/myserver/` instead of `/myserver`). The trailing slash allows all of `/myserver`, `/myserver/`, and `/myserver/xyz` to match.
- Why use an acl instead change `path_beg` to `var(req.path) -m beg`? Note that if you have more than one location in the same server you'll have name collision
  - I have switched them all back to non-ACLs. I had read that using anonymous conditions meant they were compiled at runtime rather than at config load time somewhere, so unsure about the performance difference between ACLs and anonymous conditions.
- A shorter var name like `req.path` would be fine
  - I did this as you are totally right.
- `app-root` annotation also match `path` and should be changed
  - Fixed.
- There is also an ugly workaround named `HAMatchPath` (`config.go`) which also match `path_beg` and should also be changed
  - Fixed.
- `http-request` is still after a `rsprep` on https proxy - move the `http-request` in order to avoid a haproxy warn
  - Fixed.

jcmoraisjr:
Great, we're almost there:

Setvar is inside a loop (http proxy) but you need just one.

I'm worried about the regex on behalf of consistency. I couldn't reproduce the redirect, perhaps your backend service did it? On the other hand `[/]?` is removing a slash from the url and I'm ending up with (where path is `/abc` and rewrite is`/xpto`):

Ok:

* `/abc` -> `/xpto`
* `/abcd` -> `/xptod`

Not ok:

* `/abc/` -> `/xpto`
* `/abc/d` -> `/xptod`

> I have tested without the `[/]?(.*)` and ...

Did you mean `[/]?`? This was what I mean.

ausmith:
So, I have figured out the rewrite target issue overwriting `/` unexpectedly and added a caveat to the README about rewrite target not supporting a trailing `/`. Does that cover all cases now?

ausmith:
Only change since last comment is I found and fixed the source of this warning:
```
[WARNING] 279/130942 (98) : parsing [/etc/haproxy/haproxy.cfg:194] : a 'http-request' rule placed after a 'reqxxx' rule will still be processed before.
[WARNING] 279/130942 (98) : parsing [/etc/haproxy/haproxy.cfg:195] : a 'http-request' rule placed after a 'reqxxx' rule will still be processed before.
```
Required moving `http-request set-header X-Forwarded-Proto https` and the `set-var` that this PR adds above the `reqidel` on line 284 of the template.

jcmoraisjr:
Two small changes in the readme:

* Add `[1] only in snapshot tag` - just below `[0] canary` and change the tag of this new annotation from `[0]` to `[1]`
* Link the annotation to the description - have a look at how affinity was linked

And let me know what you think about this approach (not tested, just the idea):

`... {{ $location.Path }}(.*$)  \1\ {{ $rewriteTarget }}{{ if and (ne $rewriteTarget "/") (endsWith $location.Path "/")}}/{{ end }}\2  if...`

in order to:

* use only one `reqrep` (per frontend)
* change from _"unsupported trailing slash"_ to _"use the same pattern used in ingress path"_ - both ingress path and rewrite should, or should not, have the trailing slash

`endsWith` would be a new func to be implemented (template.go)

ausmith:
I have made the README adjustments accordingly.

The `reqrep` change might be possible, but I am running out of engineering time at work to make that tweak and thoroughly test it. What do you think about supporting a trailing slash on the `rewrite-target` annotation being a future enhancement @jcmoraisjr ? I will be happy to help continue expanding the `rewrite-target` capabilities down the road but do not have the `golang` experience to quickly knock out your suggestion.

jcmoraisjr:
Sounds good. The current implementation is good enough to be merged and we can evolve the config along with the 0.5 development. Btw thanks for helping us making this controller even better.

Issue #84, opened 2017-12-24
FullSlots BackendEndpoint is incorrect


EmpireJones:
I believe in [stats_socket.go:201](https://github.com/jcmoraisjr/haproxy-ingress/blob/939bd129c86d9b27b12e6d7a50c799d8496ab8f3/pkg/controller/stats_socket.go#L201) BackendEndpoint is getting a local memory address for endpoint, which is re-used for each iteration of the loop, meaning that all of the BackendEndpoints in FullSlots are set to the last of the list of endpoints:

    // use addr:port as BackendServerName, don't generate empty slots
    for _, endpoint := range updBackendsMap[backendName].Endpoints {
    	target := fmt.Sprintf("%s:%s", endpoint.Address, endpoint.Port)
    	newBackend.FullSlots[target] = types.HAProxyBackendSlot{
    		BackendServerName: target,
    		BackendEndpoint:   &endpoint,
    	}
    }

To fix this, you should be able to make a local copy of the object, and then take the address of that.

jcmoraisjr:
Fixed on #85

PR #85, opened 2017-12-26
Correctly copy Endpoints to FullSlots

EmpireJones:
The Endpoint address copied to FullSlots was a local variable re-used by a loop. This was resulting in all of the endpoints pointing to the last one copied.

Issue: #84


jcmoraisjr:
Hi, thanks for supporting this Ingress controller! Please fix also the dynamic scaling code which has the same issue.

EmpireJones:
Good catch. Added!

jcmoraisjr:
@EmpireJones thank!

@aiharos fyi. Only the port number is being used here, because of that we didn't see this before. I'm planning to revisit this unless you already have changes to the stats_socket on your todo list.

PR #89, opened 2018-02-09
Add a cookie strategy

brianloss:
This is an additional annotation that will allow the user to select a different HAProxy cookie strategy other than insert. Supported additions are prefix and rewrite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ The following annotations are supported:
 ||`ingress.kubernetes.io/secure-backends`|[true\|false]|-|
 ||`ingress.kubernetes.io/secure-verify-ca-secret`|secret name|-|
 ||[`ingress.kubernetes.io/session-cookie-name`](#affinity)|cookie name|-|
+|`[0]`|[`ingress.kubernetes.io/session-cookie-strategy`](#affinity)|cookie strategy|-|
 ||`ingress.kubernetes.io/ssl-passthrough`|[true\|false]|-|
 ||`ingress.kubernetes.io/ssl-redirect`|[true\|false]|[doc](/examples/rewrite)|
 ||`ingress.kubernetes.io/app-root`|/url|[doc](/examples/rewrite)|
@@ -72,6 +73,7 @@ Configure if HAProxy should maintain client requests to the same backend server.
 
 * `ingress.kubernetes.io/affinity`: the only supported option is `cookie`. If declared, clients will receive a cookie with a hash of the server it should be fidelized to.
 * `ingress.kubernetes.io/session-cookie-name`: the name of the cookie. `INGRESSCOOKIE` is the default value if not declared.
+* `ingress.kubernetes.io/session-cookie-strategy`: the cookie strategy to use (insert, rewrite, prefix). `insert` is the default value if not declared.
 
 Note for `dynamic-scaling` users only: the hash of the server is built based on it's name.
 When the slots are scaled down, the remaining servers might change it's server name on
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -59,6 +59,7 @@ The following annotations are supported:
 ||`ingress.kubernetes.io/secure-backends`|[true\|false]|-|
 ||`ingress.kubernetes.io/secure-verify-ca-secret`|secret name|-|
 ||[`ingress.kubernetes.io/session-cookie-name`](#affinity)|cookie name|-|
+|`[0]`|[`ingress.kubernetes.io/session-cookie-strategy`](#affinity)|cookie strategy|-|
```
jcmoraisjr: This PR belongs to v0.6, so use `[1]` instead and create also a `* ``[1]`` only in ``snapshot`` tag` entry just before this table.

Change also `cookie strategy` text to a list of options. Have a look at eg `forwardfor` configmap option.

pkg/common/ingress/annotations/sessionaffinity/main.go
```diff
@@ -68,6 +74,13 @@ func CookieAffinityParse(ing *extensions.Ingress) *CookieConfig {
 		sn = defaultAffinityCookieName
 	}
 
+	ss, err := parser.GetStringAnnotation(annotationAffinityCookieStrategy, ing)
+
+	if err != nil || !affinityCookieStrategyRegex.MatchString(ss) {
+		glog.V(3).Infof("Invalid or no annotation value found in Ingress %v: %v. Setting it to default %v", ing.Name, annotationAffinityCookieStrategy, defaultAffinityCookieStrategy)
```
jcmoraisjr: This means an invalid configuration, so change the log level to `glog.Warningf`.

pkg/common/ingress/annotations/sessionaffinity/main_test.go
```diff
@@ -86,4 +87,8 @@ func TestIngressAffinityCookieConfig(t *testing.T) {
 	if nginxAffinity.CookieConfig.Name != "INGRESSCOOKIE" {
 		t.Errorf("expected route as sticky-name but returned %v", nginxAffinity.CookieConfig.Name)
 	}
+
+	if nginxAffinity.CookieConfig.Strategy != "insert" {
+		t.Errorf("expected route as sticky-strategy but returned %v", nginxAffinity.CookieConfig.Strategy)
```
jcmoraisjr: ...expected `insert` as sticky-...

pkg/common/ingress/annotations/sessionaffinity/main.go
brianloss: I changed to a warning if there's an invalid strategy specified, but left as an info if there is no strategy specified since the intent was to have backwards compatibility and not require people to use this option. Otherwise, anyone who was using affinity but didn't set this would start getting warnings. Hope that's ok, but can change if necessary.

brianloss:
Thanks for the feedback!

pkg/common/ingress/annotations/sessionaffinity/main.go
```diff
@@ -40,6 +43,7 @@ const (
 
 var (
 	affinityCookieHashRegex = regexp.MustCompile(`^(index|md5|sha1)$`)
+	affinityCookieStrategyRegex = regexp.MustCompile(`^(insert|prefix|rewrite)`)
```
jcmoraisjr: Add an endline match `$` here.

brianloss: Will fix, thanks.

jcmoraisjr:
Thank you very much, merged!

PR #92, opened 2018-02-14
Make docs more consistent

johscheuer:
This PR makes the default deployment a little bit more consistent (uses an extra namespace like most other examples).


jcmoraisjr:
Great, thanks! Merged.

Issue #102, opened 2018-03-06
Trouble setting http basic auth


pedro-gutierrez:
Hi, 

I am having trouble to make http basic auth work for my ingress.  

I believe I have properly setup my secret and my ingress annotations, according to the docs. As a matter of fact, requests are routed to my backend, and indeed the path prefix is stripped, as of my rewrite rule, however, no matter what I do, basic auth challenge never seems to get triggered.

I inspected the generated haproxy.cfg inside my ingress controller pod, and I noticed the txn.path variable is used before it is actually declared, see this snippet:

```
frontend httpfront-...
    ...
    http-request auth realm ... if { var(txn.path) -m ... } !{ http_auth(...)}
    ...
    http-request set-var(txn.path) path
    ...
    use_backend ... if { var(txn.path) -m ... }
...
```

(I omitted the specifics related to my setup)

I am far from an expert when it comes to HAProxy configuration, however I was wondering whether this could be an actual bug, or a simply user error from my side? As a side note, I have succesfully setup http basic auth with other ingress controllers, such as traefik and nginx.

Anyway, If you need further info please let me know. I'll be more than happy to provide with as much detail as needed in order to get this sorted out.

Thanks in advance,
Pedro.

pedro-gutierrez:
Ok I just realized the behavior I described above applies only to  the canary tag. The stable version works as expected.

pedro-gutierrez:
I submitted this PR https://github.com/jcmoraisjr/haproxy-ingress/pull/103

jcmoraisjr:
Nice catch, thanks! I've just updated my tests with scenarios like this one. I updated `master` and `release-0.5` - the later will be used to update `canary` and shotly `latest`. Let me know if this works for you. If you cannot rebuild from sources I can create a new beta before your tests.

jcmoraisjr:
Just released a new beta with this change, tags `v0.5-beta.2` and `canary`.

pedro-gutierrez:
Hi, thanks a lot for your feedback and for the new tag! It is working very well according to my regression suite. Cheers!

jcmoraisjr:
Thank you for your feedback! Closing.

pedro-gutierrez:
Hi again, actually, not sure this was happening before, but http basic auth stops being triggered as soon as I my secret is re-created. I need to re-deploy my haproxy controller and then things work again. Is this a known issue?

jcmoraisjr:
[...]

PR #103, opened 2018-03-06
#102 move declaration of txn.path before it is actually used

pedro-gutierrez:
This is a possible fix for issue #102. 
I did a very basic manual test and it seems to behave ok.


jcmoraisjr:
Hi, thanks for supporting this ingress controller! I've just uploaded f414a6d with another approach which fixes some keyword declaration order. Closing this. Please update #102 if you have any problem.

PR #122, opened 2018-03-16
Use pidof to ensure all previous haproxy processes are shutdown

westse:
This more closely matches [haproxy's documentation of the -sf option](https://cbonte.github.io/haproxy-dconv/1.8/management.html#3), as well as potentially being more safe because each new haproxy process ensures all previous processes sockets are transferred.

### DIFF
--- a/rootfs/haproxy-reload.sh
+++ b/rootfs/haproxy-reload.sh
@@ -49,12 +49,12 @@ case "$1" in
     native)
         CONFIG="$2"
         HAPROXY_PID=/var/run/haproxy.pid
-        haproxy -f "$CONFIG" -p "$HAPROXY_PID" -D -sf $(cat "$HAPROXY_PID" 2>/dev/null || :)
+        haproxy -f "$CONFIG" -p "$HAPROXY_PID" -D -sf $(pidof haproxy 2>/dev/null || :)
         ;;
     reusesocket)
         CONFIG="$2"
         HAPROXY_PID=/var/run/haproxy.pid
-        OLD_PID=$(cat "$HAPROXY_PID" 2>/dev/null || :)
+        OLD_PID=$(pidof haproxy 2>/dev/null || :)
         if [ -S "$HAPROXY_SOCKET" ]; then
             haproxy -f "$CONFIG" -p "$HAPROXY_PID" -sf $OLD_PID -x "$HAPROXY_SOCKET"
         else


jcmoraisjr:
Did you see any problem with the current approach? Are there scenarios where some missing process are not receiving the signal? Afaik the doc is just saying that the `pidof` result is valid as an argument and not that this is a preferable approach.

westse:
We did have problems with stale config which we thought might have been because a previous haproxy process did not get notified to shutdown as we thought it should. This was an attempt to address that problem. The stale config issue has gone away, but unfortunately I cannot say definitively if it was due to this change or the fact that we have since dropped multibinder in favor of haproxy's "hitless reload" feature.

We have been using pidof for a few weeks now and I can at least report with confidence that it works at least as well as the previous approach, so that may help with any concerns you may have about introducing a regression.

jcmoraisjr:
Hi, I'm also confident that this will work and HAProxy should kindly manage signals comming from more than one source - which should start to happen with process waiting for long lived connections to finish.

What I'd like to avoid is to workaround a bug patching it's symptoms. I'm a bit confident that the async reload of multibinder is at least one source of problem, perhaps there are others. Native and reusesocket shouldn't be a problem because, afaics: everything happens synchronously in a single thread with rate limit; every new HAProxy proc notify the old one from pidfile; pidfile is overwritten only if the new proc successfully notified the old one; config file is updated within this same thread before touching HAProxy procs and after HAProxy says everything is fine and running and the config file won't be used anymore.

I'm updating my test environment in order to try to reproduce this issue, so a few more questions:

* You said stale config, which in my understanding means old but valid config. Have you ever see invalid config which would fail HAProxy reload?
* Stale config means config that in some point in time was pretty valid or there was also missing backends and configurations that wasn't touched - iow: stale means only back in time or sometimes means missing config?
* Did you find, in any point in time, more than one HAProxy proc listening 80/443? This is indeed a nice monitoring since HAProxy allows more than one proc listening the same ip/port - kernel will round robin between them - and old procs waiting for long connections to finish should always stop listening tcp ports
* All these symptoms (this and other PRs) happened only with multibinder? Did you change the rate-limit?
* Would you mind helping me fix the root cause of this issue reverting this PR on your enviroment? I'm aware this is a prod environment for your developers so take this without commitment. What I want to check is if there is any issue outside multibinder config.

westse:
> You said stale config, which in my understanding means old but valid config. Have you ever see invalid config which would fail HAProxy reload?

Yes, valid but old config is what we see. When we look at haproxy's on-disk config file it looks correct, but haproxy responds to requests as though it has old config. Usually new ingress rules don't appear to work and instead we get sent to the default backend (as described in https://github.com/jcmoraisjr/haproxy-ingress/issues/127)

> Stale config means config that in some point f ifin time was pretty valid or there was also missing backends and configurations that wasn't touched - iow: stale means only back in time or sometimes means missing config?

If I understand the question, it is valid config but just not up to date. I believe the misbehavior is exclusively when reloads are occurring due to config changes.

> Did you find, in any point in time, more than one HAProxy proc listening 80/443? This is indeed a nice monitoring since HAProxy allows more than one proc listening the same ip/port - kernel will round robin between them - and old procs waiting for long connections to finish should always stop listening tcp ports

Sorry, not sure I understand the question. I don't think we have seen any trouble in this area.

> All these symptoms (this and other PRs) happened only with multibinder? Did you change the rate-limit?

Yes, the stale config issue happened when we were using multibinder, and has stopped now that we instead use the "reusesockets" reload strategy (although as I've noted elsewhere, other variables were changed as well including this PR).  No rate limit changes were tried.

> Would you mind helping me fix the root cause of this issue reverting this PR on your enviroment? I'm aware this is a prod environment for your developers so take this without commitment. What I want to check is if there is any issue outside multibinder config.

I've run out of justification to spend much more time on it, but I see how that would be helpful and may be able to do that at some point. It may be a while before I get to it.

jcmoraisjr:
@westse does your last tests use the feature of this PR?

westse:
@jcmoraisjr No, we are using the special v0.6-a80d2c6 release you made for https://github.com/jcmoraisjr/haproxy-ingress/pull/130, which is working great, so I think that is evidence this PR (122) is not necessary. So I think we are safe to merge or close this PR as you see fit.

jcmoraisjr:
Great, thanks! I'm closing this on behalf of the new strategy used on v0.6.

PR #123, opened 2018-03-16
Changes to support debugging haproxy reload not loading correct config

westse:
- Under high load situations (hundreds of services, certs, and ingress rules), we have seen infrequent but consistent cases of haproxy not reflecting the on-disk config file. One theory is a race condition on new haproxy processes pointing to an ever-changing config file. This commit gives each haproxy process its own config file that does not change. Not only should this remove the risk of loading a partially written file, it will also help debug the issue because we see a snapshot of the config used for each haproxy process.
- Config files are left on disk up to a max threshold, configurable via environment variable, to allow debugging issues.
- Not supported under with reloadStrategy=multibinder since multibinder-haproxy-wrapper currently only supports receiving a USR2 signal to reload a fixed config file

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -31,6 +31,9 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
+	"sort"
+	"strconv"
 )
 
 // HAProxyController has internal data of a HAProxyController instance
@@ -40,7 +43,10 @@ type HAProxyController struct {
 	storeLister    *ingress.StoreLister
 	command        string
 	reloadStrategy *string
-	configFile     string
+	configDir     string
+	configFilePrefix     string
+	configFileSuffix     string
+	maxOldConfigFiles     int
 	template       *template
 	currentConfig  *types.ControllerConfig
 }
@@ -112,16 +118,25 @@ func (haproxy *HAProxyController) ConfigureFlags(flags *pflag.FlagSet) {
 
 // OverrideFlags allows controller to override command line parameter flags
 func (haproxy *HAProxyController) OverrideFlags(flags *pflag.FlagSet) {
-	if *haproxy.reloadStrategy == "native" || *haproxy.reloadStrategy == "reusesocket" {
-		haproxy.configFile = "/etc/haproxy/haproxy.cfg"
-		haproxy.template = newTemplate("haproxy.tmpl", "/etc/haproxy/template/haproxy.tmpl")
-	} else if *haproxy.reloadStrategy == "multibinder" {
-		haproxy.configFile = "/etc/haproxy/haproxy.cfg.erb"
-		haproxy.template = newTemplate("haproxy.cfg.erb.tmpl", "/etc/haproxy/haproxy.cfg.erb.tmpl")
-	} else {
+	haproxy.configDir = "/etc/haproxy"
+	haproxy.configFilePrefix = "haproxy"
+	haproxy.configFileSuffix = ".cfg"
+	haproxy.template = newTemplate("haproxy.tmpl", "/etc/haproxy/template/haproxy.tmpl")
+
+	if !(*haproxy.reloadStrategy == "native" || *haproxy.reloadStrategy == "reusesocket" || *haproxy.reloadStrategy == "multibinder") {
 		glog.Fatalf("Unsupported reload strategy: %v", *haproxy.reloadStrategy)
 	}
+	
+	if *haproxy.reloadStrategy == "multibinder" {
+		haproxy.template = newTemplate("haproxy.cfg.erb.tmpl", "/etc/haproxy/haproxy.cfg.erb.tmpl")
+	}
 	haproxy.command = "/haproxy-reload.sh"
+
+	haproxy.maxOldConfigFiles = 1000
+	envVar := os.Getenv("MAX_OLD_CONFIG_FILES")
+	if maxOldConfigFiles, err := strconv.Atoi(envVar); err == nil {
+		haproxy.maxOldConfigFiles = maxOldConfigFiles
+	}
 }
 
 // SetConfig receives the ConfigMap the user has configured
@@ -156,7 +171,7 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
 		return err
 	}
 
-	err = rewriteConfigFiles(data, *haproxy.reloadStrategy, haproxy.configFile)
+	configFile, err := haproxy.rewriteConfigFiles(data)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -166,56 +181,87 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
...
+	out, err := reloadCmd.CombinedOutput()
 	if len(out) > 0 {
-		glog.Infof("HAProxy output:\n%v", string(out))
+		glog.Infof("HAProxy[pid=%v] output:\n%v", reloadCmd.Process.Pid, string(out))
 	}
 	return err
 }
 
 // RewriteConfigFiles safely replaces configuration files with new contents after validation
-func rewriteConfigFiles(data []byte, reloadStrategy, configFile string) error {
```
ausmith: I noticed that you are rewriting the `multibinder` and non-`multibinder` usages here, which concerns me mostly due to fear of change than anything else. Curious, did you consider the following flow instead and why would it be a bad approach?

```
func rewriteConfigFiles(data []byte, reloadStrategy, configFile string) error {
  // Determine temp file path using a timestamp
  // Write rendered template file there
  // if reload strategy == multibinder then do multibinder logic for moving it into place
  // else do non-multibinder work with config file
  // copy temp file to location for later investigation if required
  // rename file to the /etc/haproxy/haproxy.cfg
}
```

Seems to me that this approach would minimize the amount of code you would need to touch and still give you a rendered template somewhere for investigation? Also unsure about where in that flow a cleanup step would be most efficient.

westse: If I'm following your pseudo code, wouldn't the last step prevent each haproxy process from pointing to a unique timestamped config file? That would defeat the primary changed intended by this pull request (for the non-multibinder usage).

I agree with the sentiment to touch only the code that needs to change. I initially tried this but once I understood what it did I felt some bits of logic were confusing enough that it would be improved with what you see here. For example, I found it suprising that the non-multibinder case used the post-rename config file, while the multibinder case did not. I think this change makes the behavior in each usage more clear, but I may be wrong and am definitely open.

Thanks for taking a look.

pkg/controller/controller.go
```diff
@@ -112,16 +118,25 @@ func (haproxy *HAProxyController) ConfigureFlags(flags *pflag.FlagSet) {
...
-		haproxy.configFile = "/etc/haproxy/haproxy.cfg"
-		haproxy.template = newTemplate("haproxy.tmpl", "/etc/haproxy/template/haproxy.tmpl")
-	} else if *haproxy.reloadStrategy == "multibinder" {
-		haproxy.configFile = "/etc/haproxy/haproxy.cfg.erb"
-		haproxy.template = newTemplate("haproxy.cfg.erb.tmpl", "/etc/haproxy/haproxy.cfg.erb.tmpl")
-	} else {
+	haproxy.configDir = "/etc/haproxy"
+	haproxy.configFilePrefix = "haproxy"
+	haproxy.configFileSuffix = ".cfg"
+	haproxy.template = newTemplate("haproxy.tmpl", "/etc/haproxy/template/haproxy.tmpl")
```
jcmoraisjr: Use newTemplate as the else part of the if multibinder statement.

pkg/controller/controller.go
```diff
@@ -112,16 +118,25 @@ func (haproxy *HAProxyController) ConfigureFlags(flags *pflag.FlagSet) {
...
+	if *haproxy.reloadStrategy == "multibinder" {
+		haproxy.template = newTemplate("haproxy.cfg.erb.tmpl", "/etc/haproxy/haproxy.cfg.erb.tmpl")
+	}
 	haproxy.command = "/haproxy-reload.sh"
+
+	haproxy.maxOldConfigFiles = 1000
+	envVar := os.Getenv("MAX_OLD_CONFIG_FILES")
+	if maxOldConfigFiles, err := strconv.Atoi(envVar); err == nil {
+		haproxy.maxOldConfigFiles = maxOldConfigFiles
+	}
```
jcmoraisjr: use a command-line argument instead, declare at ConfigureFlags func just below reloadStrategy.

I'd also like to see a zero option, which means no old backup config should be used. I have some manual testing scenarios where I need a known filename, perhaps others could have similar issues.

pkg/controller/controller.go
```diff
@@ -166,56 +181,87 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
 		return nil
 	}
 
-	out, err := haproxy.reloadHaproxy()
+	reloadCmd := haproxy.reloadHaproxy(configFile)
```
jcmoraisjr: reloadHaproxy func could be removed and it's single line placed here.

pkg/controller/controller.go
```diff
@@ -166,56 +181,87 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
...
-	tmpf := "/etc/haproxy/new_cfg.erb"
-
-	err := ioutil.WriteFile(tmpf, data, 644)
-	if err != nil {
-		glog.Warningln("Error writing rendered template to file")
-		return err
+func (haproxy *HAProxyController) rewriteConfigFiles(data []byte) (string, error) {
+	// Include timestamp in config file name to aid troubleshooting. When using a single, ever-changing config file it
+	// was difficult to know what config was loaded by any given haproxy process
+	timestamp := time.Now().Format("-060102-150405.0000")
```
jcmoraisjr: What about four digits to the year and three to the milisseconds?

pkg/controller/controller.go
```diff
@@ -166,56 +181,87 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
...
+		}
+
+		// Validate haproxy config
+		if err := checkValidity(configFile); err != nil {
+			return "", err
 		}
 	}
-	err = os.Rename(tmpf, configFile)
+
+	haproxy.removeOldConfigFiles(haproxy.configFileSuffix)
```
jcmoraisjr: removeOldCOnfigFiles func has already access to this argument.

pkg/controller/controller.go
```diff
@@ -166,56 +181,87 @@ func (haproxy *HAProxyController) OnUpdate(cfg ingress.Configuration) error {
...
+	return configFile, nil
+}
+
+func (haproxy *HAProxyController) removeOldConfigFiles(suffix string) error {
+	files, err := ioutil.ReadDir(haproxy.configDir)
 	if err != nil {
-		glog.Warningln("Error updating config file")
 		return err
 	}
```
jcmoraisjr: error isn't being used, so log a warningf("...: %v", err) here and return without an argument

pkg/controller/controller.go
westse: @jcmoraisjr Do you mean a formate like `-0102-150405.000`, where year and nanoseconds are lost but month, day, hour, second, millisecond are retained? I think that will be fine since there is low risk of overwriting based on either of those.

westse:
@jcmoraisjr I've incorporated each item you mentioned, except the timestamp format (see my question). So let me know how it looks. Thanks.

pkg/controller/controller.go
jcmoraisjr: I mean following this convention: `-20060102-150405.000`.

westse:
Ok, I've fixed the timestamp format now as well.

jcmoraisjr:
Just a few more issues:

* please use zero old files as default
* check for negative numbers, converting them to zero
* you should use the pointer itself on flags.Int() otherwise the value won't be updated - see reloadStrategy usage
* when zero is used the idea is to leave the timestamp empty - it's actually filling the timestamp
* removeOld shouldn't be used if maxOldFiles is zero - it's removing the actual config file.

westse:
Oops, I forgot to implement the "zero means no timestamp" part. Good catches - the code has been updated. Also feel free to tweak as you see fit.

pkg/controller/controller.go
```diff
@@ -109,6 +108,11 @@ func (haproxy *HAProxyController) UpdateIngressStatus(*extensions.Ingress) []api
 func (haproxy *HAProxyController) ConfigureFlags(flags *pflag.FlagSet) {
 	haproxy.reloadStrategy = flags.String("reload-strategy", "native",
 		`Name of the reload strategy. Options are: native (default), reusesocket or multibinder`)
+	haproxy.maxOldConfigFiles = flags.Int("max-old-config-files", 0,
+		`Maximum old haproxy timestamped config files to allow before being cleaned up. A value <= 0 indicates a single non-timestamped config file will be used`)
+	if *haproxy.maxOldConfigFiles < 0 {
+		*haproxy.maxOldConfigFiles = 0
+	}
```
jcmoraisjr: This is just the initialization, the args wasn't parsed yet, so this if should be at OverrideFlags which happens after the parse.

Anyway afaics you are using >0 or <=0 everywhere so it's pointless to check it's value. I think it's safe to just remove it.

pkg/controller/controller.go
```diff
@@ -242,12 +246,14 @@ func (haproxy *HAProxyController) removeOldConfigFiles(suffix string) error {
 
 	matchesFound := 0
 	for _, f := range files {
-		if !f.IsDir() && strings.HasPrefix(f.Name(), haproxy.configFilePrefix) && strings.HasSuffix(f.Name(), suffix) {
+		if !f.IsDir() && strings.HasPrefix(f.Name(), haproxy.configFilePrefix) && strings.HasSuffix(f.Name(), haproxy.configFileSuffix) {
 			matchesFound = matchesFound + 1
-			if matchesFound > haproxy.maxOldConfigFiles {
+			if matchesFound > *haproxy.maxOldConfigFiles {
 				filePath := haproxy.configDir + "/" + f.Name()
 				glog.Infof("Removing old config file (%v). maxOldConfigFiles=%v", filePath, haproxy.maxOldConfigFiles)
```
jcmoraisjr: Missing a `*` on maxOldConfFiles or you'd output the memory address.

westse:
thanks, updated

pkg/controller/controller.go
ausmith: I see the difference in approach now. My hope was that all reload strategies would be capable of at least storing copies of their config at reload time (by copying the file off before putting it into place), but understand that is simply adding files that provide limited value. Look forward to verifying this code when the new version gets built and tagged!

jcmoraisjr:
Lgtm, thanks for the updates!

PR #124, opened 2018-03-16
Workaround haproxy hitless reload limit on file descriptors

westse:
The HAProxy limit is related to the Linux SCM_MAX_FD=253 cap. Not really documented, but is [loosely mentioned here](https://www.haproxy.com/blog/truly-seamless-reloads-with-haproxy-no-more-hacks). Before the changes in this commit the limit is hit very quickly in haproxy-ingress because of its use of so many sockets which are represented on Linux as files. This commit removes the `[socket]<-frontend->backend` chain for each Server. Instead there is a single frontend for both http and https (mode is http, not tcp as before) with directives matching each server.

The following features were not retained in order to simplify the refactor, and decisions need to be made on if/how to restore them.
1. `HTTPStoHTTPPort, UseHostOnHTTPS, HTTPSLogFormat`.
    1. **Reason**: Unclear how to support it given the unified http/https frontend. Maybe could have separate frontends for http and https and use go sub-template to unify logic
1. `Alias support`
    1. **Reason**: Took out because of unnecessary duplication. Ideas to restore this include:
        1.  putting common logic in a sub template, [as done here](https://github.com/kubernetes/ingress-nginx/blob/41cefeb1781a3cf16ab6b7e13fe921611726d7b2/rootfs/etc/nginx/template/nginx.tmpl#L581)
        1. [Take the approach that Nginx has done](https://github.com/kubernetes/ingress-nginx/blob/41cefeb1781a3cf16ab6b7e13fe921611726d7b2/rootfs/etc/nginx/template/nginx.tmpl#L415) with using the `server_name` directive.
1. `PassthroughBackend`.
    1. **Reason**: Unclear how to support it because https frontend has mode=http. One idea to restore this is to use a tcp frontend like before, but instead of sending to N `backend/socket/frontend` chains, it could send to a single `backend/socket/frontend` chain for the non-passthrough case, and directly to the backend in passthrough case.
1. `$authSSLCert`, `server.CertificateAuth.AuthSSLCert` and related.
    1. **Reason**. Took out because is a per-server setting, but now there is only one https frontend. One idea to restore this is to add multiple `crt-file` bind options to the `bind *:443` line.

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -100,172 +100,13 @@ backend {{ $backend.Name }}
 {{- end }}
 {{- end }}
 
-{{- $hasHTTPStoHTTP := or (gt $cfg.HTTPStoHTTPPort 0) $cfg.UseHostOnHTTPS }}
-{{- $reuseHTTPPort := eq $cfg.HTTPStoHTTPPort 80 }}
-
 ######
-###### HTTP frontend
+###### HTTP(S) frontend
 ######
-{{- if $cfg.UseHostOnHTTPS }}
-backend http-backend
-    mode http
-    server http-front unix@/var/run/haproxy-http.sock send-proxy-v2
-{{- end }}
 frontend httpfront
+    mode http
     bind *:80{{ if $cfg.UseProxyProtocol }} accept-proxy{{ end }}
-{{- if and (gt $cfg.HTTPStoHTTPPort 0) (not $reuseHTTPPort) }}
-    bind *:{{ $cfg.HTTPStoHTTPPort }}{{ if $cfg.UseProxyProtocol }} accept-proxy{{ end }}
-{{- end }}
-{{- if $cfg.UseHostOnHTTPS }}
-    bind unix@/var/run/haproxy-http.sock accept-proxy
-{{- end }}
-    mode http
-
-{{- range $server := $ing.HAServers }}
-{{- if isWildcardHostname $server.Hostname }}
-    acl host-{{ $server.HostnameLabel }} hdr_reg(host) {{ hostnameRegex $server.Hostname }}
-{{- else }}
-    acl host-{{ $server.HostnameLabel }} hdr(host) {{ $server.Hostname }} {{ $server.Hostname }}:80
-{{- end }}
-{{- if ne $server.Alias "" }}
-{{- if isRegexHostname $server.Alias }}
-    acl alias-{{ $server.HostnameLabel }} hdr_reg(host) '{{ aliasRegex $server.Alias }}'
-{{- else }}
-    acl alias-{{ $server.HostnameLabel }} hdr(host) {{ $server.Alias }} {{ $server.Alias }}:80
-{{- end }}
-{{- end }}
-{{- end }}
-
-{{- if $hasHTTPStoHTTP }}
-    acl from-https hdr(x-forwarded-proto) https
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Hi, nice to see you back here and thanks for the nice job and detailed explanation. I really appreciate refactors like this but this need to be done as an option to the user. What about a configmap option where you choose the name or the version of the template you want to use? This has two big advantages which is starting right now and zero issues with backward compatibility. There is a minor disadvantage which is the double of the work on new implementations that need to change the template.

westse:
While I have full confidence we can produce a single template that keeps all features and avoids the use of so many file descriptors, I recognize this pull request is incomplete because it drops features. I had plans to restore those dropped features but ran out of justification. So I like your idea of the config map option because then the community can help refine it. I'll defer to you on how that is done.

Thanks for being open to these refinements. Our company relies on haproxy-ingress heavily so we want to see it become even better. In particular the "hitless reload" capability is important to us because the multibinder approach (which we used previously) caused issues with occasional stale config. We use Kubernetes and haproxy-ingress for temporary development environments and often have thousands of config reloads per hour, and even occasional stale config issues become commonplace.

ausmith:
I agree with @jcmoraisjr that this level of change would have to be behind a configmap option until it has feature parity (I think this was already agreed upon, so I am being redundant). 

I am surprised that your experience with `multibinder` has not been as seamless as our own, and from your description I suspect this is because we are not reaching the volume of churn your environment seems to experience. @westse do you have an idea what volume of ingress change per second/minute/hour began to see issues with `multibinder` reloads? We have observed general ingress staleness when kube ingress config results in an invalid haproxy configuration in the environment (`multibinder` does not let the dead haproxy process take over the host port in this situation), how often was a situation like that the root cause vs stale config due to too much churn? Thanks for any additional information you can provide here so we can prepare for such a possibility.

westse:
@ausmith regarding the rate of ingress config changes, we have spurts of tens per minute, and sometimes up to hundreds and even thousand per hour.

multibinder generally has worked very well but have noticed occasional stale config even after config reload was apparently successful. . After the following changes we have not (as yet) seen recurrence:
1. Use the latest haproxy-ingress code with `--reload-strategy=reusesocket` hitless reload
1. haproxy.tmpl refactor to workaround "hitless reload" limitation (this PR).
1. Ensure each haproxy process has its own config file (PR https://github.com/jcmoraisjr/haproxy-ingress/pull/123)
1. Ensure each all existing haproxy processes gracefully shutdown, not just the most recent (PR https://github.com/jcmoraisjr/haproxy-ingress/pull/122)

Unfortunately changing all of these variables means it is not definitive which one helped. 

For completeness I should also note we have refined our haproxy-liveness probes to detect and kill old haproxy processes that have are still servicing long-running connections, such as for websockets. We are confident these not the source of stale config as we have verified all requests that fail due to stale config are not serviced by old processes.

ausmith:
@westse this PR is not the right place for the information, but I would be interested in how you went about refining "our haproxy-liveness probes to detect and kill old haproxy processes that have are still servicing long-running connections, such as for websockets". 

That aside, I am definitely interested in where these "hitless" reload config changes go. As mentioned elsewhere, I've developed a small testing suite for exercising the feature set/haproxy configuration that we rely upon so will be watching for new versions to put through that testing.

jcmoraisjr:
I've just created #130 which should fix this issue in a backward compatible way. Please have a look if it works with your workload.

jcmoraisjr:
@westse I'm closing this PR on behalf of #130, feel free to ping back if you have any issue regarding the current approach.

PR #129, opened 2018-03-23
Add support for Http 2 connections

saary:
Update haproxy.tmpl to accept http2 and http 1.1 connections.

Allow H2 connections to be accepted in addition to HTTP 1.1 connections.
Information about HAProxy 1.8 HTTP 2 support is available [here](https://www.haproxy.com/blog/whats-new-haproxy-1-8/).

A very small change providing nice performance wins.


jcmoraisjr:
Thanks!

PR #133, opened 2018-03-28
Avoid reload when shrinking backend slots

jcmoraisjr:
This changes the condition used to reload the server because of resizing of the backend slots. If the slots need to be shrinked the reload is skipped. They will be shrinked anyway if the server need to be reloaded for any other reason.


jcmoraisjr:
@aiharos as the original author, this sounds good or perhaps you see any drawback?

PR #138, opened 2018-04-18
Change equality of drain-support and dynamic-scaling

jcmoraisjr:
Changing the implementation of equality and partial equality of ingress config structs

* Changes are related with `dynamic-scaling` and `drain-support`
* Remove changing comparison of `Endpoint.Target` field
* Move from `DeepEqual` to `.Equal` func on ingress config struct, removing issues with internal states of k8s objects and unsorted arrays
* Usage of a cloned object `configCopy` for partial equality on behalf of visibility of what isn't being compared
* A new func `DeepEqualUnsorted` used to compare unsorted arrays

### DIFF
--- a/pkg/common/ingress/types_equals.go
+++ b/pkg/common/ingress/types_equals.go
@@ -259,19 +259,6 @@ func (e1 *Endpoint) Equal(e2 *Endpoint) bool {
 	if e1.Draining != e2.Draining {
 		return false
 	}
-
-	if e1.Target != e2.Target {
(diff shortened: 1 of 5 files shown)


jcmoraisjr:
@brianloss changed how config and endpoints are compared on drain-support, perhaps you want to have a look.

jcmoraisjr:
Image of this PR: `quay.io/jcmoraisjr/haproxy-ingress:v0.6-108a405`. Note that this is a temp image which should be removed as soon as the first v0.6 snapshot is tagged.

Issue #114, opened 2018-03-12
Server limiting connections to 4000 even with max-connections set to 8000 or 50000


dobesv:
This morning I saw that with 4 HAProxy instances the ingress accepted a maximum of 16000 connections.  I tried bumping the max-connections setting to 50000 but it still would not accept more than 4000 per instance.  I am running more instances but I'm puzzled about why that setting isn't working.  

Could it be something to do with all the unix sockets being used to route the connections?

jcmoraisjr:
Perhaps you also need to tune the [maxconn](http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4-maxconn) of your frontends. There isn't an option on HAProxy Ingress to do this, and this is indeed an useful feature, so you can: 1. wait a few more days so I add this to the next snapshot and you can test it; 2. fix the haproxy.tmpl or mount a new template with a configmap volume patching the frontend or default session.

dobesv:
For now I'm just running a bunch of extra HAProxy instances, which seems to resolve the problem for now.  Assuming I don't run into any other problems I can wait for an update to the ingress.

styk-tv:
@dobesv would be interesting to compare notes are you on aws?  I'm attempting to climb to 50k/s http ingress-service hits for a small http rest api with sla requirement of 100ms

dobesv:
I'm on the kubernetes slack if you want to chat.

PR #143, opened 2018-04-23
Add maxconn to defaults as well as global

dobesv:
This applies the same `maxconn` settings to each proxy in addition to
the global `maxconn` setting.

Could fix https://github.com/jcmoraisjr/haproxy-ingress/issues/114


jcmoraisjr:
Merged, thanks!

Issue #128, opened 2018-03-23
Option to have "maxconn" property to be configured per ingress resource as a new annotation 


swapnil1988:
Currently there are configurations that can be set at ingress resource level in form of various annotations. But one important property of maxconn per backend server which is supported by haproxy but not available as an option to set as an annotation at ingress resource level.

Annotation can be something like

`ingress.kubernetes.io/max-connections : 1 `

**example expected haproxy configuration with maxconn:**
```
backend exampleAppBackEnd
    balance roundrobin
   server app1 10.0.0.1:8080  maxconn 1
   server app2 10.0.0.2:8080  maxconn 1
   server app3 10.0.0.3:8080  maxconn 1
```

jcmoraisjr:
Hi, sure, already a v0.6 backlog.

swapnil1988:
Thanks @jcmoraisjr

jcmoraisjr:
Added in the current snapshot of v0.6. Closing. Doc [here](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/README.md#connection). Let me know if you have any problem updating this same issue.

swapnil1988:
Thanks @jcmoraisjr it worked for me :)

PR #148, opened 2018-05-06
Connection limits and timeout config

jcmoraisjr:
Adding annotations and configmap option to configure connection limits and timeout to backend servers.


jcmoraisjr:
Implement #128 .

Issue #147, opened 2018-05-01
Unable to override wildcard hostname with specific hostname


yunruiwatts:
This is regarding the wildcard hostname support that was added with #32. It appears that as a consequence of [sorting the servers by their hostname](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/pkg/common/ingress/controller/controller.go#L712) the wildcard will always sort before specific - `*` comes before alphanumeric character when sorting strings. As a result, it is not possible to have a specific rule beat out a wildcard rule.

Ideally we should be able to do the same thing that we can do with DNS - define `*.example.com` to go to one location, but allow `api.example.com` to be overridden and go to a different location.

I'm not sure the best way to do this - a quick and easy way to do this would be to sort in reverse so that `*` would be used last, but that would affect everyone even if they don't have any wildcard rules. Another option would be to find or implement a sort that sorts alphanumeric characters ahead of other characters. One other option that should work but less is than ideal (but is potentially faster than implementing a new sort) is to simply check if a hostname has a `*` in it, and if so sort it after the other hostname unless the other hostname also has a `*` in it.

Also, I'm not sure how this would affect [regex hostnames](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/pkg/controller/template.go#L77). My guess is that those should probably also go after non-regex, non-wildcard hostnames, so ideally whatever solution we use would do that.

yunruiwatts:
Oh, I forgot to mention that I'm happy to do a pull request with whatever solution we decide would be best, but I wanted to have a discussion about what solution would be best first.

PR #149, opened 2018-05-06
Fix precedence on hosts/alias without regexp

jcmoraisjr:
Give precedence on hosts without wildcard and alias without regex.

Implements #147.


yunruiwatts:
I just checked this in our system, and the cfg it generates does have the wildcard rules at the bottom, as expected.

jcmoraisjr:
Thanks for evaluating! Merging.

PR #172, opened 2018-06-09
Add multi process and multi thread support

jcmoraisjr:
Add new configmap options to support multi process and/or multi thread configuration.

* `nbproc-ssl`: number of processes dedicated to ssl-offloading; plain http requests are handled on another dedicated process.
* `nbthread`: number of threads for each process.

Todo:

* [x] core implementation
* [x] fix acl ssl-offload
* [x] docs and warnings


Josh-West-Bose:
We're very much looking forward to this :-)

PR #179, opened 2018-06-22
Fix redirect https if path changed with rewrite-target

### LINKED ISSUE #178: When both ssl-redirect and rewrite-target are used for a given location, client is redirected to the wrong location.
When ssl-redirect is true and rewrite-target is set for a given location, the client is redirected to a wrong location.  Specifically, the client is redirected to https but with the rewritten path rather than the original path.  This is usually not the intended behavior.

For example:

Given:
- an ingress path of `/some/path/`;
- a rewrite-target of `/`;
- and ssl-redirect set to `true` (either globally or just for the specific ingress);

When a client attempts to get `http://hostname/some/path/with/deeper/components` the client will be redirected to `https://hostname/with/deeper/components`.  However, the client should have been redirected to `https://hostname/some/path/with/deeper/components`.

One way to get the correct redirection under these conditions is to include `from-https` in the `if` condition of the resulting `reqrep` line in the configuration.

For example:

```
reqrep ^([^\ :]*)\ /some/path//?(.*$) \1\ /\2 if host-hostname { var(txn.path) -m beg /some/path/ }
```

should be

```
reqrep ^([^\ :]*)\ /some/path//?(.*$) \1\ /\2 if from-https host-hostname { var(txn.path) -m beg /some/path/ }
```

Note that `from-https` should only be include when all of the above conditions are met.  If any of the conditions are not met, then `from-https` should not be included in the `reqrep` line.

jcmoraisjr:
Fix redirect-https if used with rewrite-target.

Option 1 (current): use an acl so reqrep will only execute if redirect https isn't executed
Option 2: write the whole new location using `txn.path` instead of redirect https

Fixes #178 .

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy.tmpl
@@ -334,9 +334,9 @@ frontend httpfront-{{ if $isShared }}shared-frontend{{ else if $isDefault }}defa
 {{- $rewriteTarget := $location.Rewrite.Target }}
 {{- if ne $rewriteTarget "" }}
 {{- if eq $rewriteTarget "/" }}
-    reqrep ^([^\ :]*)\ {{ $location.Path }}/?(.*$) \1\ {{ $rewriteTarget }}\2 if {{ $server.ACLLabel }} { var(txn.path) -m beg {{ $location.Path }} }
+    reqrep ^([^\ :]*)\ {{ $location.Path }}/?(.*$) \1\ {{ $rewriteTarget }}\2 if {{ if $location.SSLRedirect }}from-https {{ end }}{{ $server.ACLLabel }} { var(txn.path) -m beg {{ $location.Path }} }
 {{- else }}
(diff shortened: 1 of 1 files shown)


qwylz:
Thanks for quickly addressing this (and thanks for a generally great ingress, too!).

I had a look at the patched template.  It fixes the issue when the specific ingress has ssl-redirect enabled, but not when that is the default.

Don't we need something like the following (only showing one of the modified reqrep lines)?

```
reqrep ^([^\ :]*)\ {{ $location.Path }}/?(.*$) \1\ {{ $rewriteTarget }}\2 if {{ if $server.SSLRedirect }}from-https {{ else }}{{ if $location.SSLRedirect }}from-https {{ end }}{{end}}{{ $server.ACLLabel }} { var(txn.path) -m beg {{ $location.Path }} }
```

jcmoraisjr:
Indeed. What about `server or location` inside the same `if` statement?

jcmoraisjr:
> > ... but not when that is the default 

> Indeed. What about server or location inside the same if statement?

Reverting ... `$location` actually inherit the default configuration. Did you see any issue with default ssl-redirect config?

qwylz:
Ha!  You're right.  I didn't realize that and was making assumptions based on other parts of the template file that tested whether ssl redirection was enabled at the server level and location level separately.  84fd631 works correctly.  :-)

PR #183, opened 2018-07-03
Fix ssl-passthrough annotation

Scukerman:
ACLLabel is not expected to be here.
This commit (a80d2c648369c7a58ae32cc278e9d8cbbb588669) breaks the `ingress.kubernetes.io/ssl-passthrough` annotation.

/milestone v0.6
/label fix


jcmoraisjr:
Thanks! Merging. Just to let you know this should also support wildcard, I'll take care of this before tag beta2.

Scukerman:
Thanks. That was a quick fix to keep it working on my side.

PR #190, opened 2018-07-20
Add http/https port map configuration

Nefelim4ag:
On some haproxy-ingress setups several instances of haproxy can coexists
on same node, to avoid port mapping clashes, add configuration options
to allow change ports

PR #208, opened 2018-08-24
Remove ports 80/443 from reserved

Nefelim4ag:
As default http/https ports can be overriden by ConfigMap
remove them from reserved


Nefelim4ag:
In continuation of #190

jcmoraisjr:
Make sense, merging, tks!

PR #219, opened 2018-09-23
Change dh param size

jvanbruegge:
The SSL test is [capped at B](https://www.ssllabs.com/ssltest/analyze.html?d=janco-it.de) with the default settings. Changing dh-param in the configmap to `2048` [results in an A+](https://www.ssllabs.com/ssltest/analyze.html?d=cerberus-systems.com)


jcmoraisjr:
It's a fair upgrade although this will break some old clients. I'll provide a backward compatibility note. Tks!

Issue #210, opened 2018-08-31
spike of "secret not found" warnings


zapman449:
version v0.5-beta.3

Brought some nodes online yesterday, and they went healthy in the ASG/ NLB-TG, then they went unhealthy, then they came healthy again.  In digging into that, I noticed that there was a huge spike of `W0830 12:08:13.595614       6 config.go:146] secret not found: L.......` which corresponded to the `ssl-dh-param`.  

Huge spike means 236 warnings in 1 second in each of 7 pods.

Anyone seen something like this?

jcmoraisjr:
Hi, this misbehavior can be partially explained eg on a race starting the controller, changing the secret of the dh param or during an outage of the api server. So some questions: this happened only once or every now and then? Only once and just after start the controller or after some minutes or hours after the start? This log output is only a consequence of something else and wouldn't explain an outage of the controller but only that a customized DH wouldn't be used - so, was there something else in the logs beyond the `secret not found` message?

zapman449:
We expose haproxy-ingress as a daemonset, and at the time there were around 45 worker nodes.  So, the instances were pretty fresh.

And that message is the only one in the 10 minutes around the issue.  (technically, there's one "service does not have any active endpoints" message, but that's not relevant)

jcmoraisjr:
I can only reproduce this behavior removing the secret with the dh file, which would also warn every time a configuration change. Please double check the secret name in the configmap option, the secret content, the dhparam.pem key in the secret. If everything sound to be fine and you have these messages only once just after start the controller, perhaps we have a race condition here. I'm evolving some ssl stuff right now and will have a closer look into this.

Regarding the spike, it happens because this configuration is misplaced and will be fixed in v0.7, thanks for pointing this out.

Regarding the behavior, these warnings are just telling a secret couldn't be found and a customized dh won't be used. Perhaps this could generate an invalid config file but the logs would have a lot of evidences if HAProxy couldn't be started for any reason.

The failures continue to happen? Perhaps you can add `--v=5` to the command-line options (this is noisy) which can help us understand what's going on here. Every single detail can help. Feel free to mail me with the logs and other configuration details.

ausmith:
I believe this can be closed now?

jcmoraisjr:
@zapman449 were the warnings generated by a misconfiguration or do you still have this problem configuring a valid dh-param secret?

jcmoraisjr:
[...]

PR #220, opened 2018-09-24
Update ssl-dh-param documentation for clarity

ausmith:
As documented in https://github.com/jcmoraisjr/haproxy-ingress/issues/210, we were seeing the DH parameter configuration spam our logs. This turned out to be an issue with our configuration where we included the certificate in the `ssl-dh-param` instead of having it reference a secret with that data. 

This documentation PR hopes to help others from making the same mistake by providing an example of the secret and specifying that the secret should be referenced in your haproxy-ingress configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,21 +615,26 @@ http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#tune.ssl.default-dh
 Define DH parameters file used on ephemeral Diffie-Hellman key exchange during
 the SSL/TLS handshake.
 
+When stored locally, the DH secret may look like:
+
+```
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Hi, great! A few suggestions:

* Leave the link of the HAProxy doc to the bottom
* Say somewhere that the secret has the base64 encoding of the dh-param - or perhaps add also the command-line syntax: create a file named `dhparam.pem` and type `kubectl create secret generic ingress-dh-param --from-file dhparam.pem`

ausmith:
@jcmoraisjr how do those changes look?

jcmoraisjr:
LGTM, thanks! Merging.

PR #223, opened 2018-09-29
Use a multi pem file instead the ssl dir on shared frontend

jcmoraisjr:
A small improvement on v0.7 after adding client cert for backends. All crt/key pairs are added to the ssl directory. This PR create a new `+default.pem` file with all and only the crt/keys that should be used in the shared frontend.

This will change again on v0.8 after create an array of frontends, each one with their array of servers/crt/key.


gfeun:
Hello, 
This seems to break non default TLS enabled sites. 
With this patch only the default certificate is loaded.
Reading through https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.1-crt , it is not clear whether it is possible to put different certificates and keys inside a single PEM file. The example given is a concatenation of a cert and a key for a single common name, not various certificates and keys.
Our TLS configuration works on v0.6.beta3 and not v0.7.
Thank you

jcmoraisjr:
Hi, I probably misunderstood the docs and also the result of my tests. I'll have a closer look. Thanks for pointing this.

PR #226, opened 2018-10-10
Fix documentation for auth-tls client

jiphex:
The documentation (README.md) states that the secret used for TLS client certificate authentication requires a key named `ca` containing the certificate data, when in fact it [needs to be `ca.crt`](https://github.com/jcmoraisjr/haproxy-ingress/blob/ab62cfe38c8766a00a261fc590523e9f0400fc9d/pkg/common/ingress/controller/controller.go#L580).

I'm doing some testing with client certificates at the moment.


jcmoraisjr:
Indeed, thanks! Merging.

Issue #240, opened 2018-10-23
sort-backends option not actually respected


gjtempleton:
### Versions tested with: 
v0.6-beta.3 and v0.6-beta.6

### Expected behaviour:
Sort-backends flag is respected and if unset backends are randomised in order

### Observed behaviour:
Sort-backends flag is respected within the internal data _until_ passed to the templating of the HAProxy config flag. This means that servers under backends are always sorted alphabetically. This has led to unbalanced connection distribution in our larger clusters.

### Detail:

The sort-backends flag documented here:
https://github.com/jcmoraisjr/haproxy-ingress#sort-backends

isn't actually respected, as it is finally passed into a range function in the template.

This has its behaviour documented here: https://golang.org/pkg/text/template/#hdr-Actions and immediately sorts the backends by their key:

```
{{range pipeline}} T1 {{else}} T0 {{end}}
	The value of the pipeline must be an array, slice, map, or channel.
	If the value of the pipeline has length zero, dot is unaffected and
	T0 is executed; otherwise, dot is set to the successive elements
	of the array, slice, or map and T1 is executed.
```

I have a tested workaround making some changes at both the template and `dynconfig` layers to handle this. MR imminent.

jcmoraisjr:
Hi, I've just pushed #247 which should solve this problem. There are some few tests pending before merge and a few other implementations before the next v0.7 snapshot. I'll update this issue in the next few days as soon as a new version is tagged.

There is a lot of changes in this PR to be merged to v0.6, so perhaps v0.6 stable will be released without this fix. I'm however considering a v0.6.1 after confirm that everything that was touched is still working.

gjtempleton:
Hi, sorry for the delay in getting back to you.

Upgrading hasn't really been a blocker or priority for us as we've been happily running my patch internally since I raised this issue as we don't use the dynamic backends behaviour I neglected to test in #241 .

I wouldn't worry about putting in the work to back-port the fix, certainly not on our account, however I'll look at testing and rolling out the latest 0.7 release in our environment.

Thanks so much for the quick responses and work put in to fix this.

Closing issue.

PR #241, opened 2018-10-23
Respect sort-backends flag

gjtempleton:
Resolves the issue discussed in #240 

* Adds a new field to `HAProxyBackendSlots` struct called `FullSlotIndices` of type []string - this is used purely to track the order of processing of endpoint targets

* Appends new targets to `FullSlotIndices` as each upstream is processed and added to the backend

* Updates the template to range over `FullSlotIndices` for each backend rather than `range $target, $slot := $BackendSlots.FullSlots` as this causes the map to become ordered alphabetically

* Tested in a local fork and build with both the default behaviour of `sort-backends=false` and explicitly setting `sort-backends=true`

### DIFF
--- a/pkg/controller/dynconfig/dynconfig.go
+++ b/pkg/controller/dynconfig/dynconfig.go
@@ -18,12 +18,13 @@ package dynconfig
 
 import (
 	"fmt"
+	"reflect"
+	"sort"
+
 	"github.com/golang/glog"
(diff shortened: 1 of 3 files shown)


gjtempleton:
Hi, no worries, and thanks for the feedback.

That's my bad, I didn't test it with dynamic backends as we don't make use of that internally, I'll grab some time today to have another look.

Thanks for the feedback!

PR #273, opened 2018-12-28
[WIP] - Add optional client authentication verification

rikatz:
When using optional client certificate authentication, HAProxy should not redirect users to the errorpage (495, 496 or whatever the user have customized), and must continue the communication flow, as this authentication is Optional.

TODO: 

* Improve go templating for this conditional loop
* Document the annotation ``ingress.kubernetes.io/auth-tls-verify-client: optional`` and improve, as if the 'optional' authentication is used and the user provides a certificate, but with invalid CA, the error495 must be presented to him.


rikatz:
Will make a better PR, into the right fork and branch

PR #275, opened 2019-01-03
Improve optional client cert auth

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@serpro.gov.br>

Improve optional client cert auth. 

The following will happen:

* When the annotation `ingress.kubernetes.io/auth-tls-verify-client` is set to `optional` or `on`, the following validations will be made

* If the certificate is invalid, expired, not recognized by any CA configured as the `auth-tls-secret` the error 495 will be used (the customized one, or the default one)

* If the certificate isn't presented by the client and it's required, using the `auth-tls-verify-client: "on"` the error 496 will be used (the customized one, or the default one)


jcmoraisjr:
Nice catch, thanks! Sounds very good, merging.

PR #278, opened 2019-01-07
Add support to customize haproxy logformat

primeroz:
Supported format from haproxy are : rfc5424 or rfc3164

I personally need this to make log forward work with filebeat which expect rfc3164 format rather than 5424 and currently print out errors like 

```
haproxy-ingress-7d59695959-q5sjc haproxy-beat 2019-01-07T13:35:02.759Z  ERROR   [syslog]        syslog/input.go:131     can't not parse event as syslog rfc3164 {"message": "<134>1 2019-01-07T13:35:02+00:00 haproxy-ingress-7d59695959-q5sjc ingress 44 - - 10.244.5.0:48548 [07/Jan/2019:13:35:02.165] httpfront-shared-frontend~ aaa-aaaa-web-staging-8000/10.244.13.99:8000 0/0/1/593/595 404 9085 - - ---- 4/2/1/1/0 0/0 \"GET /health/ HTTP/1.1\"\n"}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,7 @@ The following parameters are supported:
 |`[0]`|[`stats-ssl-cert`](#stats)|namespace/secret name|no ssl/plain http|
 |`[0]`|[`strict-host`](#strict-host)|[true\|false]|`true`|
 ||[`syslog-endpoint`](#syslog-endpoint)|IP:port (udp)|do not log|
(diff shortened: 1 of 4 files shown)


jcmoraisjr:
Hi, sorry about taking so long to have a look at this PR. Merging, thanks!

Note that this same configuration can be done on v0.7 using `config-global`. A second `log` declaration will overwrite a previous one.

PR #287, opened 2019-01-18
Add backend health and agent checks.

brianloss:
It appears there was some existing health check code that was copied from the nginx ingress controller. I overwrote that with new code that supports the options needed for HAProxy.

Note that to have an agent control weights in haproxy, you can't really have the default weight of 1, since any reduction in the weight by the agent will cause the weight to be 0 (the agent reports weight changes as a percentage), so I changed the default in the bluegreen code. Hopefully that doesn't cause a problem.


brianloss:
Rebased to master, but would like this to go in the 0.7 release if that's possible.

brianloss:
No worries about the delay. Thanks for the review, comments, and patch!

I figured that was the case on the reuse, and just pointed it out in case my assumption was wrong and I was deleting something that was actually used.

I will make the suggested changes, but have 2 questions...
1. Interval already is a string. I was going to change the readme to say `time with suffix` as appears in other places, but in the code I believe it is a string everywhere. Did I miss something?
2. For defaulting the port, it appears $backend.Port is an IntOrString so the port could be specified as a string. If it is a string, I think I can look through the associated service and find the spec with the matching name, but that's a lot to try and do in the template. Any advice on where to put that logic?

jcmoraisjr:
> Interval already is a string.

Now I see. The problem apparently is that the annotation is read with `GetIntAnnotation`.

> For defaulting the port, it appears $backend.Port is an IntOrString so the port could be specified as a string. If it is a string, I think I can look through the associated service and find the spec with the matching name, but that's a lot to try and do in the template. Any advice on where to put that logic?

This is already implemented and the result is on the `Endpoint` object which is used on dynconfig to assembly `slot.Target`. This can be implemented either parsing `slot.Target` on the template or adding `Endpoint` to `HAProxyBackendSlot`.

brianloss:
> > Interval already is a string.
> 
> Now I see. The problem apparently is that the annotation is read with `GetIntAnnotation`.

Ah, forgot about that. Will fix to pull as a string annotation instead.

> 
> > For defaulting the port, it appears $backend.Port is an IntOrString so the port could be specified as a string. If it is a string, I think I can look through the associated service and find the spec with the matching name, but that's a lot to try and do in the template. Any advice on where to put that logic?
> 
> This is already implemented and the result is on the `Endpoint` object which is used on dynconfig to assembly `slot.Target`. This can be implemented either parsing `slot.Target` on the template or adding `Endpoint` to `HAProxyBackendSlot`.

I had seen that, but I don't believe I can use that for the [server-template case](https://github.com/jcmoraisjr/haproxy-ingress/blob/b42d9069415c16dc0eccc7cb0867fcf42c84d3ba/rootfs/etc/haproxy/template/haproxy.tmpl#L162), which uses the same backend template. I could document that agent-port is required, and only add the agent options if the agent-port annotation is included. Would that be better?

jcmoraisjr:
> I had seen that, but I don't believe I can use that for the server-template case, which uses the same backend template. I could document that agent-port is required, and only add the agent options if the agent-port annotation is included. Would that be better?

Indeed, yes. I'd follow the agent-port required way. Improvements could be made on the new controller codebase.

jcmoraisjr:
Thanks! Merging.

PR #288, opened 2019-01-19
Allow the syslog tag to be specified in the config map.

brianloss:


jcmoraisjr:
Hi, welcome back and thanks again for your support! Two minor changes:

* Does your use-case need to use the default value in the template? Otherwise I'd suggest you to move default to config.go, following the convention.
* Please rebase the PR to master, I'll cherry pick before tag the next beta.

brianloss:
No problem. Thanks for pulling these changes in. It definitely makes it easier if I can get to the point where I'm not working off a fork.
Requested changes have been made. No need for the default value to appear in the template, so I moved it to config.go.

jcmoraisjr:
Merging, thanks!

I'd also propose to postpone this to v0.8.

Why I was merging to v0.7? This implementation is simple enough to be considered to be included. Why not anymore? This same feature can be configured with `config-global: log-tag xxx`.

Issue #274, opened 2019-01-01
Create a real HAProxy config representation


jcmoraisjr:
A new ingress synchronizer and its dependencies is under active development. This is the most relevant change since the very first version of HAProxy Ingress controller. Most of the internals are being rewritten from scratch.

Background: HAProxy Ingress started [here](https://github.com/kubernetes/ingress/pull/214) as another implementation of the so called "Ingress Controllers". The reusable code grew in a wrong way and also didn't match exactly what an HAProxy instance need. Because of that there are some workarounds in place that are making it difficult to evolve HAProxy Ingress in a fast and reliable way.

The redesign of the internals are being built in a backward compatible way. Please help us making tests as soon as the first v0.8 snapshot is deployed, opening new issues if you find any regression. Subscribe to this issue to get notified. The rule is simple: if something doesn't work like in v0.7 you found a regression - this should be fixed either changing the code or updating the doc of the compatibility changes.

I'll try to start a WIP PR in a few days or weeks so you can see what's being done and can give some feedback. Planning to start snapshots late feb/19 and perhaps beta version 3 or 4 weeks later.

This is the majority of the work being done or packages/units being rewritten or at least revisited:

* Draft, PoC, new core types
* Ingress resource parser
* Abstract annotation extractor
* Annotation and configmap parsers and converters
* TCP services
* HAProxy config builder and synchronizer
* Bindings to listers and caches
* Remove legacy code

Some of the benefits to the sysadmin or the end user of the k8s api:

* Move most of the configuration to HAProxy's backend
* Move a big amount of the template logic to a testable Go code
* Backend annotations via ingress as well as service resources
* Fix some annoying reloads if there is any service, endpoint or cache error

Some of the benefits to the ingress controller developer:

* Codebase easier to understand, maintain, evolve, colaborate and test
* New code with about 100% of code coverage, using a lot of real use case scenarios
* Remove the `controller -> config -> dynconfig -> template` overhead
* Ingress/services/annotations converter decoupled from listers/caches and from internal details of the HAProxy config schema
* HAProxy config schema decoupled from the converter and from the k8s api

Happy new year!

jcmoraisjr:
Hi, just to let you know the first v0.8 snapshot is out! Have a look at the release notes:

https://groups.google.com/forum/#!topic/haproxy-ingress/UFQX56_C4TQ

brianloss:
Looks good. I’ll check it out. Thanks!

[...]

PR #289, opened 2019-01-20
New ingress converter and haproxy types (part 1)

jcmoraisjr:
Core types and starting implementation of a new ingress synchronizer, parser, and HAProxy updater.

A converter implementation parses k8s objects and convert to an internal representation of an HAProxy instance status. The only current implementation knows ingress resources and its dependencies, providing almost 100% of backward compatibility.

An HAProxy instance configuration has internal status which is updated by a converter. The HAProxy configurations doesn’t have coupling with k8s api. It has the responsibility of hold HAProxy current and previous status, compare each other in order to run dynamic update (if possible), update configuration files based on the internal status and reload HAProxy if dynamic update cannot be done.

The controller listen k8s api, create and hold all used objects (listers / cache), create and hold the HAProxy instance reference, create and exec an ingress converter every time the list of watching objects is changed.

This will be the default implementation on v0.8, deprecating common ingress. v0.9 will remove all legacy and unused common ingress code.

Some missing pieces of this starting version:

* a lot of backend annotations
* all global configs
* template files
* api docs
* a few converters and haproxy tests

Implements #274


rikatz:
Let me know if you need somewhere to test this. Some people say I've a good and huge environment for those kind of refactoring :P

jcmoraisjr:
Hey, thanks for the offer! I'm aware of that environment. This'd be indeed an invaluable help on functional and also non functional tests.

PR #302, opened 2019-02-18
Add CRL Support in the TLS Secret for Client Authentication

chazmead:
Added an additional check for ca.crl in the TLS secret, and storing the
provided PEM file separately allowing the HAproxy server to verify a
client certificate being used has not been revoked.


jcmoraisjr:
Hi, sorry about the very long delay and thanks for supporting this controller!

Please rebase the PR to the master branch, a cherry pick is done when needed and only to fix maintained versions. v0.7 is now feature freeze.

v0.8 has now a new controller which is being merged tomorrow from `jm-v08-controller-2` branch. Perhaps you want to have a look and also implement there - if so use `--v07-controller=false` to test.

PR #306, opened 2019-02-27
Fix broken Template link

jeunii:
The current links to both the templates `haproxy.tmpl` and `spoe-modsecurity.tmpl` are broken. Pointed them to the correct location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ in-memory parsed template.
 
 |Mounting directory|Configmap keys (filenames)|Source (choose a proper tag)|
 |---|---|---|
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Hi, thanks for pointing this out and sorry the long delay.

The template will continue with the same name, v0.7 version was renamed in order to coexist with v0.8 template version - which will be merged tomorrow! So I think temporarily link to the directory is better than link to the filename. Please use the filename without the version number and remove the name from the link in order to point to the directory instead.

jcmoraisjr:
Hi, thanks again for the PR. v0.8 is now in place which fixed the issue. Closing.

PR #328, opened 2019-04-21
Add CRL Support in the TLS Secret for Client Authentication

chazmead:
Added an additional check for ca.crl in the TLS secret, and storing the
provided PEM file separately allowing the HAproxy server to verify a
client certificate being used has not been revoked.

### DIFF
--- a/pkg/common/ingress/controller/backend_ssl.go
+++ b/pkg/common/ingress/controller/backend_ssl.go
@@ -111,7 +111,8 @@ func (ic *GenericController) getPemCertificate(secret *apiv1.Secret) (*ingress.S
 		}
 
 	} else if ca != nil {
-		s, err = ssl.AddCertAuth(nsSecName, ca)
+		crl, _ := secret.Data["ca.crl"]
+		s, err = ssl.AddCertAuth(nsSecName, ca, crl)
 
 		if err != nil {
 			return nil, fmt.Errorf("unexpected error creating pem file: %v", err)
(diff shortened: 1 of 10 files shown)


pkg/common/ingress/controller/backend_ssl.go
```diff
@@ -111,7 +111,13 @@ func (ic *GenericController) getPemCertificate(secret *apiv1.Secret) (*ingress.S
 		}
 
 	} else if ca != nil {
-		s, err = ssl.AddCertAuth(nsSecName, ca)
+		crl := []byte{}
+		_, crlok := secret.Data["ca.crl"]
+		if crlok {
+			crl = secret.Data["ca.crl"]
+		}
+
```
jcmoraisjr: Just a `crl, _ := secret.Data["ca.crl"]` should be enough.

chazmead: Yeah, not sure what was going through my mind when I wrote this!  will amend this into the commit.

jcmoraisjr:
Hi, just doing some tests and observed that you documented and implemented `ingress.kubernetes.io/auth-tls-secret`, used to validate client cert, but configured `ingress.kubernetes.io/secure-verify-ca-secret` instead, used to validate cert of the backend. You can find the correct place [here](https://github.com/jcmoraisjr/haproxy-ingress/blob/fb887aaabfb874f3bcc9fcee6fd1826f155d4123/rootfs/etc/haproxy/template/haproxy-v07.tmpl#L335).

jcmoraisjr:
Hi, no need to fix, I'll merge just like it's now and implement v0.9 support on top of it. Thanks!

nrvmodi:
Hello, Is this support for tcp-services-configmap?

PR #330, opened 2019-04-26
Fix port retrieval for terminatingPod with named targetPort.

brianloss:
Discovered a bug where services with a named targetPort would end up with the port set to 0 in the haproxy config when drain support is on and the pod is terminating. This is because the port could not be converted to an int and instead needs to be retrieved by name from the pod's containers' ports.

I put this into release-0.6 figuring it could be merged up through that, release-0.7 and master to cover any bugfix releases for the release 0.6 and 0.7 branches. However, if you prefer, I can just re-target this to master.


jcmoraisjr:
Hi, thanks! And yes, please rebase it to master. I'll cherry-pick the commit to the release branches.

brianloss:
It seems you cannot change the source branch on a PR. I will have to re-submit this to pull from master->master.

PR #331, opened 2019-04-29
Fix port retrieval for terminatingPod with named targetPort.

brianloss:
Discovered a bug where services with a named targetPort would end up with the port set to 0 in the haproxy config when drain support is on and the pod is terminating. This is because the port could not be converted to an int and instead needs to be retrieved by name from the pod's containers' ports.

Please cherry-pick back to release-0.6 and release-0.7 so it's included in any future bug fix releases from those branches.

### DIFF
--- a/pkg/common/ingress/controller/controller.go
+++ b/pkg/common/ingress/controller/controller.go
@@ -28,9 +28,6 @@ import (
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Merging, thanks! I'll tag a new v0.6 and v0.7 as soon as HAProxy 1.8.20 is released.

PR #352, opened 2019-06-26
Fix duplication of ConfigFrontend snippets for DefaultBackend

gavinbunney:
This fixes an issue where any frontend configuration is duplicated twice for the default backend.

Example for a `config-frontend` in the ConfigMap of `http-response add-header my-header` results in two headers `my-header` in the response from the default backend.

This is particularly bad when forcing http to https redirection using `config-frontend` we were seeing an infinite loop of redirects for unknown ingress endpoints rather than a 404.


jcmoraisjr:
Hi, good catch. Makes sense since the default backend is always behind another frontend, hence the duplication. Merging and cherry-picking to v0.7.

PR #354, opened 2019-07-10
Remove Basic Auth challenge for pre-flight OPTIONS request when CORS is enabled

peterbroadhurst:
When CORS is enabled on a k8s ingress definition that also has Basic Auth, the connection from a browser fails because the browser (correctly) does not send the Basic Auth credentials on the pre-flight `OPTIONS` request.

Example annotations to recreate:
```yaml
    ingress.kubernetes.io/auth-realm: Authentication Required
    ingress.kubernetes.io/auth-secret: somesecret
    ingress.kubernetes.io/auth-type: basic
    ingress.kubernetes.io/cors-allow-credentials: "true"
    ingress.kubernetes.io/cors-allow-methods: POST,PUT,GET,DELETE
    ingress.kubernetes.io/cors-allow-origin: '*'
    ingress.kubernetes.io/enable-cors: "true"
```

This PR suggests updating the `http-request auth` rule to be disabled for `OPTIONS` requests when CORS is enabled on a backend definition. This would make haproxy consistent with the behaviour in nginx ingresses (we are evaluating a switch from nginx to haproxy to benefit from features only currently available in haproxy).

I did look at creating a PR to the master branch, but hit some unrelated issues testing the master branch with CORS. For our needs it would be great to be able to get this change into an upcoming 0.7 release so we avoid being on a long lived fork getting to production ready 0.8. So I'm kicking off the discussion here against the 0.7 release. However, if you think it's more expedient to work on the 0.8 PR first, let me know and I can raise a separate PR now to discuss there.

PR #356, opened 2019-07-15
Disable HTTP Basic Auth on CORS pre-flight OPTIONS request

peterbroadhurst:
> port of https://github.com/jcmoraisjr/haproxy-ingress/pull/354 to master

When CORS is enabled on a k8s ingress definition that also has Basic Auth, the connection from a browser fails because the browser (correctly) does not send the Basic Auth credentials on the pre-flight OPTIONS request.

Example annotations to recreate:

```yaml
    ingress.kubernetes.io/auth-realm: Authentication Required
    ingress.kubernetes.io/auth-secret: somesecret
    ingress.kubernetes.io/auth-type: basic
    ingress.kubernetes.io/cors-allow-credentials: "true"
    ingress.kubernetes.io/cors-allow-methods: POST,PUT,GET,DELETE
    ingress.kubernetes.io/cors-allow-origin: '*'
    ingress.kubernetes.io/enable-cors: "true"
```

This PR suggests updating the http-request auth rule to be disabled for OPTIONS requests when CORS is enabled on a backend definition. This would make haproxy consistent with the behaviour in nginx ingresses (we are evaluating a switch from nginx to haproxy to benefit from features only currently available in haproxy).

Questions from PR 354 moved to discussion here:

> What kind of features did you find on haproxy you didn't on nginx?

Hitless reload is the key feature driving us to adopt haproxy.

> Which issues you have found testing on master?

Swapping the image from the `0.7` branch to one built with this PR, I see this error processing the CORS config:

```
E0710 22:10:00.594793       8 controller.go:350] unexpected failure restarting the backend: 
template: haproxy-v07.tmpl:727:12: executing "CORS" at <$cors.CorsExportHead...>: can't evaluate field CorsExportHeaders in type *cors.CorsConfig
```

Related to these lines of the config template:
https://github.com/jcmoraisjr/haproxy-ingress/blob/d85fa4e4889ec10a49cb1965efbbe3ef78d35c4b/rootfs/etc/haproxy/template/haproxy-v07.tmpl#L727-L729

### DIFF
--- a/rootfs/etc/haproxy/template/haproxy-v07.tmpl
+++ b/rootfs/etc/haproxy/template/haproxy-v07.tmpl
@@ -459,7 +459,7 @@ frontend http{{ if and $hasSSL (not $hasBalance) }}s{{ end }}front-{{ if $isShar
 {{- $listName := $location.Userlist.ListName }}
 {{- if ne $listName "" }}
 {{- $realm := $location.Userlist.Realm }}
-    http-request auth {{ if ne $realm "" }}realm "{{ $realm }}" {{ end }}if {{ $server.ACLLabel }}{{ $location.HAMatchPath }} !{ http_auth({{ $listName }}) }
+    http-request auth {{ if ne $realm "" }}realm "{{ $realm }}" {{ end }}if {{- if $location.CORS.CorsEnabled }}!METH_OPTIONS{{ end }} {{ $server.ACLLabel }}{{ $location.HAMatchPath }} !{ http_auth({{ $listName }}) }
 {{- end }}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Hi, sorry about the long delay. Couldn't find time to anything else these days during the development of the current snapshot.

The change is fine, there is just a formatting issue on `{{- if` which would place together `if` and `!METH_OPTIONS` without space. I'll release snapshot.3 shortly and will fix this on my own PR.

> Hitless reload is the key feature driving us to adopt haproxy.

Great. HAProxy 1.8+ does a nice job here.

> can't evaluate field CorsExportHeaders

Sure, I saw the problem here. I didn't use v07 controller anymore so I didn't see this one. snapshot.3 will have this issue fixed.

Regarding rename, git tracks the change and knows that haproxy-v07.tmpl here is haproxy.tmpl there.

Merging now and snapshot.3 is coming soon. I'd like to issue the next v0.7 and v0.6 with haproxy 1.8.21 so I'll wait just a few more days. If we don't have a new release there, I'll tag new releases here anyway.

Thanks for evaluating haproxy-ingress!

PR #363, opened 2019-07-30
Add --wait-before-shutdown flag to wait defined time when SIGTERM received

mkubaczyk:
I've encountered a case, when deploying haproxy ingress controller with helm chart and some changes in Deployment object (which already had terminationGracePeriodSeconds as 4 minutes), my controller's Pods once received SIGTERM, shut down themselves almost immediately, causing the connections to be broken.
I'd like to propose a parameter of --terminate-grace-period defined as integer of seconds, which causes the Stop function to wait defined time before it shuts down the controller's components. Thanks to this, if terminationGracePeriodSeconds and --terminate-grace-period are defined, long living connections might have a chance to complete, while the Pod is taken of out Service's endpoints, so no new requests are coming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -986,6 +986,7 @@ The following command-line arguments are supported:
 
 ||Name|Type|Default|
 |---|---|---|---|
+||[`termination-grace-period`](#termination-grace-period)|seconds as integer|`0`|
 ||[`allow-cross-namespace`](#allow-cross-namespace)|[true\|false]|`false`|
 ||[`default-backend-service`](#default-backend-service)|namespace/servicename|(mandatory)|
 ||[`default-ssl-certificate`](#default-ssl-certificate)|namespace/secretname|(mandatory)|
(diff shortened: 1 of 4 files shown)


jcmoraisjr:
Hi, thanks for supporting this controller!

I have some changing suggestions:

1. grace period is related to the amount of time a controller will wait for an event to occur, eg how much time kubelet wait a pod to finish before issuing a sigkill. Your implementation sounds more like a "wait-before-terminate" but perhaps you have a better name.
2. thanks for the doc!! but please place it in alphabetical order.

mkubaczyk:
Hi @jcmoraisjr,

I've renamed it to `wait-before-shutdown`, should be ok now. And followed the alphabetical order you asked for. :-)

jcmoraisjr:
Hi, the last two changes:

* The flags.Int() usage is printing the double quotes and plus sign. Just remove them from inside the back quotes, which does support line breaks. Follow the other usage declarations.
* Add everything inside a `if hc.cfg.TerminationGracePeriod > 0` so one doesn't need to see `waiting 0s` if not declared.

mkubaczyk:
Hi @jcmoraisjr,

thanks for reply. I've followed your proposition with the newest commit. :-)

mkubaczyk:
I'll take a closer look at failing tests soon

jcmoraisjr:
Hi, don't worry, just an intermitent test of a legacy code. Merging, thanks!

PR #391, opened 2019-09-22
Add acme v02 support

jcmoraisjr:
Add starting implementation of acme v02 / RFC 8555 order, challenge and validation support.

* [x] acme client (orchestration)
* [x] acme server (challenges)
* [x] fronting proxy config
* [x] manual integration tests
* [x] automated integration tests
* [x] unit tests
* [x] leader / concurrency
* [x] rate limit auth failures
* [x] docs


abh:
What's the benefit of using this over using cert-manager?

Cert manager has a much more extensive support for other validation types, can support non-HTTP systems, etc etc.

jcmoraisjr:
Hi, this is a fair question, thanks for asking.

These are some issues we found on cert-manager in no special order:

* excessively complex
* nginx-ingress driven (look for nginx in the code)
* need haproxy reload in the current implementation
    * will use a lot of memory due to lots of haproxy instances when websocket/tunnel timeout is high; or
    * would be even slower if we rate limit reloads
* it doesn't work on our clusters partially because its own complexity
    * https://github.com/jetstack/cert-manager/issues/362
    * https://github.com/jetstack/cert-manager/pull/1338

PR #415, opened 2019-10-21
Kubernetes version 1.16 removed several deprecated APIs

irizzant:
Kubernetes version 1.16 has removed several deprecated APIs that you'll likely recognize and may use on a daily basis


jcmoraisjr:
Hi, thanks! Merging.

PR #443, opened 2019-11-08
Add support for Mod Security DetectionOnly Mode

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@serpro.gov.br>

This PR adds support for a DetectionOnly mode in Mod Security for each backend.

The idea here is when specifying the annotation 'waf-mode: "DetectionOnly"' it will send the requests to the ModSecurity filter to generate the appropriate logging, but no action will be taken.

To keep the compability with following versions, the behavior here is: if there's a ``waf`` annotation and no ``waf-mode`` annotation, it will block bad requests.

To enable DetectionOnly mode user must create the ``waf-mode`` annotation with DetectionOnly as its value

### DIFF
--- a/pkg/converters/ingress/annotations/backend.go
+++ b/pkg/converters/ingress/annotations/backend.go
@@ -737,15 +737,20 @@ func (c *updater) buildBackendWAF(d *backData) {
 				c.logger.Warn("ignoring invalid WAF module on %s: %s", waf.Source, waf.Value)
 				return nil
 			}
+			wafMode, foundWAFMode := values[ingtypes.BackWAFMode]
+			if !foundWAFMode {
+				return values
+			}
+			if wafMode.Value != "deny" && wafMode.Value != "detect" {
+				c.logger.Warn("ignoring invalid WAF mode '%s' on %s, using 'deny' instead", wafMode.Value, wafMode.Source)
+				wafMode.Value = "deny"
+			}
 			return values
 		},
 	)
 	for _, cfg := range config {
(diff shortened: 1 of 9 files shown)


pkg/converters/ingress/annotations/backend.go
```diff
@@ -727,23 +727,31 @@ func (c *updater) buildBackendTimeout(d *backData) {
...
 	)
 	for _, cfg := range config {
-		d.backend.WAF = append(d.backend.WAF, &hatypes.BackendConfigStr{
-			Paths:  cfg.Paths,
-			Config: cfg.Get(ingtypes.BackWAF).Value,
+		wafModule := cfg.Get(ingtypes.BackWAF).Value
+		wafMode := cfg.Get(ingtypes.BackWAFMode).Value
+		if wafMode != "On" && wafMode != "DetectOnly" {
+			wafMode = "On"
+		}
```
jcmoraisjr: * What if mode was `deny` and `detect` in lowercase? This could be used in future improvements where `On` could lead to ambiguity. This would also change the doc and the PR description.
* Add a warning log if mode doesn't match, something like `invalid '%s' on %s, using 'on' instead`, so the admin can have a hint if his proxy is misbehaving.
* Add the default value to `default.go` so this will be documented in code in one single place.

rikatz: :heavy_check_mark:

jcmoraisjr:
Thanks! Merging.

PR #445, opened 2019-11-11
docs: defaults for cors-allow-methods and -headers

ktosiek:
The defaults for cors-allow-methods and cors-allow-credentials were swapped in the docs.


jcmoraisjr:
Thanks! Merging.

PR #460, opened 2019-11-25
supporting build on different architectures

l27-0-0-l:
example: make GOARCH=arm


jcmoraisjr:
Hi, plase fork to master instead. Please also provide some words in the readme file, A few words in a single line, between the make targets and the last paragraph, saying something like "use GOARCH var to change the architecture, eg GOARCH=arm" is enough.

PR #465, opened 2019-11-28
do ssl-redirect only if tls declares the hostname

jcmoraisjr:
Ingress.spec.tls declares hostnames/domains and an optional secret with crt+key used to ssl-offload https requests on that domains. If a secret name is not declared or is invalid, the default crt+key is used instead.

This commit fixes a misbehaviour when ssl-redirect is true, declared or inherited from configmap, without a TLS being declared. Now ssl-redirect only return 302 redirect if ingress.spec.tls declares the hostname/domain.

Should be merged to v0.8.

### DIFF
--- a/pkg/haproxy/config.go
+++ b/pkg/haproxy/config.go
@@ -318,7 +318,7 @@ func (c *config) BuildFrontendGroup() error {
 				backend := c.FindBackend(path.Backend.Namespace, path.Backend.Name, path.Backend.Port)
 				base := host.Hostname + path.Path
 				hasSSLRedirect := false
-				if backend != nil {
+				if host.TLS.HasTLS() && backend != nil {
 					hasSSLRedirect = backend.HasSSLRedirectHostpath(base)
 				}
 				// TODO use only root path if all uri has the same conf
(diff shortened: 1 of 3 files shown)


prometherion:
I tested this branch and it seems to address the issue: well done!

However, considered the following ingress:
```yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: foo
spec:
  tls:
    - hosts:
      - 42.dontpanic.space
  rules:
    - host: 42.dontpanic.space
      http:
        paths:
          - path: /
            backend:
              serviceName: dontpanic
              servicePort: 80
```

I can reach the host both on plain HTTP and HTTPS: in other Ingress controllers like NGINX, the default behavior is to force automatically the upgrade to HTTPS but it seems not happening here.

Even considered that this could be the default behavior, do you plan to achieve that goal?
From an end-user perspective, if I'm declaring an Ingress with a TLS host, I'm expecting to get the redirect by default.
Despite of that, if I would like to serve both HTTP and HTTPS I just need to use the annotation `ingress.kubernetes.io/ssl-redirect=false`

prometherion:
Nah, my bad: ignore my last comment... it works as expected but I was building this branch with other changes that weren't handling correctly the redirection.

PR #476, opened 2019-12-11
Fix haproxy.cfg permissions

abh:


jcmoraisjr:
Great, thanks! Merging.

PR #480, opened 2019-12-13
deduplicate changes within a small time range

jcmoraisjr:
Ingress parsing is a somewhat expensive task on big clusters (1000+ ingress objects). Thanks `--rate-limit-update` two consecutive parsings cannot happen below a configurable amount of time.

However a flood of changes wasn't being deduplicated. If eg rate limit is 0.25 (at least 4s between updates) and 10 changes would happen, the controller would parse all the configuration 10 times, in the next 40s (ok 36s). The change here is to enqueue the same object, which will be deduplicated by the work queue, instead of the object received to be enqueued. Reasons why this is not a problem at all:

- Only status and controller parsers (v07 and v08+) uses this work queue
- Status already starts the job using the same dummy string
- Controller receives the changed object but doesn't use it - all the configuration is parsed on every single change - this can be reviewed in the future

### DIFF
--- a/pkg/common/task/queue.go
+++ b/pkg/common/task/queue.go
@@ -59,6 +59,8 @@ func (t *Queue) Run(period time.Duration, stopCh <-chan struct{}) {
 	wait.Until(t.worker, period, stopCh)
 }
 
+var notification = Element{Key: nil, Timestamp: 0}
+
 // Enqueue enqueues ns/name of the given api object in the task queue.
 func (t *Queue) Enqueue(obj interface{}) {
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
This was done in the wrong place. The majority of false positive updates comes from the pod lister when drain support is disabled, and the fact that the "before sync" timestamp in the work queue happens before the rate limit.accept() inside ingress.sync. Btw the work queue actually does a workaround saving the timestamp before sync in order to check if new events happen before or after calling sync(). Not so nice but it works. Closing.

PR #495, opened 2020-01-09
Fix TLS deployment

mtatheonly:
Fix deployment of TLS enabled haproxy, requiring a secret to be created in the same namespace as the ingress-controller


jcmoraisjr:
Thanks! Merging.

PR #497, opened 2020-01-10
[DOCS] Rename configmap to ConfigMap

antcs:
* As in the official Kubernetes documentation it is called `ConfigMap` and not `configmap`
* https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/


jcmoraisjr:
Sounds fair, merging, thanks!

PR #507, opened 2020-01-24
add backend-server-naming key

jcmoraisjr:
This fixes a small regression since v0.8: when dynamic scaling is disabled, backend servers were named with target's ip/port which give some context to logs. The behaviour is slightly different now: default value is sequence despite the dynamic update config, the admin/dev can change it using svc/ing annotation or globally using configmap.


brianloss:
Looks good to me. Thanks for the quick turn around!

PR #525, opened 2020-02-28
GlobalSSLRedirectCode was missing a default value

coldeasy:
This caused the haproxy.tmpl rendering to fail unless `ssl-redirect-code` was
specified by the user.


jcmoraisjr:
Hi, thanks for supporting this controller!

What's the haproxy configuration that was broken if a `ssl-redirect-code` isn't provided? I couldn't reproduce the error. The actual implementation hides the configuration if not provided, which defaults to the default haproxy redirect code, `302`.

jcmoraisjr:
Closing due to inactivity. Please update with a new comment if you still have any problem.

Issue #350, opened 2019-06-11
[Question] Is regex in HTTPIngressPath supported?


mvpmvh:
[this issue](https://github.com/jcmoraisjr/haproxy-ingress/issues/207) is closed, but I can't tell if anything ever came of it.
Is it possible today to use regex to route to backends?

jcmoraisjr:
Hi, this sounds a fair addition to v0.8. I'll take care of this.

sergiimk:
We have also encountered this limitation as we are transitioning to `haproxy-ingress` from `ingress-nginx` and needed feature parity with its [regex matching](https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#regular-expression-support).

We worked around this by introducing the `ingress.kubernetes.io/path-type: regex` annotation and having controller use [HTTPFrontsMap.RegexFile](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/rootfs/etc/haproxy/template/haproxy.tmpl#L789) map file  instead of the `MatchFile` for paths that are part of such annotated ingress.

Our implementation is definitely on a "quick and dirty" side (though happy to share if needed), so it would be great to see this feature properly supported. Regex rules are very useful for bulkhead-style QoS and slicing up old monoliths into smaller services.

jcmoraisjr:
Hey, thank you for choosing haproxy ingress and sorry about this missing feature. Please share your code creating a new PR, I'll have a look before cutting v0.10 to beta.

sergiimk:
I've tried upgrading to latest version today that includes #641. I followed the docs and used:
```yaml
path: ^/v[0-9]+/(foo|bar)/
```
But this gets rejected by k8s:
```
The Ingress "foo" is invalid: spec.rules[0].http.paths[0].path: Invalid value: "^/v[0-9]+/(foo|bar)/": must be an absolute path
```
Which is related to kubernetes/kubernetes#68413

When using:
```yaml
path: /v[0-9]+/(foo|bar)/
```
I of course get an incorrect regex in `/etc/haproxy/maps/_front_https_host__regex.map`:
```
^foo\.example\.net/.*/v[0-9]+/(foo|bar)/ ...
```
which will not match a path like `/v1/foo/x` as it expects an extra slash...

@jcmoraisjr did you encounter this problem during #641? Perhaps there's a workaround to bypass kube's (quite useless) validation that I'm missing?

jcmoraisjr:
[...]

PR #551, opened 2020-04-08
Support regex path matching

sergiimk:
Addresses #350 

@jcmoraisjr, as I mentioned this is a quick and dirty implementation and I don't think it should be merged as-is, but posting this to give you an idea of the approach we took.

// Kudos for all the work you've put into this project!


jcmoraisjr:
Hi, thanks for choose and support this controller! The PR looks good, there is eg no better place to read the annotation due to its scope.

Following a few remarks:

* Change to host annotation scope. This won't change how you read it but it'll be clear that it cannot be declared in the service object
* Reuse `host.AddPath()`, add a new `isRegex bool` argument. Yes, this will break dozens of lines of test
* What if a hostname `*.example.local` is used with path regex? Perhaps `hostsMap.AppendHostname()` should also be reused? Please add also some lines to the `AppendHostname()` testcase as well.
* Add a few lines to the doc, so people will know this functionality exists.

jcmoraisjr:
Hi, thank you so much for the contribution. I'm closing this PR on behalf of #641 which implements match type on a more extensible way.

sergiimk:
Thanks @jcmoraisjr it's great to know that regex path matching will become a supported feature. I felt that what I proposed here was a very hacky implementation that worked for our use case but was not good enough to be upstreamed.

Issue #565, opened 2020-04-20
Server-alias doesn't not work on http:80 => 0.7.2 to 0.8.3


AlexisDuf:
Hello, 

I tried to upgrade from 0.7.2 to 0.8.3 but it seems I'm facing to some issues when I use `server-alias` annotations. 

I'm currently run kubernetes on 1.16 version. 
My main domain is *.k8s-hpr.my.domain.fr
For one of my app the subdomain myapp.k8s-hpr.my.domain.fr and I created a server alias myapp.domain.fr

It was working perfectly on 0.7.2 but with 0.8.5 is not working anymore. 

After digging in the config file of the ingress controller I noticed some strange things. (By the way I'm not an expert at all of Haproxy conf) 

Problems seems to be on map file, I did a grep on the map to check if I can find the server alias. 
```bash
grep 'myapp' /etc/haproxy/maps/_*
/etc/haproxy/maps/_front001_host.map:myapp.k8s-hpr.my.domain.fr/ myapp_occ-service-integration_8080
/etc/haproxy/maps/_front001_host.map:myapp.domain.fr/ myapp_occ-service-integration_8080
/etc/haproxy/maps/_global_http_front.map:myapp.k8s-hpr.my.domain.fr/ myapp_occ-service-integration_8080
/etc/haproxy/maps/_global_https_redir.map:myapp.k8s-hpr.my.domain.fr/ no
/etc/haproxy/maps/_global_k8s_ns.map:myapp.k8s-hpr.my.domain.fr/ -
/etc/haproxy/maps/_public.list:myapp.k8s-hpr.my.domain.fr
```
According to me we should have:
```bash
myapp.domain.fr/ myapp_occ-service-integration_8080
``` 
In `/etc/haproxy/maps/_global_http_front.map` 

Because based the `haproxy.cfg` for http without https redirect we use `_global_http_front.map` map conf
```
haproxy.cfg
...
...
frontend _front_http
    mode http
    bind *:80
    http-request set-var(req.base) base,lower,regsub(:[0-9]+/,/)
    http-request redirect scheme https if { var(req.base),map_beg(/etc/haproxy/maps/_global_https_redir.map,_nomatch) yes }
....
    http-request set-var(req.backend) var(req.base),map_beg(/etc/haproxy/maps/_global_http_front.map,_nomatch)
    use_backend %[var(req.backend)] unless { var(req.backend) _nomatch }
    default_backend _default_backend
```

[...]

PR #570, opened 2020-04-30
Server-alias doesn't not work on http:80 => 0.7.2 to 0.8.3 #565

AlexisDuf:
I added the missing call functions if we don't have ssl redirect


jcmoraisjr:
Hi thanks for testing this patch. As you can see [here](https://travis-ci.org/github/jcmoraisjr/haproxy-ingress/builds/681386995) there is a single unit test broken, just add the suggested line to fix it and note that `spaces != tabs`. Run a local `make test` so you can confirm everything is fine.

AlexisDuf:
Hi @jcmoraisjr 

Sure i will check it this week

jcmoraisjr:
Hi, don't worry, just fixed and merged, thanks!

PR #585, opened 2020-05-18
Typo on configuration keys docs

RobertTheProfessional:
Had me looking at my config map crazy


jcmoraisjr:
Hey there, what a bad place to have a typo. Thanks for the fix! Merging.

PR #588, opened 2020-05-21
Allow overriding CPU Map

coldeasy:
- Support to entirely disable CPU pinning or have a custom form


jcmoraisjr:
Hi, thanks for supporting this controller! Extra bonus for the docs and the tests.

Following a few remarks:

* Use a single section `## cpu-map` declaring both keys. There are some examples out there.
* Add both keys in the Configuration Keys table in alphabetical order
* Use alphabetical order everywhere - doc, defaults, declaration.
* If possible squash and force push the resulting changes into a single commit

coldeasy:
Thanks for the quick feedback @jcmoraisjr - updated based on your comments and rebased off master.

jcmoraisjr:
Thanks! Merging.

PR #589, opened 2020-05-21
TCP Services : SSL : Optionally Verify Client

hileef:
Hello again @jcmoraisjr ,

I would like to propose a tentative implementation of ssl client verification for tcp services.
Currently we have a second instance of HAProxy whose sole job is to do ssl-offload AND verifying the client, ensuring the certificate presented by the client is indeed signed by the provided CA.

If we can do that ssl verification of the client directly at the ingress controller level, it would simplify things for us 😄 .

Let me know what you think !


jcmoraisjr:
HI, thanks for supporting this controller! Tests and doc also updated, thanks again.

It's becoming quite complex such positional params, and we also missed the timing to add ca/crl just after the crt. Well, no problem at all, planning a small and backward compatible refactor in the tcp-service to the next release.

Ok, back to the pr, following a few suggestions:

* Please add also a few more cases in the `tcpservices_test` as well.
* If possible squash and force push all the changes into a single commit.

hileef:
> we also missed the timing to add ca/crl just after the crt. 

Yah, it's unfortunate, but I'll let you decide on the best strategy for a future refactor 🙂 

> Please add also a few more cases in the tcpservices_test as well.
> If possible squash and force push all the changes into a single commit.

Alright, will do that today 👍

hileef:
Weird, I had a failed build and pushed again to get it working 🤔 
Is there a known flakiness/concurrency issue in the tests ?

Here are the logs of the failed build [here](https://travis-ci.org/github/jcmoraisjr/haproxy-ingress/jobs/690175573)

```
--- FAIL: TestDeduplicate (0.20s)
    queue_test.go:119: items differ, expected: [<nil>  1 1]; actual: [<nil>  1 1 <nil> ]
```

The second build went through okay 👌 
Let me know if you need any further changes 😸

jcmoraisjr:
Thanks! Merging.

Issue #573, opened 2020-05-07
Host scoped bind options


Unichron:
For my use-case, I need to configure different cipher suites and minimum tls versions for different hosts. This would be possible with host scoped ssl-options and ssl-ciphers/cipher-suites. If I'm correct I think that it would be also possible with just a host scoped bind-https, which would enable any bind related option to be host scoped, acting as a nice catch-all for edge-cases like mine.

What do you think?

jcmoraisjr:
Hi, sure, missed moving such global config options when the frontend refactor was done on v0.9. Thanks for the record.

Unichron:
@jcmoraisjr I would like to create a pull request for this because my company will need it soon-ish. Is there something design related that needs to be discussed or do you think it's straightforward enough?

jcmoraisjr:
It seems to me straightforward enough although it isn't a trivial change. Ping here if you have any doubt.

PR #609, opened 2020-06-25
Host scoped cipher options

Unichron:
This is for https://github.com/jcmoraisjr/haproxy-ingress/issues/573

So far I only implemented the ciphers list, I'm planning to do add the generic bind-options as well but I wanted to do a quick feedback loop to verify if I'm going in the right direction and/or if I'm missing some stuff. I saw the todo about the template, but I'm not sure it's worth it, so I went with an approach I think is good enough and scalable for what it needs to do.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -200,9 +200,9 @@ The table below describes all supported configuration keys.
 | [`session-cookie-shared`](#affinity)                 | [true\|false]                           | Backend | `false`            |
 | [`session-cookie-strategy`](#affinity)               | [insert\|prefix\|rewrite]               | Backend |                    |
 | [`slots-min-free`](#dynamic-scaling)                 | minimum number of free slots            | Backend | `0`                |
-| [`ssl-cipher-suites`](#ssl-ciphers)                  | colon-separated list                    | Global  | [see description](#ssl-ciphers) |
+| [`ssl-cipher-suites`](#ssl-ciphers)                  | colon-separated list                    | Host    | [see description](#ssl-ciphers) |
 | [`ssl-cipher-suites-backend`](#ssl-ciphers)          | colon-separated list                    | Backend | [see description](#ssl-ciphers) |
-| [`ssl-ciphers`](#ssl-ciphers)                        | colon-separated list                    | Global  | [see description](#ssl-ciphers) |
+| [`ssl-ciphers`](#ssl-ciphers)                        | colon-separated list                    | Host    | [see description](#ssl-ciphers) |
 | [`ssl-ciphers-backend`](#ssl-ciphers)                | colon-separated list                    | Backend | [see description](#ssl-ciphers) |
 | [`ssl-dh-default-max-size`](#ssl-dh)                 | number                                  | Global  | `1024`             |
 | [`ssl-dh-param`](#ssl-dh)                            | namespace/secret name                   | Global  | no custom DH param |
@@ -1449,9 +1449,9 @@ service's ClusterIP is used instead.
 
 | Configuration key           | Scope     | Default | Since |
 |-----------------------------|-----------|---------|-------|
-| `ssl-cipher-suites`         | `Global`  |         | v0.9  |
+| `ssl-cipher-suites`         | `Host`    |         | v0.9  |
 | `ssl-cipher-suites-backend` | `Backend` |         | v0.9  |
-| `ssl-ciphers`               | `Global`  |         |       |
+| `ssl-ciphers`               | `Host`    |         |       |
 | `ssl-ciphers-backend`       | `Backend` |         | v0.9  |
 
 Set the list of cipher algorithms used during the SSL/TLS handshake.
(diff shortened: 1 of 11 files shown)


Unichron:
So I ran into a potential problem. The mapper always returns a value for the ssl-ciphers (because it has a default value and also the name of a global option), which makes the crt-list file really bloated. I'm not sure if it's a real problem or just unnecessary. I could of course check if the host value equals the global/default value and only put it into the crt-list if it is different. On the other hand it made me think, since haproxy allows a very specific set of options("npn", "alpn", "verify", "ca-file", "no-ca-names",
"crl-file", "ecdhe", "curves", "ciphers","ssl-min-ver", "ssl-max-ver") in the crt-list, maybe it would be better to just have a distinct host scoped only "bind-options" in which you can configure whatever haproxy supports/will support here. What's your thought?

jcmoraisjr:
Hi, thanks for the contribution! Sorry about taking so long to give a feedback.

Following a few remarks, let me know if I can help in any way:

* annotation reading will receive a few more lines, so create a new `c.buildHostTLSConfig(data)` func
* remove the global `ssl-cipher` config declaration; see how `BackTimeoutServer` was implemented, which also has a global default
* check if `mapper.Get(ingtypes.HostSSLCiphers).Source != nil` which means a config override, so only overrides will be configured
* implement also at least `ssl-cipher-suites` (tls1.3)
* use alphabetical order in `HostTLSConfig`
* edit the scope in the doc (host instead of global) - main table and also the ssl ciphers table
* if possible please squash+rebase everything into a single commit on top of master

Unichron:
Hi, thanks for the feedback!

Host level ciphersuites doesn't seems to be supported in haproxy even in 2.2: https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#5.1-crt-list

Unichron:
> Hi, thanks for the feedback!
> 
> Host level ciphersuites doesn't seems to be supported in haproxy even in 2.2: https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#5.1-crt-list

Okay so I also found it a bit strange that it is not supported so I looked at the source and it is there for some time now (at least in 2.0.0), but the documentation doesn't reflect that. I'm going to implement it.

docs/content/en/docs/configuration/keys.md
```diff
@@ -202,7 +202,7 @@ The table below describes all supported configuration keys.
 | [`slots-min-free`](#dynamic-scaling)                 | minimum number of free slots            | Backend | `0`                |
 | [`ssl-cipher-suites`](#ssl-ciphers)                  | colon-separated list                    | Global  | [see description](#ssl-ciphers) |
 | [`ssl-cipher-suites-backend`](#ssl-ciphers)          | colon-separated list                    | Backend | [see description](#ssl-ciphers) |
-| [`ssl-ciphers`](#ssl-ciphers)                        | colon-separated list                    | Global  | [see description](#ssl-ciphers) |
+| [`ssl-ciphers`](#ssl-ciphers)                        | colon-separated list                    | Host    | [see description](#ssl-ciphers) |
```
jcmoraisjr, line 205: Edit also `ssl-cipher-suites`

docs/content/en/docs/configuration/keys.md
```diff
@@ -1451,7 +1451,7 @@ service's ClusterIP is used instead.
 |-----------------------------|-----------|---------|-------|
 | `ssl-cipher-suites`         | `Global`  |         | v0.9  |
 | `ssl-cipher-suites-backend` | `Backend` |         | v0.9  |
-| `ssl-ciphers`               | `Global`  |         |       |
+| `ssl-ciphers`               | `Host`    |         |       |
```
jcmoraisjr, line 1454: Edit `ssl-cipher-suites`

docs/content/en/docs/configuration/keys.md
```diff
@@ -1470,6 +1470,8 @@ Default values on HAProxy Ingress v0.9 and newer:
 * TLS up to v1.2: `ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384`
 * TLS v1.3: `TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256`
 
+`ssl-ciphers` was `Global` scoped prior to v0.10.
+
```
jcmoraisjr, line 1474: `ssl-ciphers` and `ssl-cipher-suites` was `Global` scope up to v0.10.

pkg/converters/ingress/annotations/host.go
```diff
@@ -92,3 +92,12 @@ func (c *updater) buildHostSSLPassthrough(d *hostData) {
 	backend.ModeTCP = true
 	d.host.SetSSLPassthrough(true)
 }
+
+func (c *updater) buildHostTlsConfig(d *hostData) {
```
jcmoraisjr, line 96: rename to `buildHostTLSConfig` and linter will stay happier.

Unichron:
I think this can be merged now. I will add another PR for the other options. Few questions. So haproxy supports the following options in the crt-list: npn, alpn, verify, ca-file, no-ca-names, crl-file, ecdhe, curves, ciphers, ciphersuites, allow-0rtt, ssl-min-ver and ssl-max-ver. Out of these verify, ca-file, crl-file are covered by Auth TLS config. Ciphers and ciphersuites are covered by this PR. Alpn can be done with making tls-alpn host instead of global. What about the rest? They can be implemented with specific controller options or just a catch-all option. I'm not sure what's the the vision here.

jcmoraisjr:
Lgtm, thank you so much! Merging now.

I have no opinion atm about how to implement other bind options. A catch-all config sounds fair but I need to see if there is any corner case. There are some configs using a similar approach, eg config snippet and bind config keys. I'd try a catch-all approach and see how it evolves.

Unichron:
Thanks for merging!

So can I transform `ssl-options` to `Host` scope to implement the rest? Or do you have a better idea?

jcmoraisjr:
Hi, convert `ssl-options` sounds good, no better idea atm.

PR #613, opened 2020-07-05
Update deprecated APIs in Docs

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

This PR intends to start using the newer Ingress APIs instead of the deprecated ones. 

Instead of using extensions/v1beta1 it will use the newer networking.k8s.io/v1beta1

Right now I'm changing only the docs, there's a need of decision whether the older API will still be supported (as Ingress NGINX) or be fully deprecated in further versions.

Creating a Ingress resource with networking.k8s.io works fine in the current API and Ingress Controller (tested with v1.17.0) and the extensions/v1beta1 will be fully removed from Kubernetes API in version v1.22

# TODO

* [x] Decide if this PR is going also to change the behavior of Haproxy Ingress to support only the newer API version, otherwise check how to deal with both versions and print a warning to the user.

### DIFF
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -148,7 +148,7 @@ remove old configuration files. If `0`, the default value, a single `haproxy.cfg
 Some infrastructure tools like `external-DNS` relay in the ingress status to created access routes to the services exposed with ingress object.
 
 ```
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 ...
 status:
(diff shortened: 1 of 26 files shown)


jcmoraisjr:
Hi, thank you! It seems pretty good btw. Perhaps `examples/rbac/ingress-controller-rbac.yml` could also be updated?

jcmoraisjr:
About the todo - lets make this version a k8s v1.14+. I'll add this backward compatibility item in the changelog.

rikatz:
:man_facepalming: forgot this one 

Did a new commit

jcmoraisjr:
Thanks! Merging and sorry about the long delay. I'm aware we already have v1, let's update it in another step.

PR #617, opened 2020-07-10
Host scope tls-alpn and ssl-options

### LINKED ISSUE #573: Host scoped bind options

Unichron:
This is a second PR to close #573. I'm not entirely happy with the documentation of ssl-options, but this is the best I could come up with for now. Also I think it would make sense to deprecate/remove the no-tls* and force-tls* options in favor of ssl-min-ver and ssl-max-ver.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -211,7 +211,8 @@ The table below describes all supported configuration keys.
 | [`ssl-headers-prefix`](#auth-tls)                    | prefix                                  | Global  | `X-SSL`            |
 | [`ssl-mode-async`](#ssl-engine)                      | [true\|false]                           | Global  | `false`            |
 | [`ssl-options`](#ssl-options)                        | space-separated list                    | Global  | [see description](#ssl-options) |
-| [`ssl-options-backend`](#ssl-options)                | space-separated list                    | Global  | [see description](#ssl-options) |
+| [`ssl-options-backend`](#ssl-options)                | space-separated list                    | Backend | [see description](#ssl-options) |
+| [`ssl-options-host`](#ssl-options)                   | space-separated list                    | Host    | [see description](#ssl-options) |
 | [`ssl-passthrough`](#ssl-passthrough)                | [true\|false]                           | Host    |                    |
 | [`ssl-passthrough-http-port`](#ssl-passthrough)      | backend port                            | Host    |                    |
 | [`ssl-redirect`](#ssl-redirect)                      | [true\|false]                           | Backend | `true`             |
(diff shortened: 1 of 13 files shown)


jcmoraisjr:
Hi, thanks for the patch! I'd suggest the following changes, let me know what do you think:

* Create a `ssl-options-host` (since v0.11) host scoped config key, add its global value to the bind line (just before `crt-list`), and the host scoped value in the crt-list file (just like you are doing in this pr)
* Leave `ssl-options` just like it is now (before this pr) and update the doc to something like "* `ssl-options`: Default options for all the TLS frontend connections - HAProxy being the server."
* Remove the global `ssl-options-backend` attribute and also the `ssl-default-server-options` keywork in the template. Adjust the code to always store the backend scoped value because there isn't a global config anymore
* Add a blank line before a list in the doc (see "Supported options as an ingress annotation:"), otherwise the md parser will make a mess
* Try to squash and rebase the merges and your commit on top of master

In the end we'll have:

* `ssl-options`, global, with its limited list of keywords
* `ssl-options-<host|backend>`, host/back scoped, they support everything the bind / server line supports, but for the sake of simplicity, documenting only what bind, crt-list and server accept is more than enough

These changes should leave a bit easier to understand how stuff works, hence to make a nice doc.

Unichron:
Okay, I think i understand everything. Why do you want to remove the global default for the ssl-options-backend?

jcmoraisjr:
> Why do you want to remove the global default for the ssl-options-backend?

Because the haproxy's `ssl-default-server-options` keywork has only a partial ssl-related feature list. [`default-server`](https://cbonte.github.io/haproxy-dconv/2.1/configuration.html#4-default-server) could also be an option here but this would linearly configure to all backends despite it is ssl or not.

> Why add the global value of the ssl-options-host to the bind line? I think it only creates more confusion and the bind-https key already allows arbitrary stuff for the bind line.

`ssl-default-bind-options` keywork has also a partial ssl-related feature list, just like the `server` one. The idea here is to allow the same configuration options both in the global and local scopes. Thinking a bit more about this, the use of the bind like wouldn't be a good idea because a local declared value wouldn't override a global one, but they would instead be merged. The `default-server` has this same problem.

In the end my proposal is to leave `ssl-options` as is in the global scope and its limited feature list. `ssl-options-backend` and a new `ssl-options-host` has global and local scopes, they both added only in the server line (backend) or in the crt-list file (host). How does it sound?

Unichron:
Okay, sounds good. I'll make the changes in the next few days.

Unichron:
Hi,
I think I did everything. Not sure if the fork syncing merge commits are a problem, but I also have no idea how to get rid of them, squashed all my work though into 1 commit.

jcmoraisjr:
Thank you so much, it looks pretty good to me. The last two changes:

* `tls-alpn was Global scope up to v0.11.` in the `tls-alpn` doc. It should be v0.10 instead? `tls-alpn` will be host scoped in v0.11.
* "squash and rebase on top of master" is pretty much `git reset a5b3360; git add .; git commit; git push --force` - `a5b3360` is the upstream commit of your latest merge from master.

Unichron:
Done

jcmoraisjr:
Thanks! Merging.

PR #623, opened 2020-07-20
add backend-shards command-line option

jcmoraisjr:
Big clusters waste some time and cpu executing the haproxy template every time a configuration changes, even if such configuration is dinamically applied, leaving memory and disk configurations in sync. Most of the time is spent rebuilding all the referenced services, which is converted to haproxy backends.

Splitting the backends into smaller shards give haproxy ingress the ability to only update files that have changes to be applied, reducing io and cpu usage. The shard of a backend is chosen from the hash of its ID, so the same backend will always stay in the same shard provided that the number of shards isn't changed. Updating the number of shards will rebalance the backends.


tsjnsn:
@jcmoraisjr any guidelines on picking the number of shards to use?

jcmoraisjr:
The rule of thumb is: if the number of k8s-services/haproxy-backends is <500 you're fine and don't need to worry, otherwise use the result of `<num-svc> / 100-150` which will create shards enough to leave no more than about 150-200 backends per configuration file.

PR #667, opened 2020-09-15
Add better handling for cookie affinity with preserve option: add 'session-cookie-preserve' and 'session-cookie-value-strategy' options

griever989:
First off, wanted to say, great project and I appreciate you making this open source and all your hard work maintaining it. I didn't see a contributors file but I've attempted to make this submission as complete as possible. If you are interested in including these changes but want me to make any modifications, please let me know and I'd be happy to accommodate.

The use case that I had for adding these is when using the "preserve" option of cookie affinity, I want to allow the servers to assign their own cookie using Set-Cookie, and have haproxy simply route it to the server specified by a Set-Cookie header. An example may be something like a MMO server where there is a load balancer in front of many servers, but the users character should always be routed to the server their character exists on after they log in. This requires the servers to specify what cookie value is set, and it requires that cookie value to always be valid; it will break if a dynamic update was to reassign things in such a way that the cookie matched with a server changes.

This is supported by haproxy via the "preserve" option, which preserves the cookie set by the server.

Obviously for this to work properly, the servers as well as the ingress both need to agree on what the cookie values for a particular server are, which is easy to do as long as you use something accessible via the kubernetes Downward API, such as a pod name, or in the case I wanted, a pod UID.
- The haproxy config can set the cookie values to a known unique field that is accessible to the servers (eg. pod name, or a pod UID)
- When no cookie is present, haproxy will automatically route you to a server based on the `balance` mode.
- Any server can respond with a `Set-Cookie` to change the server, eg. say if it was a MMO server that wanted to redirect the client to which server their character is on after they log in.

Without any changes, this doesn't work when dynamic updating is on because the cookie will just be something like `srv001` until reloading. By making the dynamic updater understand when "preserve" cookie mode is enabled, we can tell it to not do a dynamic update (and do a reload instead) if it would cause a cookie value mismatch, fixing this, while allowing dynamic updates to still be used if the cookie matches.

Changes:
- add backend option `session-cookie-value-strategy` which allows changing how the cookie value is calculated. Right now it supports choosing between naming the cookie value as the endpoint Name (the default, and current behavior), and a pod UID.
- add backend option `session-cookie-preserve` which sets `preserve` option on the cookie. This was already achievable by setting `preserve` as part of `session-cookie-keywords`, however there wasn't a good way to make special behavior for it such as having different logic for dynamic updates without disabling dynamic updating altogether, and the existing documentation wasn't totally clear that dynamic updating breaks cookies when used like this. With this option, when dynamic updating, it knows to reload if a cookie value change is inevitable, making sure that servers are always assigned the correct cookie.
- documented the above
- added tests for the above

The changes in here don't affect backwards compatibility and aren't breaking:
- if "preserve" is set as "session-cookie-keywords" it keeps the old behavior and just logs a message suggesting to consider using the new option
- since the new preserve option is off by default, it doesn't affect dynamic reload behavior unless you turn it on
- the old cookie values for disabled servers were set to the default naming scheme, srv001, srv002, etc, and those are what got used when using dynamic update; this logic has been preserved when using the old existing configurations

Tested:
- ✔️ cookie affinity on, multiple servers, with and without the new preserve option and cookie naming UID option
- ✔️ cookie affinity on, dynamic cookie enabled, multiple servers
- ✔️ cookie affinity off, multiple servers
- ✔️ dynamic update with and without the new preserve option via scaling services up/down
- ✔️ `make test` all pass


jcmoraisjr:
Thank you for the kind words and for supporting this humble controller. Btw thank you also for such a detailed description and a carefully built pr.

Sorry about the long delay, I'll need a few more days to have a look in the changes. In the mean time there is some suggestions I'd like to share:

* there is a new `backend.CookieAffinity()` func that I think you planned to use in the template but didn't use;
* try to separate non functional but needed changes in another commit in order to clearly see what's a requisite and what's the functionality itself. You can eg create a new commit to change how dynupdate references an endpoint (name to object). Types+code+doc+tests however is preferable into a single commit;
* try also to use up to 50 chars in the commit title and up to 72 chars in the commit message. A commit message isn't mandatory but it's a good addition if isn't that clear what's the intention behind the change.

Regarding a contributions guideline it's a concept being built along the years writing the code but never translated to words. Just to point an example, the git commit title and message has about one week - this is both a [common practice](https://chris.beams.io/posts/git-commit/) and I observed that gitlab doesn't break lines in the commit preview, so we need to do it ourselves. I need to stop a few days to revise some docs for new users and new contributors, this should help in the visibility of the project.

griever989:
Thanks for the guidelines! I can definitely follow those. Did you want me to reconstruct the commits of this pull request directly by rewriting the history, or do you want me to make a new branch and PR with the rewritten commits? Some projects are ok with rewriting pull request branches but I just wanted to check with you first.

jcmoraisjr:
No, lets make it simple. Just a rebase is fine.

griever989:
Done. When recreating the commits, to make things simple, the following changed from the original submission:
- removed some unnecessary changes to the test endpoints in `pkg/haproxy/instance_test.go`, to keep number of changes down
- used the `backend.CookieAffinity()` func in `haproxy.tmpl`

pkg/converters/ingress/annotations/backend.go
```diff
@@ -61,7 +61,26 @@ func (c *updater) buildBackendAffinity(d *backData) {
 	}
 	d.backend.Cookie.Keywords = keywords
 	d.backend.Cookie.Dynamic = d.mapper.Get(ingtypes.BackSessionCookieDynamic).Bool()
+	d.backend.Cookie.Preserve = d.mapper.Get(ingtypes.BackSessionCookiePreserve).Bool()
 	d.backend.Cookie.Shared = d.mapper.Get(ingtypes.BackSessionCookieShared).Bool()
+
+	if strings.Contains(d.backend.Cookie.Keywords, "preserve") {
+		// just warn, no error, for keeping backwards compatibility where "preserve" may have been used in the "keywords" section
+		c.logger.Warn("session-cookie-keywords contains 'preserve'; consider using 'session-cookie-preserve' instead for better dynamic update cookie persistence")
```
jcmoraisjr, line 69: Add the source of the annotation here, something like "session-cookie-keywords on %s contains 'preserve'...". `d.mapper.Get(...).Source` has the source in string format and `%s` does not need to be quoted. This way one can easily find its ingress or service on a large k8s cluster. It'd suggest also to extract to a local var and use `.Bool()` and `Source` from there.

pkg/converters/ingress/annotations/backend.go
```diff
@@ -61,7 +61,26 @@ func (c *updater) buildBackendAffinity(d *backData) {
...
+	}
+
+	cookieStrategy := d.mapper.Get(ingtypes.BackSessionCookieValue).Value
+	switch cookieStrategy {
+	case "pod-uid":
+		d.backend.EpCookieStrategy = hatypes.EpCookiePodUid
+	case "server-name":
+		d.backend.EpCookieStrategy = hatypes.EpCookieName
+	default:
+		c.logger.Warn("invalid session-cookie-value-strategy '%s', using 'server-name' instead", cookieStrategy)
```
jcmoraisjr, line 79: Add source here as well.

pkg/converters/ingress/ingress.go
```diff
@@ -597,6 +611,27 @@ func (c *converter) addBackend(source *annotations.Source, hostname, uri, fullSv
...
+			switch backend.EpCookieStrategy {
+			default:
+				ep.CookieValue = ep.Name
+			case hatypes.EpCookiePodUid:
+				if ep.TargetRef != "" {
+					pod, err := c.cache.GetPod(ep.TargetRef)
+					if err == nil {
+						ep.CookieValue = fmt.Sprintf("%v", pod.UID)
+					} else {
+						c.logger.Error("error calculating cookie value for pod %s: %v", ep.TargetRef, err)
```
jcmoraisjr, line 627: `backend.AcquireEndpoint()` doesn't fill CookieValue, shouldn't be assigned to ep.Name in the case of an error reading the pod UID?

pkg/converters/ingress/ingress.go
```diff
@@ -597,6 +611,27 @@ func (c *converter) addBackend(source *annotations.Source, hostname, uri, fullSv
...
+func (c *converter) syncBackendEndpointCookies(backend *hatypes.Backend) {
+	cookieAffinity := backend.CookieAffinity()
+	for _, ep := range backend.Endpoints {
+		if cookieAffinity {
+			switch backend.EpCookieStrategy {
+			default:
+				ep.CookieValue = ep.Name
+			case hatypes.EpCookiePodUid:
+				if ep.TargetRef != "" {
+					pod, err := c.cache.GetPod(ep.TargetRef)
```
jcmoraisjr, line 623: Update `--disable-pod-list` command-line doc that the pod list is also being used by session-cookie-preserve as well.

jcmoraisjr (commented):
Thanks again for the contribution and sorry about the long delay. I needed to familiarize myself again with sticky session related options. Below has a few minor suggestions and doubts.

griever989:
Thank you for the review. I'll check it out and address the comments in a couple days.

pkg/converters/ingress/annotations/backend.go
griever989: Implemented in https://github.com/jcmoraisjr/haproxy-ingress/pull/667/commits/474d3c5f9bb044346cef58112f6e557844178696 

By `.Bool()` I am thinking you meant just `.Value` as it's a string value. Let me know if I misinterpreted that part.

pkg/converters/ingress/annotations/backend.go
griever989: Implemented in https://github.com/jcmoraisjr/haproxy-ingress/pull/667/commits/e7cf7f5ca3b779da2a376c299622235a0558c21d

pkg/converters/ingress/ingress.go
griever989: Implemented in https://github.com/jcmoraisjr/haproxy-ingress/pull/667/commits/0d58edb40287f5f1e19770498566e8776fb5cd50

pkg/converters/ingress/ingress.go
griever989: Implemented in https://github.com/jcmoraisjr/haproxy-ingress/pull/667/commits/099b1f64b6c5a5fc3c48514f9d2e671b826acd71

pkg/converters/ingress/annotations/backend.go
jcmoraisjr: Sure, yes, `.Value`. I did a little mess with `session-cookie-preserve` which would use `.Bool()`.

jcmoraisjr:
LGTM, thanks! Merging.

Issue #686, opened 2020-10-30
Configuration parsing failure when edit basic auth's usr/pwd 


jcmoraisjr:
**Description of the problem**

Configuration parsing break due to a missing userlist:

```
W1030 01:14:38.527364       6 instance.go:431] output from haproxy:
[ALERT] 303/011438 (106) : parsing [/etc/haproxy/haproxy.cfg:200] : unable to find userlist 't004_pwd' referenced in arg 1 of ACL keyword 'http_auth' in proxy 't004_echoserver_8080'.
[ALERT] 303/011438 (106) : Fatal errors found in configuration.
```

**Expected behavior**

All linked userlists should be created.

**Steps to reproduce the problem**

1. Configure basic authentication
2. Remove the basic auth configuration
3. Reinclude basic auth configuration
4. Edit the secret with users+passwords

**Environment information**

HAProxy Ingress version: `v0.11-beta.3`

jcmoraisjr:
Fixed in v0.11-beta.4

PR #688, opened 2020-10-31
Fix basic auth backend tracking

jcmoraisjr:
Basic auth users and passwords are stored in secret resources. We're properly tracking secret->userlist in the event of a secret update, and backends are also tracked but only when a new userlist is created.

A proper partial parsing need to track either a host or a backend from the haproxy model. If an untracked backend uses a tracked userlist, changes to this userlist will remove it without recreate, which will lead to failures in the configuration parsing.

This update enforces that a backend is always tracked in the event of a secret update, ensuring that partial parsing works properly.


jcmoraisjr:
#686

Issue #684, opened 2020-10-23
Acme signer reissue a certificate if private key isn't RSA


pbabilas:
**Description of the problem**

My ingress configuration is configured to work with acme cert generation, but when I set my own client custom cert I would like to not overwrite the client cert with new LE cert.
I've done tests and after apply new tls secret, haproxy seams to rewrite client cert with LE one. Can't find any reason of that :(

```I1021 07:35:32.150444       6 instance.go:102] starting certificate check (started leading)
I1021 07:35:32.150719       6 instance.go:110] finish adding 65 certificate(s) to the work queue
I1021 07:35:32.164514       6 signer.go:142] acme: authorizing: id=1 secret=aps/aps-tls domain(s)=apspolska.pl,www.apspolska.pl endpoint=https://acme-v02.api.letsencrypt.org reason='certificate does not exist'
I1021 07:35:35.134830       6 signer.go:147] acme: new certificate issued: id=1 secret=aps/aps-tls domain(s)=apspolska.pl,www.apspolska.pl
I1021 07:35:35.137018       6 backend_ssl.go:57] updating secret aps/aps-tls in the local store
```

**Expected behavior**

When I set custom cert, ingres should not rewrite this cert with LE cert.

**Steps to reproduce the problem**

1. apply any other cert to secret
2. wait until refresh is started
3. check secret was changed

**Environment information**

HAProxy Ingress version: `v0.11-snapshot.3`

Command-line options:

```
--configmap=ingress-controller/haproxy-ingress
--acme-server
--acme-track-tls-annotation
```

Global options:

```
acme-emails: pawel.babilas@merce.com
  acme-endpoint: v2
  acme-terms-agreed: "true"
  dynamic-scaling: "false"
  max-connections: "50000"
  prometheus-port: "9105"
  ssl-redirect-code: "301"
  use-proxy-protocol: "true"

```

pbabilas:
ping, anyone?

jcmoraisjr:
Hi, I'm working in some doc related stuff these days and could have a closer look into this issue. I'll ping back shortly.

pbabilas:
Thansk, for reply.
More details:

`I1022 07:32:53.641110       6 signer.go:142] acme: authorizing: id=5 secret=aps/aps-tls domain(s)=apspolska.pl,www.apspolska.pl endpoint=https://acme-v02.api.letsencrypt.org reason='certificate does not exist'
[...]

PR #689, opened 2020-10-31
Improve certificate validation of acme signer

jcmoraisjr:
A certificate validation need to parse a x509 certificate in order to identify if it's still valid or if it need to be reissued.

Tasks:

* [x] Add the root cause a certificate couldn't be read, parsed and analyzed;
* [x] Remove the parsing of the private key, which isn't used to analyze the certificate.


jcmoraisjr:
#684

PR #693, opened 2020-11-04
Allow signer to work with wildcard dns certs

pbabilas:
As in topic, this pr allows acme signer to work with wildcard dns in cert.


jcmoraisjr:
Hi, thank you so much for the contribution! Just a few suggestions:

* Please write against master, this will be cherry-picked later up to v0.9
* A wildcard domain `*.my.local` is valid for `sub.my.local` but it isn't for `other.sub.my.local` so this should be taken into account. `tls.Crt.VerifyHostname()` should be of some help here.

pbabilas:
ok, created new one:
https://github.com/jcmoraisjr/haproxy-ingress/pull/695

Thanks for hint with VerifyHostname!
Closing this one.

Issue #672, opened 2020-10-08
Implement missing pieces of v1 api


jcmoraisjr:
https://opensource.googleblog.com/2020/09/kubernetes-ingress-goes-ga.html

* [x] ~update to v1 api~ v0.13
* [x] ingress class resource
  * [x] controller name
  * [x] parameters
  * [x] ~default ingress class~ v0.13
* [x] ~resource backend~ v0.13

jcmoraisjr:
Closing and moving missing implementations to #706

PR #694, opened 2020-11-04
Add ingress class support

jcmoraisjr:
Follow the 1.18 ingress [spec](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class) and implement `ingressClassName` support.

This doesn't remove `kubernetes.io/ingress.class` annotation support and its behavior, which continue to have precedence in the case of a conflict.

**Breaking backward compatibility**

Starting v0.12, HAProxy Ingress will ignore unclassified ingress resources. Deployments that does not declare `--ignore-ingress-without-class` should now declare `--watch-ingress-without-class`. This configuration can be ignored if all ingress resources of the cluster are classified: has either the `kubernetes.io/ingress.class` annotation or configures the `ingressClassName` field.

**Tasks**

* [x] Strictly follow ingress spec - this breaks backward compatibility
* [x] Add ingress class lister and informer
* [x] Filter ingress resource based on `ingressClassName` attribute
* [x] Notify ingress class update
* [x] Track existing and missing ingress class
* [x] ~Add default ingress class support~
* [x] Add parameters support
* [x] Option to run more HAProxy Ingress instances in the same cluster
* [x] Unit tests
* [x] Exploratory tests
* [x] Documentation


jcmoraisjr:
#672

PR #695, opened 2020-11-04
Allow signer to work with wildcard dns certs

pbabilas:
As in topic, this pr allows acme signer to work with wildcard dns in cert.
Instead of https://github.com/jcmoraisjr/haproxy-ingress/pull/693

### DIFF
--- a/pkg/acme/signer.go
+++ b/pkg/acme/signer.go
@@ -125,7 +125,7 @@ func (s *signer) verify(secretName string, domains []string) (verifyErr error) {
 	duedate := time.Now().Add(s.expiring)
 	tls := s.cache.GetTLSSecretContent(secretName)
 	strdomains := strings.Join(domains, ",")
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Thanks for the contribution and also for the provided test!

I just realized that the current test can be extended to explore the wildcard scenario, what about add these new testcases in the array? You can use a distinct secret/crt changing the input accordingly and adding the new one in the cache mock.

pbabilas:
You are right! Changes just commited.

jcmoraisjr:
Finally travis released this, lgtm, thanks! Merging.

pbabilas:
ok, so when new version with docker image will be released? :)

jcmoraisjr:
v0.11 before the weekend, v0.10 and hope v0.12 next week, v0.9 this year.

pbabilas:
gr8! waiting ;]

PR #699, opened 2020-11-18
Adding new entries for networking.k8s.io/ingress

marcelo-devsres:
1.19 has ingress under networking.k8s.io, so with this RBAC, it fails.

### DIFF
--- a/examples/rbac/ingress-controller-rbac.yml
+++ b/examples/rbac/ingress-controller-rbac.yml
@@ -40,14 +40,22 @@ rules:
       - get
       - list
       - watch
+
   - apiGroups:
-      - "extensions"
+      - "networking.k8s.io"
     resources:
       - ingresses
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
Good catch, time to have a look on some old docs.

Why not just add a new apiGroup? Something like this:

```diff
   - apiGroups:
       - "extensions"
+      - "networking.k8s.io"
     resources:
       - ingresses
```

Instead of this:

```diff
+  - apiGroups:
+      - "networking.k8s.io"
+    resources:
+      - ingresses
+    verbs:
+      - get
+      - list
+      - watch
```

jcmoraisjr:
Hey there, we've definitely moved to the helm chart which is already updated, and we should clean up these provision artifacts in the short term. So no need to update here anymore. Closing.

Issue #669, opened 2020-09-30
Host-scope configuration question/feature request


Unichron:
Hi,

I have a bit of a niche problem. We need to manage tls configuration dynamically, because our service supports custom domains, meaning that when a new tenant is created we need to make haproxy aware of the new certificate/private key. For this currently there is a service that creates a new ingress. Now this creates a lot of redundancy, because that ingress resource needs to contain the same backend configuration over and over again. The backend config is of course static, but may sometimes change, making the it difficult to maintain, because the service component needs to be updated, and also already created resources need to be patched.

Ideally what I would like to be able to do, is configure the backend rules once (without host restriction), and have per host configuration like tls secret and host annotations somewhere else.

I know that the kubernetes api is a bit restrictive in this regard, because an ingress without a backend rule is not valid. After some trial and error I found out that the api allows for and ingress object like this:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: test
spec:
  rules:
    - host: ""
  tls:
    - hosts:
        - example.com
      secretName: ingress-tls-example.com
```

The problem is that haproxy ingress skips over it, because it doesn't contain the `http` part under the rule. Also I'm not sure that this road would lead to something that is not considered obscure and hackish.

An alternative solution that I thought about is if we could also define host-scope config in a configmap. In this case, the ingress would remain perfectly normal. We would have 1 ingress total. The tls config can be managed by appending/removing entries in this, and managing other host-scope config would be in separate configmap(s?).

What's your take on this?

jcmoraisjr:
Hi, we have a somewhat similar problem here and our ingress is configured like this:

```yaml
spec:
  rules:
  - host: domain.local
  tls:
  - hosts:
    - domain.local
    secretName: domain-crt
```

Is this syntax also useful for your usecase? Didn't test myself with two or more domains but this should work as well.

Unichron:
Hi,

I don't seem to make it work with this. How does this actually work for you? When I create an ingress like this tls doesn't work. I took a look at the sni maps and it's not there. Also I took a look at the code and it basically skips the whole tls config of such ingress. 

See ingress.go line 364:
```go
if rule.HTTP == nil {
   continue
}
```

jcmoraisjr:
Hi, sorry about the noise. You're absolutely right, the ingress is skipped if the host isn't configured. On our usecase the empty ingress is just used to annotate the domain to auto update the crt using acme.

[...]

PR #702, opened 2020-11-29
Independently configure rules and TLS

jcmoraisjr:
Since the very first controller version, the TLS spec was only configured if a corresponding hostname was found in the rules of the same ingress resource. This commit changes this behavior and reads the TLS array as an independent configuration.

Such update changes backward compatibility since a misconfigured TLS would be silently ignored. However the new behavior better implements the ingress spec.

### DIFF
--- a/pkg/converters/ingress/ingress.go
+++ b/pkg/converters/ingress/ingress.go
@@ -440,31 +440,30 @@ func (c *converter) syncIngress(ing *networking.Ingress) {
 				}
 			}
 		}
-		for _, tls := range ing.Spec.TLS {
-			for _, tlshost := range tls.Hosts {
-				if tlshost == hostname {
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
#669

Unichron:
Hi,

Thank you so much for picking this up! I have a question though. Would it be possible to include annotation processing as well when only TLS is configured? Technically a decent part of the annotations configure TLS behavior. This would remove sooo much complexity for us.

jcmoraisjr:
> Would it be possible to include annotation processing as well when only TLS is configured?

Yup, this is exactly how it should work - no matter where you configured a hostname, either in rules or just in tls, they should be added to the configuration in the same way.

Unichron:
Ah sorry I missed the part where you added that line. Thank you!

Issue #360, opened 2019-07-26
Enable OAuth on a per-path basis


speedst3r:
I have OAuth working successfully with [pusher/oauth2_proxy](https://github.com/pusher/oauth2_proxy) and keycloak, however there are some paths on my domain I do not wish to have OAuth over.

As a feature enhancement, it would be useful to be able to configure OAuth support and then separately enable it on a path basis using an annotation on a path-specific ingress.

jcmoraisjr:
Hi, there are some future plans to oauth config, which include choose protected paths and reuse a single oauth service.

jcmoraisjr:
Added in `v0.12-snapshot.3`. Closing.

PR #704, opened 2020-11-30
Change oauth2 to path scope

jcmoraisjr:
Configures OAuth2 from backend scope to path. This will allow to only configure OAuth2 in some paths of the same backend.

As of any backend -> path scope change, this should lead to backward compatibility change if the configuration has a previous conflict on its configuration.

Tasks:

* [x] Implementation
* [x] Unit tests
* [x] Update doc
* [x] Exploratory tests


jcmoraisjr:
#360

Issue #426, opened 2019-10-26
Annotations for different paths are being treated as the same


GMartinez-Sisti:
Hi,

I'm trying to do the following (redacted) with vbeta4:

```yaml
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-global
  annotations:
    ingress.kubernetes.io/limit-connections: "5"
    ingress.kubernetes.io/limit-rps: "15"
spec:
  rules:
    - host: site.com
      http:
        paths:
          - path: /
            backend:
              serviceName: http
              servicePort: http
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-login
  annotations:
    ingress.kubernetes.io/limit-connections: "2"
    ingress.kubernetes.io/limit-rps: "2"
spec:
  rules:
    - host: site.com
      http:
        paths:
          - path: /login
            backend:
              serviceName: http
              servicePort: http
```

However haproxy complains that:

```sh
annotation 'ingress.kubernetes.io/limit-connections' from ingress 'default/ingress-global' overrides the same annotation with distinct value from [ingress 'default/ingress-login']
```

This type of configuration worked with v0.7, but the new way the configuration is created places both of the ingresses in the same backend. One way to go around this would be to create a new service with a different name than `http` that would create a new different backend in the haproxy config for the `/login` path, but I would like to avoid having to create a new service every time I need different annotations.

Any suggestions?

Thanks

jcmoraisjr:
Hi, at this right moment configure a new service is the way to go. I'll have a look to configure more backend confs to a per path basis just like whitelist and many others. Btw this behaviour - configs to the whole backend / configs per path - need to be properly documented.

GMartinez-Sisti:
Thanks for the reply, so service it is.

Let me know if you need help with something, I'm getting pretty familiar with this awesome ingress.

Regards

jcmoraisjr:
Hi, I've improved the documentation (pending merge) to describe a new scope `Path`. I'm considering this issue done since the main problem here was a proper documentation. Please fill a new issue if you consider a good addition to migrate limits to a per path scope. I'd implement it to v0.12, but it uses sticky-tables and won't be that trivial to do in a proper way.

jcmoraisjr:
Doc updated with `Path` scope.

PR #705, opened 2020-11-30
doc: add path scope

jcmoraisjr:
Path scope exists since ages but wasn't properly documented. This commit adds how it works and also changes to Path scope all the configuration keys that in fact are from this group.

Should be merged to all doc versions.


jcmoraisjr:
#426

Issue #674, opened 2020-10-15
Haproxy persistence cookie attr support to append SameSite attributes


thyn:
I need to add attributes in persistence  cookie SameSite=None, Secure. How can I do it?

Also it should be used if request support ssl

Please review

https://github.com/haproxy/haproxy/issues/361

https://web.dev/samesite-cookie-recipes/

jcmoraisjr:
Thanks for the link. Starting on v0.11 you can workaround this using `session-cookie-keywords`, doc [here](https://haproxy-ingress.github.io/v0.11/docs/configuration/keys/#affinity). If you configure ssl-redirect as true, this will never be used on http requests.

jcmoraisjr:
Added in `v0.12-snapshot.3`. Closing.

PR #707, opened 2020-11-30
Add SameSite cookie attribute

jcmoraisjr:
Add a configuration key to enable `SameSite=<Lax|None>` on persistence cookie attribute. This is a syntax sugar to add `attr SameSite=None secure` in the `session-cookie-keywords` configuration key.


jcmoraisjr:
#674

Issue #690, opened 2020-11-01
First conciliation fails if external haproxy isn't running


jcmoraisjr:
**Description of the problem**

The first controller conciliation fails if using haproxy as a sidecar and the haproxy container starts after the controller.

**Expected behavior**

Controller should retry to connect to the haproxy master socket, or crash and restart if the socket doesn't exist or refuses connection.

**Steps to reproduce the problem**

1. force the controller to start before haproxy, eg removing haproxy image and forcing a rolling update

**Environment information**

HAProxy Ingress version: `v0.12-snapshot.1`

Command-line options:

```
--master-socket
```

jcmoraisjr:
Fixed in v0.12-beta.1. Closing.

PR #720, opened 2021-01-09
Fix first conciliation if external haproxy is not running

jcmoraisjr:
The very first config conciliation might happen before haproxy is properly running and listening to its master socket. This can happen if eg the haproxy image should be download and the controller image is present. The first conciliation will now wait until the admin socket can successfully answer `show info`, and will only exit to the loop after send a command without error or if the controller receives a sigterm.


jcmoraisjr:
#690

Issue #509, opened 2020-01-28
Blacklisting IP's from file


arakis:
Hello,
how can i use files like:
```
        # deny access from specific countries
        acl src_blacklist src -f /haproxy-config/blocked-ip-ranges.csv
        tcp-request connection reject if src_blacklist
```
My guess is, this is possible via config maps. How to reference them with the ` -f` argument?

jcmoraisjr:
Hi, in theory, if you deploy haproxy-ingress as a k8s object, you can:

1. create a configmap with this blacklist, one IP or CIDR on its own line
2. mount this configmap as a volume in the haproxy-ingress' pod template
3. reference the mounted dir/file using `-f /dir/file`
4. such -f configuration would be in the `config-backend` annotation or global configmap option

Note also that changing the blacklist wouldn't update haproxy, you'd need to wait the file content to be updated (this could happen in 10 seconds to 1 minute) and then trigger a reload eg updating max-connection with +/-1

`whitelist-source-range` in the global configmap is also an option but will create the same acl on every backend of your cfg.

arakis:
I successfully mounted a file /blocked-ip-ranges.csv (it's x.x.x.x/y format, one per line) into the pods of haproxy-ingress in kubernetes, checked via `kubectl exec -it --namespace ingress-controller haproxy-ingress-xxxxx -- cat /blocked-ip-ranges.csv`

Now, on the deployment of a specific service (consumer of haproxy-ingress), i added a "ingress.kubernetes.io/config-frontend"-section:

```
  annotations:
    kubernetes.io/ingress.class: "haproxy"

    ingress.kubernetes.io/config-frontend: |
      # deny access from specific countries
      acl src_blacklist src -f /blocked-ip-ranges.csv
      tcp-request connection reject if src_blacklist
```
It seems nothing is blocked. I guess the reason is the google loadbalancer, maybe the src is that IP? How to archive that the client's IP is blocked?

Btw, a `blacklist-source-range` option would be very usefull.

arakis:
i also tried `req.hdr_ip(X-Forwarded-For, 1)` instead of `src`. I know, that the my production pods are receiving `x-forwarded-for=<client-ip>, <another-ip>`, so even the `-1` should be correct.

I'm using docker image v.0.7.5.

arakis:
I was able to get it working:

mount config map as described above. You can do this with a small script like:
```
[...]

PR #722, opened 2021-01-10
Add deny access list and exception ip/cidr

jcmoraisjr:
Added new config key `denylist-source-range` to create a deny access list. It receives a list of IPs/CIDRs just like the whitelist, but matching source IPs will be denied instead.

`whitelist-source-range` was deprecated on behalf of `allowlist-source-range`. Backward compatibility is maintained.

Added also option to configure exception IPs/CIDRs to the allow and deny lists, allowing a fine grained access control configuration.


jcmoraisjr:
#509

Issue #716, opened 2021-01-07
I can not scrape haproxy and also HAProxy Ingress metrics from the prometheus side


efotopoulou:
I have followed the guidelines at https://haproxy-ingress.github.io/docs/getting-started/ and succesfully deployed a haproxy ingress controller. I can fetch haproxy metrics in prometheus format at the port 9105.

However i have difficulties at scrape haproxy and also HAProxy Ingress metrics succesfully from the prometheus side. I followed the guidelines at https://haproxy-ingress.github.io/docs/examples/metrics/ but i can not see any target a the prometheus.

![image](https://user-images.githubusercontent.com/1914256/103899056-7e8cbd00-50fe-11eb-90f8-a88932968a9b.png)

At the service discovery part my target labels are dropped
![image](https://user-images.githubusercontent.com/1914256/103899621-56518e00-50ff-11eb-908f-5e23b2cfec95.png)

while the configuration part is as follows:
```
global:
  scrape_interval: 10s
  scrape_timeout: 10s
  evaluation_interval: 1m
scrape_configs:
- job_name: haproxy-ingress
  honor_timestamps: true
  scrape_interval: 10s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  relabel_configs:
  - source_labels: [__meta_kubernetes_pod_label_run]
    separator: ;
    regex: haproxy-ingress
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_pod_container_port_number]
    separator: ;
    regex: "10254"
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_pod_node_name]
    separator: ;
    regex: (.*)
    target_label: hostname
    replacement: $1
    action: replace
  kubernetes_sd_configs:
  - role: pod
    namespaces:
      names:
      - ingress-controller
- job_name: haproxy-exporter
  honor_timestamps: true
  params:
    scope:
    - global
    - frontend
    - backend
  scrape_interval: 10s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  relabel_configs:
  - source_labels: [__meta_kubernetes_pod_label_run]
    separator: ;
    regex: haproxy-ingress
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_pod_container_port_number]
    separator: ;
    regex: "9105"
    replacement: $1
    action: keep
  - source_labels: [__meta_kubernetes_pod_node_name]
[...]

PR #723, opened 2021-01-11
docs/metrics: Fix prometheus config

jcmoraisjr:
Metrics doc depends on the controller pod selector to allow Prometheus find the controller. Controller labels was changed when upgrading to Helm deployment. This update changes also the label used in the Prometheus config.


jcmoraisjr:
#716

Issue #706, opened 2020-11-30
Implement missing pieces of v1 api


jcmoraisjr:
https://opensource.googleblog.com/2020/09/kubernetes-ingress-goes-ga.html

Tasks:

* [x] update to v1 api
* [x] default ingress class
* [x] ~~resource backend~~

PR #726, opened 2021-01-18
Update to networking.k8s.io/v1 api

jcmoraisjr:
This migration maps v1beta1 to v1 changes, without adding any new functionality:

- spec.backend to spec.defaultBackend
- backend.serviceName/Port to backend.service
- port (intOrStr) to port.name/number

Resource backend wasn't implemented.


jcmoraisjr:
#706

Issue #724, opened 2021-01-15
paths with distinct path types chooses the wrong backend


JanHolger:
**Description of the problem**
cert-manager creates an ingress with the pathType "ImplementationSpecific" which doesn't seem to be supported by haproxy-ingress yet so unless i manually change the pathType to Prefix, the acme challenge will fail. I've never seen this pathType before so i'm not sure if that's a problem with cert-manager but it did work totally fine with nginx-ingress so i guess this is a problem on your side.

**Expected behavior**
Properly handle those ingress objects so cert-manager does work

**Environment information**

HAProxy Ingress version: `v0.11.1`
cert-manager version: `1.1.0`

Ingress objects:

```
kind: Ingress
apiVersion: extensions/v1beta1
metadata:
  name: cm-acme-http-solver-bzgtf
  namespace: example
  labels:
    acme.cert-manager.io/http-domain: '4084921234'
    acme.cert-manager.io/http-token: '1019081234'
    acme.cert-manager.io/http01-solver: 'true'
  annotations:
    nginx.ingress.kubernetes.io/whitelist-source-range: '0.0.0.0/0,::/0'
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: >-
              /.well-known/acme-challenge/B1xLzUROPkdp1IkQKVyQ5mU2OFhMdxj0K6rRq1Labcd
            pathType: ImplementationSpecific
            backend:
              serviceName: cm-acme-http-solver-69c6s
              servicePort: 8089
```

jcmoraisjr:
Hi, HAProxy Ingress supports pathType since v0.11, doc [here](https://haproxy-ingress.github.io/docs/configuration/keys/#path-type). Kubernetes since v1.18 (or so) added pathType field whose default value is `ImplementationSpecific` if not declared. Ingress v1, added in Kubernetes v1.19, has pathType as a mandatory field.

I couldn't reproduce the issue, either using `ImplementationSpecific` or `Prefix` pathTypes. I created this same ingress resource with this same path in front an echoserver, curl it just like Let's Encrypt would do and haproxy sent the request to my backend.

Maybe you have other configurations around which broke this one. Please share with us if you create a minimal deployment that reproduces this issue, so with a reproducer we can provide a proper fix.

JanHolger:
[...]

PR #728, opened 2021-01-31
Fix path precedence of distinct match types

jcmoraisjr:
All hostnames and path matches, used eg to select a backend, uses haproxy's map to increase performance on clusters with hundreds of distinct hostnames+paths. This however doesn't work very well if paths of distinct match types are used - there is one map converter per match type and the configuration was classifying paths only under the same map file.

This commit checks path overlap of distinct match types and creates new map files if a precedence should be enforced. The config file and map files lay out remain the same if there isn't overlap or if they happen under the same match type.

Should be merged down to v0.11 when path types was first implemented.


jcmoraisjr:
#724

PR #731, opened 2021-02-04
Add support for native redirection of default backend

rikatz:
This PR adds support for native redirection.

There are some environments that instead of directing the all-catch traffic to a service/backend needs to direct to something external to Kubernetes (or to a Ingress Object that complies with some different rules).

This way we "teach" HAProxy that when a backend does not exists and also no default backend is configured, simple state to the user a redirect code with a new location.


docs/content/en/docs/configuration/keys.md
```diff
@@ -326,6 +326,8 @@ The table below describes all supported configuration keys.
 | [`cors-expose-headers`](#cors)                       | headers                                 | Path    |                    |
 | [`cors-max-age`](#cors)                              | time (seconds)                          | Path    |                    |
 | [`cpu-map`](#cpu-map)                                | haproxy CPU Map format                  | Global  |                    |
+| [`default-backend-redirect`](#default-redirect)     | Default Location to Redirect a user if the resource does not exists     | Global  |                    |
+| [`default-backend-redirect-code`](#default-redirect)         | Response Code to be used when redirecting a user to the default location | Global  | `302`                   |
```
jcmoraisjr, line 330: Realign the table, adding a space in the former and removing some of them in the later. Description can also be as short as `Location` and `HTTP status code` or something like this.

docs/content/en/docs/configuration/keys.md
```diff
@@ -1117,6 +1119,26 @@ See also:
...
+| `default-backend-redirect`       | `Global` |         |       |
+| `default-backend-redirect-code`  | `Global` | `302`   |       |
+
+Define a redirect location of the HAProxy for unknown resources.
+
+* `default-backend-redirect`: Defines a location in which Ingress should redirect 
+an user if the incoming request doesn't match any hostname, or the requested path 
+doesn't match any location within the desired hostname. An internal
+404 error page is used if not declared and also if `default-backend-service` was
+no configured on command line.
```
jcmoraisjr, line 1135: ... was not configured ...

rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -800,7 +800,11 @@ backend _acme_challenge
 #
 backend _error404
     mode http
+    {{- if $global.DefaultBackendRedir }}
+    redirect location {{ $global.DefaultBackendRedir }} code {{ $global.DefaultBackendRedirCode }}
+    {{- else }}
     http-request use-service lua.send-404
+    {{- end }}
```
jcmoraisjr, line 807: Remove the indentation from the `{{ cmds }}` above

jcmoraisjr (commented):
Thanks for the PR! Just some few cosmetic changes below.

rikatz:
Corrected everything as asked, already squashed the commits :)

jcmoraisjr:
Thanks! Merging.

Issue #710, opened 2020-12-13
Update x509 cert without reloading


jcmoraisjr:
**What are you trying to do**

Update a certificate without haproxy being reloaded

**What HAProxy Ingress should do or how it should behave differently**

Dynamically update x509 certs in the running haproxy instance without the need to reload.

jcmoraisjr:
Implemented in `v0.13.0-snapshot.1` which was just released. Closing.

PR #734, opened 2021-02-14
Add dynamic update of frontend's tls certificate

jcmoraisjr:
Evolve dynupdate to recognize changes to the frontend's TLS certificate and apply it using the runtime api. The update also happens even if was detected that a reload should also be performed, so the configuration will be as updated as possible if for any reason the reload fails.

This code updates the minimal HAProxy version to 2.2. HAProxy 2.1 also implements TLS cert update, but this version cannot update crt-lists with filters - which we use in the default certificate.


jcmoraisjr:
#710

PR #735, opened 2021-02-16
Add protocol to allow content sources other than secret

jcmoraisjr:
A backward compatible implementation to allow specify a protocol where only a secret name was being used. If, instead a secret name a `proto://content` is used, such content will be searched based on the protocol. If a protocol isn't provided, `secret` is used as default.

Actually only `file` protocol is implemented which expects an absolute path, but others like `cm/configmap`, `http/s` can also be implemented in the future.

Internal names wasn't changed and so the public interface - currently all namings reference "secret". Docs updated where the protocol can be used.

### DIFF
--- a/pkg/controller/cache_test.go
+++ b/pkg/controller/cache_test.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2021 The HAProxy Ingress Controller Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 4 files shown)


pkg/controller/cache_test.go
```diff
@@ -0,0 +1,75 @@
...
+			input:   "file:///tmp/data",
+			proto:   "file",
+			content: "/tmp/data",
+		},
+		// 5
+		{
+			input:   "file://",
+			proto:   "file",
+			content: "",
+		},
```
narg95, line 64: It would be great if you add a test for multiple files. It is not clear what is the expected format and the final behavior, is it “file://one, file://two”? Or “file://one, two”?
If the later then an example in the docs would be great. If the former I guess there is a bug because the second file will keep the prefix.

jcmoraisjr: The later is the correct, `file:///tmp/ca,/tmp/crl`, which corresponds to "a comma-separated list of filenames prefixed with `file://`". Such test belongs to the GetXX() funcs but actually cache is badly tested due to its coupling with k8s' listers. Need a small refactor here to allow such tests.

Regarding the doc - it's more like a preview, both the functionality and the documentation. It needs a proper section to be linked to instead to try to describe the details in the annotation's doc. I''ll evolve this before merging.

Issue #737, opened 2021-02-21
Crash waiting for the external haproxy instance


jcmoraisjr:
**Description of the problem**

Controller crashes while waiting for the external haproxy instance. Only happens when using haproxy as a sidecar, changing to the embedded approach stop the crashes.

**Expected behavior**

A crash must not happen.

**Steps to reproduce the problem**

Apparently happens only when haproxy is very busy and/or the host has noisy neighborhood - in our deployment calico node started to burn about half of the host's CPUs on all instances that crashed. This is the stack trace:

```
...
I0220 21:14:23.216368       6 instance.go:480] waiting for the external haproxy...
panic: runtime error: slice bounds out of range [:48] with length 38

goroutine 209 [running]:
github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils.buildProcTable.func2(0xc0339a53d8, 0x26, 0x20, 0x30, 0xc0339a53e8, 0x6)
	/home/travis/gopath/src/github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils/utils.go:132 +0x123
github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils.buildProcTable(0xc0339a5000, 0x3fe, 0x0)
	/home/travis/gopath/src/github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils/utils.go:145 +0x1e3
github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils.HAProxyProcs(0x7ffda2e92a4e, 0x1c, 0x0, 0xc010c69cd8, 0x1)
	/home/travis/gopath/src/github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/utils/utils.go:99 +0x192
github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy.(*instance).reloadExternal(0xc00048db00, 0xc000371900, 0x1)
	/home/travis/gopath/src/github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/instance.go:500 +0x184
github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy.(*instance).reload(0xc00048db00, 0x1882650, 0xc)
	/home/travis/gopath/src/github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/instance.go:455 +0x62
[...]

PR #739, opened 2021-02-24
Read the whole input when the response fills the buffer

jcmoraisjr:
HAProxy responses has just a few bytes, usually saying that a backend server was successfully updated. A 1k buffer is just enough to handle such responses. Since the external process implementation, one of the responses has a list of all running procs, which will easily fill that buffer depending on the number of old workers alive. These updates:

* continue reading the buffer up to the last byte.
* validate the length of incoming response, leading to zeroes or empties if truncated - instead of crash


jcmoraisjr:
#737

Issue #741, opened 2021-02-27
Initial weight not configured correctly


mhyllander:
I have set `initial-weight` to "100" on an Ingress, but the `server-template` in the generated config has `weight 1`. 

When I configure:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/agent-check-port: "9991"
    ingress.kubernetes.io/initial-weight: "100"
    ingress.kubernetes.io/timeout-tunnel: 60s
    ingress.kubernetes.io/use-resolver: kubernetes
```

I get the following config in haproxy.cfg:

```
backend platform_re-worker-service_websocket
    mode http
    balance roundrobin
    timeout tunnel 60s
    acl https-request ssl_fc
    http-request redirect scheme https if !https-request
    option forwardfor
    http-response set-header Strict-Transport-Security "max-age=15768000"
    server-template srv 100 _websocket._tcp.re-worker-service.platform.svc.cluster.local resolvers kubernetes resolve-prefer ipv4 init-addr none weight 1 check inter 2s agent-check agent-port 9991
```

Notice that it has "weight 1".

I expected it to be "weight 100".

jcmoraisjr:
Hi, I can confirm this behavior. The initial weight config is being written before it can be read from the ingress resource. You can work around it annotating the service resource instead.

mhyllander:
Thanks. I guess I'll have to split the service in two, since I don't want the configuration to apply to all server endpoints.

jcmoraisjr:
Yep, already fixed and planning a new round of releases this weekend. Maybe worth to wait a little more. Btw thanks for reporting.

jcmoraisjr:
Fixed in v0.10.5, v0.11.4 and v0.12.1. Closing.

PR #742, opened 2021-02-27
Fix initial weight configuration

jcmoraisjr:
Weights can be used to distribute distinct load to distinct servers in the same backend. An initial value greater the default 1 (one) can be configured so agents that changes server weight can properly work and adjust the weight in lower granularity.

The initial weight configuration isn't properly working when declared in the ingress resource because its usage occurs before the value is actually read from the ingress annotation. It works however when declared in the service - service has priority and is read in the same block the backend is created and configured with initial weight and other defaults. Moving down the reading just before starting the endpoint configuration fixes this.


jcmoraisjr:
#741

Issue #745, opened 2021-03-01
Multiple reloads of haproxy config without any apparent changes


mhyllander:
After recently migrating from haproxy-ingress 0.7.6 to 0.12.1 (!), I saw a change in behavior when rolling out a new version of a service deployment. haproxy reloaded it's configuration multiple times during the rollout, apparently without any changes to the configuration.

We have a service with very long-lived websocket connections. Therefore we have configured haproxy-ingress to avoid restarting and dropping those connections as much as possible. We are using reusesockets, dynamic scaling and DNS resolvers. We also set `timeout-stop` to 24h so that existing connections are kept alive, so that most clients have time to re-connect during the next 24h. (If we didn't do this we would get a burst of clients re-connecting at the same time.)

What happened is that our service with 30 pods was rolled out with 6 pods at a time. At each rollout step, haproxy forked another instance, so the haproxy-ingress pod memory usage grew in steps, and of course the count of current connections restarted from zero each time.

Note also that I am running an "external" haproxy in a side-car.

**Expected behavior**

Since backend server discovery is done through DNS lookups, haproxy.cfg does not change, so there should not be a need to restart haproxy. (Even if we didn't use DNS lookups, dynamic scaling should have allowed updates without restarting haproxy.) This is the way it used to work in 0.7.6.

**Steps to reproduce the problem**

 

**Environment information**

HAProxy Ingress version: `v0.12.1`

Command-line options:

```
  - args:
    - --configmap=core/re-ingress-haproxy
    - --ingress-class=re
    - --master-socket=/var/run/haproxy/master.sock
    - --sort-backends
    - --default-backend-service=core/re-ingress-haproxy-default-backend
    - --default-ssl-certificate=core/re-ingress-cert
    - --reload-strategy=reusesocket
    - --watch-ingress-without-class=true
```

Global options:

```
data:
  backend-server-slots-increment: "100"
  config-frontend: |
    #-------------------------------------------------------------
    capture request header Host len 40
    capture request header User-Agent len 100
    capture request header X-Forwarded-For len 100
    #-------------------------------------------------------------
  dns-hold-obsolete: 30s
  dns-hold-valid: 30s
  dns-resolvers: kubernetes=10.128.0.10
  dynamic-scaling: "true"
  forwardfor: update
  healthz-port: "10253"
[...]

PR #746, opened 2021-03-02
Fix incorrect reload if endpoint list grows

jcmoraisjr:
Dynamic updates need to check the size of the endpoint list size. Up to now the backend server list cannot dynamically grow, even when using DNS updates. HAProxy Ingress has a list of empty slots to dynamically update endpoints, forcing a reload only when no new empty slots can be found.

Kubernetes' ep based updates always rebalance new endpoints with empty slots so its size is at least as big as the old one, but DNS based only do this if a reload is needed. This update rebalance the new list, so future DNS based dynamic updates can work without needing to reload haproxy.


jcmoraisjr:
#745

Issue #413, opened 2019-10-18
improved oauth support


billimek:
Currently the [oauth documentation](https://github.com/jcmoraisjr/haproxy-ingress#oauth) and [example](https://github.com/jcmoraisjr/haproxy-ingress/tree/master/examples/auth/oauth) suggest that it is only possible to configure an ingress to use oauth by configuring a specific ingress backend for the `/oauth` (or configured `oauth-uri-prefix` path).

This limits the ability to leverage oauth with purely annotations as is [currently possible with something like the nginx ingress controller](https://github.com/kubernetes/ingress-nginx/tree/master/docs/examples/auth/oauth-external-auth).

It would be good if we could do oauth configuration purely with annotations when using the haproxy ingress controller.

jcmoraisjr:
Hi, do you mean authenticate requests only on paths whose ingress objects
has the oauth annotation? This is the current behavior since v0.8, pleaae
have a look and let me know if you'd expect something different.

~jm


Em qui, 17 de out de 2019 22:57, Jeff Billimek <notifications@github.com>
escreveu:

billimek:
Hi @jcmoraisjr, I think I may mean something different or perhaps I don't understand how to use it properly now.  What I mean is this:

For a given ingress object, to use haproxy and oauth together, the ingress object must look something like this snippet:

```yaml
  annotations:
    ingress.kubernetes.io/oauth: oauth2_proxy
  name: some-app
spec:
  rules:
  - host: some-host
    http:
      paths:
      - backend:
          serviceName: oauth2-proxy
          servicePort: 4180
        path: /oauth2
      - backend:
          serviceName: some-app-service
          servicePort: 8080
        path: /
```

The part above about about the `oauth2-proxy` and `/oauth` path are net-new required parts of the ingress itself beyond the application backend. From what I understand, haproxy expects there to be a `/oauth2` path under the application host URL.

In the nginx ingress world, we reference an oauth2 as simply annotations to the ingress object, like so using the above snippet:

```yaml
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/auth-url: "https://some-host/oauth2/auth"
    nginx.ingress.kubernetes.io/auth-signin: "https://some-host/oauth2/start?rd=$escaped_request_uri"
  name: some-app
spec:
  rules:
  - host: some-host
    http:
      paths:
      - backend:
          serviceName: some-app-service
          servicePort: 8080
        path: /
[...]

PR #748, opened 2021-03-05
Add external authentication

jcmoraisjr:
This update adds the ability to configure an URL used to authenticate a request. Every single request passes through auth-request.lua -> haproxy-lua-http.lua -> frontend listening a local tcp port -> target. The local port uses an unassigned range by default and can be configured. An optional signin redirect can be configured in the case of an authentication failure, useful with oauth2.

`auth-type` configuration key was deprecated, more than one authentication types can be configured at the same time to the same path. This however doesn't change backward compatibility, basic authentication has the secret name as mandatory.

`auth-request.lua` was updated, breaking compatibility when oauth-headers is used - the variable name with the email now is `req.auth_response_header.x_auth_request_email` instead of `[txn.]auth_response_email`. This will have no impact if neither the header nor the script were configured.

OAuth2 example updated, changing from the deprecated Bitly version to `https://github.com/oauth2-proxy/oauth2-proxy`

## Compatibility notes

* The haproxy variable with the logged user email was renamed. This will only affect oauth2 configurations that changed the `oauth2-headers` without changing the `auth-request.lua` script, or changed the `auth-request.lua` script without changing the `oauth2-headers`.
* The `auth-request.lua` dependency changed from `lua-socket` to `lua-json4`, this will only impact deployments that uses an external haproxy instance.
* `auth-type` was deprecated and doesn't need to be used to configure basic authentication. This will only impact users that configures the `auth-secret` without configuring the `auth-type` - the basic authentication config will be applied.

Backlog:

* [x] `auth-url` and `auth-signin` config keys
* [x] Optional `_front__auth` logging
* [x] Remove unused local ports and helper backends
* [x] At least a new protocol (https or backend) for security (implemented both)
* [x] Migrate oauth2 config keys
* [x] Unit testes
* [x] Exploratory tests
* [x] Documentation


jcmoraisjr:
#413

Issue #581, opened 2020-05-13
Add a non scoped configuration snippet - eg for cache


vitobotta:
Hi!

Is it possible to implement microcaching with an ingress? I am looking to cache pages for 5-10 seconds to reduce load on my app. I can do this with nginx but I am not sure of how to do it with haproxy ingress. Thanks!

jcmoraisjr:
Hi, this is currently not possible due to the lack of a way to create a cache section. You can however overwrite the template file, add a hardcoded [cache section](https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#10) and configure cache-use/cache-store via [`config-backend`](https://haproxy-ingress.github.io/docs/configuration/keys/#configuration-snippet) config key. Added to the backlog.

vitobotta:
Hi @jcmoraisjr , gotcha. I was hoping I could do this with haproxy out of the box :)

jcmoraisjr:
Hi, reopening so this won't be forgotten.

jcmoraisjr:
Implemented in `v0.13.0-snapshot.1` which was just released and adds a way to create new haproxy sections in the config file, as well as other custom config options. Closing.

PR #749, opened 2021-03-06
Add custom-sections global option

jcmoraisjr:
The `custom-sections` global configuration key adds unindented configuration in the config file, giving the ability to create new HAProxy sections.


jcmoraisjr:
#581

Issue #136, opened 2018-04-08
TCP services support using annotations


wenygao:
With tcp-services-configmap, it seems the all traffics to one specific port will route to one fixed backend service. If we need to route to different backend services from same port, how can I do it. For example, there are many backend postgresql services listening on port 5432, the haproxy need to route traffic to postgresql serivces according to host name respectively.

I see the ingress-nginx has already a below solution to meeting the above requirement, is there a equivalent solution in haproxy-ingress?

https://github.com/kubernetes/ingress-nginx/issues/153

jcmoraisjr:
Hi. That proposal was to add an annotation to the ingress resource - which wasn't implemented - and not to support name based routing. Besides that I don't have a clear idea if this is even possible - perhaps using the sni extension from a tls connection - which I also don't know if a psql client support. Afaik using plain tcp does not have support to name based routing. Any help would be welcome here.

jodem:
@jcmoraisjr I agree it's not possible with the hostname, but I would love to see "service port" support. I mean, If we could associated a service port to an ingress to enable a tcp backend it would allow us to configure tcp routing 'per app' instead of having to do it globally with the tcp-service-configmap. It is what they do also in voyager  : https://appscode.com/products/voyager/6.0.0/guides/ingress/tcp/overview/

jcmoraisjr:
Finally implemented a way to use the rich HAProxy Ingress' configuration key list with TCP services. Now TCP services can be configured as an ingress resource. Closing.

PR #750, opened 2021-03-08
Add ingress based TCP service option

jcmoraisjr:
Implementation of TCP services using ingress resources. The same backends used by http and passthrough requests can be used to TCP services as well, which means that any backend configuration that doesn't need L7 can be used - eg whitelist and blue/green deployments are supported, while cookie based sticky session and hsts are not. All configuration keys will be updated, describing its compatibility with TCP services.

The old configmap based configuration is now deprecated, but will continue to work without changes or backward compatibility issues.

Backlog:

* [x] haproxy model implementation
* [x] ingress converter implementation
* [x] ~host annotation parser~
* [x] SNI support
* [x] haproxy model tests
* [x] ingress converter tests
* [x] partial update tests
* [x] exploratory tests
* [x] docs - tcp service, compatibility, deprecation note


jcmoraisjr:
#136

Issue #751, opened 2021-03-10
Path Matching with Wildcard Host


yeddahbi:
**Description of the problem**

Hello there,

Is it possible to do path matching while using wildcard host using haproxy-ingress v0.10.5?
We are having trouble working with both. 
We have 2 ingresses with the same wildcard host that are pointing to the same backend. However, we need to rewrite to a target path for one of the ingresses. See example below. Note that we have tried to the same configuration with a regular hostname, and it worked fine.

Not sure I fully understand what is happening, but from what I understand, the rewrite never happens because the path mapping to path1 is not working since the map file contains the wildcard hostname and the mapping is done using the “map_beg” function that does string matching.
https://github.com/jcmoraisjr/haproxy-ingress/blob/release-0.10/rootfs/etc/haproxy/template/haproxy.tmpl#L333

**Expected behavior**
Path matching/routing would work using a wildcard host.

**Steps to reproduce the problem**
The generated configuration for backend looks like this:

```
backend example-net-namespace_webapi- example-net-service_srv-port
    mode http
    balance roundrobin
    acl https-request ssl_fc
    # path01 = *.example.net/v2
    # path02 = *.example.net/status
    http-request set-var(txn.pathID) base,lower,map_beg(/etc/haproxy/maps/_back_example-net-namespace_webapi- example-net-service_srv-port_idpath.map)
    http-request set-header X-Original-Forwarded-For %[hdr(x-forwarded-for)] if { hdr(x-forwarded-for) -m found }
    http-request del-header x-forwarded-for
    option forwardfor
    http-request replace-path ^/v2/(.*)$       /to/another/path/v2\1     if { var(txn.pathID) path01 }
    http-response set-header Strict-Transport-Security "max-age=15768000"
    server srv001 10.10.11.111:8443 weight 1 ssl verify none check inter 2s
    server srv002 10.10.11.112:8443 weight 1 ssl verify none check inter 2s

```
Here is the content of the map file :

```
/ # cat /etc/haproxy/maps/_back_example-net-namespace_webapi- example-net-service_srv-port_idpath.map
# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# #
# #   HAProxy Ingress Controller
# #   --------------------------
# #   This file is automatically updated, do not edit
# #
#
*.example.net/status path02
[...]

PR #752, opened 2021-03-14
Fix backend matches if hostname uses wildcard

jcmoraisjr:
Some hostname+path matches happen in the backend to identify if a configuration should or should not be applied. Examples are whitelist, cors, hsts and rewrite path. Frontend matches are correctly using a regex if a wildcard hostname is used. Backends also work but only if all paths share the same configuration, in this case a path ACL is not used. However backends that have distinct path configurations were not properly configuring regex based ACLs - they were trying to match wildcard hostnames with -m begin.

This update adds to the backend the same hostname check used in the frontend, which converts begin match to a regex match if a wildcard hostname is used.

### DIFF
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -1189,6 +1189,87 @@ frontend _front001
 	}
 }
 
+func TestMatchRegexBackend(t *testing.T) {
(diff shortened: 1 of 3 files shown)


jcmoraisjr:
#751

Issue #715, opened 2021-01-03
Provide redirection to primary hostname when server-aliases are defined


technicalguru:
**What are you trying to do**

I want to provide canonical URIs. That means that different server aliases will automatically redirect to the primary hostname. 

**What HAProxy Ingress should do or how it should behave differently**

Mark one hostname as the primary hostname, e.g. via annotation, and redirect all clients to this hostname when they request one of the defined server aliases.

jcmoraisjr:
Implemented in `v0.13.0-snapshot.1` which was just released. Closing.

PR #754, opened 2021-03-17
Add server redirect options

jcmoraisjr:
A server redirect allows to configure source domains that, if provided in a request, should redirect to the domain configured in the ingress' spec. Valid hostnames will always have precedence, which means that in the case of a match, the redirect will not happen.

As a host scoped config, two new maps were created (redirect source and status code) which will only be used if the functionality is also used.


jcmoraisjr:
#715

Issue #587, opened 2020-05-19
Add config snippet on other haproxy sections


thyn:
Currently it's possible to add custom logic to http frontend and backend

like

`{
	"config-frontend": "acl network_allowed src -f /etc/haproxy/lists/allowedips.lst
		http-request deny if !network_allowed
		"
}`

It works only for http(s).

But it's impossible to do it in stats and tcp frontends. Also if ssl pass through is enabled all such sites are not filtered.

Please, add such extensions points.

jcmoraisjr:
Related with #581

jcmoraisjr:
Implemented in `v0.13.0-snapshot.1` which was just released and adds lots of new custom configuration options. Closing.

PR #755, opened 2021-03-17
Add custom-proxy configuration

jcmoraisjr:
Allows to configure any internal proxy - listen, frontend or backend - using a single configuration key as a multi section value. Each section of the configuration adds custom snippet to distinct proxies.


jcmoraisjr:
#587

Issue #753, opened 2021-03-15
Request path that contains a hostname results in wrong routing


chitoku-k:
**Description of the problem**

Let's say we have the ingress in the last section of this issue. This ingress only specifies that it should route traffic by its hostname; however, when the request path contains a subdomain, it routes to a wrong one as in the following:

OK: example.com goes to nginx-alice
```
$ curl example.com
nginx-alice-7f9b8cd46d-8rll5
```

OK: subdomain.example.com goes to nginx-bob
```
$ curl subdomain.example.com
nginx-bob-848d7b78d9-sv4cv
```

WRONG: subdomain.example.com/example.com goes to **nginx-alice**
```
$ curl subdomain.example.com/example.com
nginx-alice-7f9b8cd46d-8rll5
```

**Expected behavior**

example.com goes to nginx-alice
```
$ curl example.com
nginx-alice-7f9b8cd46d-8rll5
```

subdomain.example.com goes to nginx-bob
```
$ curl subdomain.example.com
nginx-bob-848d7b78d9-sv4cv
```

subdomain.example.com/example.com goes to **nginx-bob**
```
$ curl subdomain.example.com/example.ocm
nginx-bob-848d7b78d9-sv4cv
```

**Steps to reproduce the problem**

1. Deploy the following manifest to a Kubernetes cluster:
    (It contains two nginx Pods that print each Pod name respectively whatever path it is given)
    ```yaml
    ---
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: nginx-conf
      namespace: default
    data:
      default.conf.template: |
        server {
          listen 80 default;
          listen [::]:80 default;
          server_name _;
          keepalive_timeout 0;
    
          location / {
            add_header Content-Type text/plain;
            return 200 "$BODY\n";
          }
        }
    ---
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: nginx-alice
      namespace: default
    spec:
      replicas: 1
      selector:
        matchLabels:
          app: nginx-alice
      template:
        metadata:
          labels:
            app: nginx-alice
        spec:
          containers:
            - name: web
              image: nginx:1.19.7-alpine
              env:
                - name: BODY
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.name
              volumeMounts:
                - name: nginx-conf
                  mountPath: /etc/nginx/templates
              ports:
                - name: http
                  containerPort: 80
          volumes:
            - name: nginx-conf
[...]

PR #756, opened 2021-03-18
Fix prefix path type if the path matches a domain

jcmoraisjr:
Map keys are built using sysadmin/user defined hostnames and paths. We concatenate both when building the map file, and also concatenate hdr(host) and path from the incoming request. This works pretty well on beg and reg match types. Since v0.11 we have also dir (Prefix from ingress pathType) and str (Exact). The later also work well with this construction, but str (Prefix) does not behave as expected from the end user perspective.

Take this map example:

```
   domain.local/ backend1
   sub.domain.local/ backend2
```

and this request:

```
   sub.domain.local/domain.local
```

A begin or regex match type would choose backend2, and this is the expected behavior for prefix as well, but due to how haproxy implements -m dir, backend1 would be chosen.

When using dir match type, haproxy slices the request using the slash as separator. These slices tries to match a whole map entry, but without starting from the beginning.

A hash # is added after the hostname and before the first slash, leaving the first item of the slice as unique. This is only needed for dir / Prefix, but will not hurt other match types and does not have any performance degradation.

Now the map looks like this:

```
   domain.local#/ backend1
   sub.domain.local#/ backend2
```

and the request is internally converted to this:

```
   sub.domain.local#/domain.local
```

Should be merged as far as v0.11.

### DIFF
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -140,8 +140,8 @@ func TestBackends(t *testing.T) {
     http-response set-header Access-Control-Allow-Credentials "true" if { var(txn.pathID) path01 }`,
 			expCheck: map[string]string{
 				"_back_d1_app_8080_idpath__begin.map": `
-d1.local/sub path02
-d1.local/ path01`,
+d1.local#/sub path02
+d1.local#/ path01`,
 			},
 		},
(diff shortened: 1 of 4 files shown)


jcmoraisjr:
#753

PR #757, opened 2021-03-18
Add custom-tcp configuration

jcmoraisjr:
Allows to customize configmap based TCP services with configuration snippets. All TCP services receives the same configuration.


jcmoraisjr:
#587

Issue #773, opened 2021-04-28
Ingress without host and rewrite breaks when adding rule with host without rewrite


gillius:
**Description of the problem**

Add an ingress with URL rewriting and no host specified, and it will work as expected. Then add an ingress with host for unrelated domain and without rewrite and the first ingress rule will stop rewriting.

**Expected behavior**

A host-specific ingress rule should not affect behavior of a non-host-specific ingress rule when facing a request with a Host header not matching the domain.

**Steps to reproduce the problem**

What I have is 2 ingresses for same service, one is no host with a rewrite the other is a specific host. When I apply only the first ingress, the rewrite works. When I apply second ingress, so both are active, then rewrite stops working on the first. The traffic is routed but the path is not rewritten.

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: echoserver-1
  namespace: echoserver
  annotations:
    ingress.kubernetes.io/rewrite-target: /
    kubernetes.io/ingress.class: haproxy
spec:
  rules:
    - http:
        paths:
          - path: /example/echoserver
            pathType: Prefix
            backend:
              service:
                name: echoserver
                port:
                  number: 80
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: echoserver-2
  namespace: echoserver
  annotations:
    kubernetes.io/ingress.class: haproxy
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /echoserver
            pathType: Prefix
            backend:
              service:
                name: echoserver
                port:
                  number: 80
```

With echoserver-1 only in server, I run through kubectl proxy:

```
GET http://localhost:59514/example/echoserver/123
Host: localhost
```

And I see real `path=/123` as expected. After adding resource echoserver-2, I run same http request as above but now I get real `path=/example/echoserver/123.`

When I use the domain:

```
GET http://localhost:59514/echoserver/123
Host: example.com
```

it returns `real path=/echoserver/123` as expected (no rewriting) given the echoserver-2 rule.
I would not have expected echoserver-2 resource to impact echoserver-1 at all because the host of echoserver-2 (example.com) does not match the localhost request.

**Environment information**

The echoserver pod is image `gcr.io/google_containers/echoserver:1.4` and environment is AWS EKS 1.19, haproxy-ingress version is 0.12.3 deployed via Helm.

** Workaround **

[...]

PR #777, opened 2021-05-03
Fix per path filter of default host rules

jcmoraisjr:
This update fixes two issues in rules with default host (empty hostname)

The first one is in the backend, the per path matching was trying to match the hostname with an internal representation. In this case only the path should be matched after the non empty hosts has been failed.

The other one is in the haproxy's frontend, the path was being matched with path_beg despite of the path type configuration. Now default host match uses a map just like non empty hosts.


jcmoraisjr:
#773

Issue #770, opened 2021-04-22
Using "Exact" pathType and "Prefix" pathType for the same path ignores one of them


mahcinek:
**Description of the problem**

I'm trying to point to two backend resources using the ingress controller. One of them should be accessible on the root path and the other should be accessible on any other path. Unfortunately the ingress controller seems to ignore one of the path configurations (I believe it's always the later one in the ingress) which use the standard Kubernetes notation ("Exact" and "Prefix").

**Expected behavior**

Ingress controller points to the be-1 service on the root path and to be-2 on any other path.

**Steps to reproduce the problem**

1.  Create two different pods
2.  Create two services that expose them
3.  Use an ingress configuration similar to the one provided here

**Environment information**

HAProxy Ingress version: `v0.13`

Command-line options: default

Global options: default

Ingress objects:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: haproxy
  name: example-ingress
  namespace: default
spec:
  rules:
  - http:
      paths:
      - pathType: Exact
        path: /
        backend:
          service:
            name: be-1
            port: 
              number: 80
      - pathType: Prefix
        path: /
        backend:
          service:
            name: be-2
            port: 
              number: 80
```

jcmoraisjr:
Thanks for reporting this use case. Due to some internal premises this wasn't that trivial to fix, but was properly handled and should be in the next v0.13 snapshot.

This should be merged to v0.12 and v0.11, which also implements path type, but I'm a bit reluctant to do this. The update is rather big and, specially, it might bring a compatibility change on clusters wrongly configured - paths hidden behind a duplication will eventually be configured and change how requests are handled. I'll check what the community says, maybe this can be done under a configurable flag.

jcmoraisjr:
Fixed in `v0.13.0-snapshot.2`, closing.

PR #779, opened 2021-05-08
Add the ability to use the same host+path more than once

jcmoraisjr:
Since the `pathType` implementation, host+path cannot be considered unique because distinct path types can be used to the same path - mostly exact with begin or prefix. This update adds the ability to configure distinct backends to the same URI, provided that they have distinct path types.

This same update also improves the haproxy's map files management, not only to the visible functionality, but also to the future Gateway API implementation - the Gateway API allows matching not only by hostnames and paths, but also by HTTP headers, requiring further control in the match type and priority in the case of overlaps.

Exact path type priority was also changed: since an exact path type never overlaps, now all of them are processed before any other type. This will ensure that an `exact` path `/dir` would handle a `/dir` request before a `prefix` path `/dir`.

# Compatibility notes

Clusters with duplicated hostname and path can generate distinct configuration. Before this update a cluster configuration with duplicated paths would configure just one of them - the one that was created first. Now duplicated paths will be configured, provided that they have distinct path types, and this can create distinct configuration if they use distinct backends. If one of the paths is exact, it will always be checked before the others. If prefix or begin it depends on the `path-type-order` configuration - by default prefix will be processed first.


jcmoraisjr:
#770

PR #782, opened 2021-05-11
Add option to copy headers to and from auth external

jcmoraisjr:
Add three new configuration keys used to copy headers to and from the external authentication service. The default behavior is to copy all the headers and the configuration accepts a comma-separated list of a wildcard like syntax. This syntax is converted to a Lua Pattern and used to match incoming and outgoing headers.

`auth-request.lua` was adapted to this functionality with a new action name that receives the headers that should be copied. It's pretty much cheaper easier to make the Lua script implement all the new behavior compared with implementing static configuration in the HAProxy side.


jcmoraisjr:
#413

Issue #785, opened 2021-05-25
Solve/silence deprecation warnings coming from the api server


Unichron:
Hi!

Since k8s 1.19 the api server can attach a Warning response header to any response. One use-case of this is to communicate api deprecation warning to the client. The go-client library by default is quite noisy in this regard. E.g. I see this in our cluster, basically spamming (although the frequency is low but still...):

```
W0525 11:27:43.486752 8 warnings.go:67] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0525 11:32:23.388293 8 warnings.go:67] networking.k8s.io/v1beta1 IngressClass is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 IngressClassList
```

I guess for backward compatibility reasons it might not be feasible to rewrite the code to always use the V1 version (correct me if I'm wrong), but it would be still nice to hide this implementation detail from the ingress users. Fortunately the go client can be customized to hide the warnings or at least dedup them: https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

jcmoraisjr:
v0.12.4 was just released and has now a `--disable-api-warnings` comand-line option. Closing.

PR #789, opened 2021-05-30
Add option to disable API server warnings

jcmoraisjr:
Most of the API server warnings are related with API deprecation, and these logs are generated every now and then, flooding controller logs from clusters that have few syncs. This option adds the ability to disable these logs.


jcmoraisjr:
#785

Issue #790, opened 2021-06-01
Domain name validation regex not allowing "-"


mhyllander:
**Description of the problem**

I tried to use the `secure-sni` configuration key and set it to `data-graph.azurewebsites.net`. I got an error 

```
W0601 09:49:40.915372       6 backend.go:646] skipping invalid domain (SNI) on service 'core/data-graph': data-graph.azurewebsites.net
```

I believe the error is that this [line](https://github.com/jcmoraisjr/haproxy-ingress/blob/613f22728d798691fa2e83a7cc5e7145a116870a/pkg/converters/ingress/annotations/backend.go#L750): 

```golang
var validDomainRegex = regexp.MustCompile(`^[A-Za-z0-9.]*$`)
```

does not allow "-" in domain names.

**Environment information**

HAProxy Ingress version: `v0.12.2`

jcmoraisjr:
Fixed in v0.12.4! Closing.

PR #791, opened 2021-06-01
Fix domain validation on secure backend keys

jcmoraisjr:
The domain validation on `secure-sni` and `secure-verify-hostname` config keys were missing a dash `-`, incorrectly issuing a valid `some-domain.tld` domain as invalid. This update also check if a dot is used and the tld is 2 <= size <= 6.


jcmoraisjr:
#790

Issue #783, opened 2021-05-12
tls


xlovepython:
tls not configured ,but ssh-redirect is effect

**Expected behavior**

**Steps to reproduce the problem**

1.  problem:
[root@node1 ~]# curl 127.0.0.1 -H "host: xxx.com" -I
HTTP/1.1 302 Found
content-length: 0
location: https://xxx.com/
cache-control: no-cache
    
2.  then i edit configmap to add ssl-redirect: "false"   then it's good

**Environment information**

HAProxy Ingress version: `v0.12.3`

Command-line options:

```
[root@node1 ~]# curl 127.0.0.1 -H "host: xxx.com" -I
HTTP/1.1 302 Found
content-length: 0
location: https://xxx.com/
cache-control: no-cache
```

Global options:

```
config-defaults: |
    compression algo gzip
    compression type text/xml text/plain text/css application/javascript application/x-javascript application/rss+xml application/atom+xml application/xml application/json application/octet-stream
  forwardfor: ifmissing
  healthz-port: "10253"
  stats-port: "1936"
```

Ingress objects:

```
metadata:
  annotations:
    kubernetes.io/ingress.class: haproxy
  name: xxx
  namespace: default
spec:
  rules:
  - host: xxx.com
    http:
      paths:
      - backend:
          service:
            name: nginx
            port:
              number: 80
        path: /
        pathType: ImplementationSpecific
```

jcmoraisjr:
Hi, thanks for reporting and sorry about the long delay. I can confirm this behavior and it's a kind of issue hard to fix without breaking deployments out there. Planning to fix with a global configuration that can be used to add tls and default crt without the need to configure in the ingress resource.

jcmoraisjr:
Fixed in v0.13. Please add `ssl-always-add-https: "false"` in the global configmap to change/fix this behavior in v0.12.4 release. This config defaults to false in v0.13. Closing.

PR #793, opened 2021-06-04
Add ssl-always-add-https config key

jcmoraisjr:
Since at least v0.8, controller adds domains in both HTTP and HTTPS maps despite the use of the `tls` attribute. This behavior goes against the Ingress spec but cannot be simply fixed due to backward compatibility. This config key adds the ability to configure HTTPS without the need of the `tls` attribute and can be globally changed. Its default value is
`true` to preserve the behavior of current and older versions.

Should be merged to v0.12.

# Compatibility notes

v0.13 will merge this update changing the default value to `false`, which will impact deployments with Ingress resources that doesn't configure `tls` attribute. Such deployments by default will not configure HTTPS anymore. This configuration can be changed to `true` to bring back the v0.12 behavior, or `tls` attribute should be configured on every domain that needs HTTPS.


jcmoraisjr:
#783

PR #794, opened 2021-06-05
Add option to copy client method to auth-url

jcmoraisjr:
This update adds the ability to call the external authentication service using the same method provided by the user. Current option are only static and pre configured methods. The reading of the request method was delegated to the Lua script, because parameters to a Lua script cannot be interpolated.


jcmoraisjr:
#413

Issue #788, opened 2021-05-29
Server templates are using the wrong port name


mhyllander:
**Description of the problem**

When using a DNS resolver to look up pod IP addresses, the generated server template is using the wrong port name from the kubernetes service. It is using the `targetPort` name instead of the port `name`. Lookups fail if the `targetPort` is different from `name`.

**Expected behavior**

I am using an AKS cluster with coredns as DNS server. For a service, both A and SRV records are created. The SRV record has the form `_<port>._<protocol>.<service>.<namespace>.svc.cluster.local`. The port here is the `name` from the service endpoint:

```YAML
  ports:
  - name: http-admin
    port: 9990
    protocol: TCP
    targetPort: admin
```

The SRV record created in coredns for the service endpoint in this example is `_http-admin._tcp.hivejs.platform.svc.cluster.local`, using the `name` for the port.

But the server template generated by haproxy-ingress has `_admin._tcp.hivejs.platform.svc.cluster.local`, using the `targetPort` for the port, which obviously will fail because that SRV record does not exist.

**Environment information**

HAProxy Ingress version: `v0.12.2`

jcmoraisjr:
Hi thanks for the report! I'm addressing this using the following rules, let me know if something can be improved:

- if the target port is a number: `<svcname>.<namespace>.svc.cluster.local:<port>`
- if the target port is not a number and there is a port name: `_<portname>._tcp.<svcname>.<namespace>.svc.cluster.local`
- if the target port is not a number and there is not a port name: `_<targetport>._tcp.<svcname>.<namespace>.svc.cluster.local`

mhyllander:
Hi! I'm not sure about the third alternative. I don't think SRV records are ever created using the target port. At least that's the way I read the k8s docs: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#srv-records. And I tried to remove the port name, which resulted in an error.

I'm thinking, since this is all about when service discovery is done using haproxy DNS lookups, don't you need to distinguish between normal and headless services, too? Something like:

* if the target port is a number:
  * if the service is headless: `<svcname>.<namespace>.svc.cluster.local:<targetport>`
  * if the service is normal: `<svcname>.<namespace>.svc.cluster.local:<port>`
* if the target port is not a number (special case to signify using SRV lookups): `_<portname>._tcp.<svcname>.<namespace>.svc.cluster.local`

jcmoraisjr:
[...]

PR #796, opened 2021-06-07
Use the port name on DNS resolver template

jcmoraisjr:
A haproxy backend naming uses the service namespace and name, and also the port of the service. Kubernetes has a number of ways to reference a port, which can be a number or a name, and both can be of the service itself or the target (a pod). Current implementation just use the target reference, and being a number make it in sync with the port number in the target server. Anyway this is just a naming convention, except when DNS resolvers are in place.

The DNS resolver feature expects the name of the service port because this is the name the DNS knows. When port name and target port differs, the server-template is wrongly built with the target name. This commit adds a new `DNSPort` attribute that the DNS template should use to build the correct configuration.


jcmoraisjr:
#788

Issue #477, opened 2019-12-11
cors-allow-origin - use more than 1 URL


hAh0L13:
Hello all.

We use HAProxy-ingress 0.7.2 in our cluster (OKD, openshift)

Can I use two or more URLs in ingress annotation **ingress.kubernetes.io/cors-allow-origin** ?

If I adding two annotations, it use last applied. If I adding list of URLs delimited by comma, controller switch header to any - "*"

Docs didn't cover this case

jcmoraisjr:
Hi, from [this mdn doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#The_HTTP_response_headers):

> Access-Control-Allow-Origin specifies either a single origin, which tells browsers to allow that origin to access the resource; or else — for requests without credentials — the "*" wildcard, to tell browsers to allow any origin to access the resource.

So it sounds you can only use one origin or a wildcard `*` which the current implementation already supports.

hAh0L13:
Hi. Big thanks for this. Should have been more attentive.Thanks for link

maxisam:
FYI, nginx is going to support it. They are fixing a bug now

hAh0L13:
> FYI, nginx is going to support it. They are fixing a bug now

I would like a link to the confirmation of this

maxisam:
@hAh0L13  can't believe I forgot to paste the link https://github.com/kubernetes/ingress-nginx/issues/5496

jcmoraisjr:
How one would implement this? It's technically pretty simple but it's apparently unsupported?

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMultipleAllowOriginNotAllowed

It seems to me that I cannot add more than one header and this one should have one origin or a wildcard.

hAh0L13:
> It seems to me that I cannot add more than one header and this one should have one origin or a wildcard.

It looks like the solution is in a conditional block where CORS headers are configured based on a list of allowed domains.
https://github.com/kubernetes/ingress-nginx/issues/5496#issuecomment-662798662
ingress-nginx supports configurations snippets that allow this

jcmoraisjr:
Ah sure, a dynamic response based on the request. It seems promising, will have a look on this.

jcmoraisjr:
[...]

PR #797, opened 2021-06-08
Allow a list of origins in cors-allow-origin config

jcmoraisjr:
`cors-allow-origin` configuration key now allows a list of Origins. The `Access-Control-Allow-Origin` response header will be built based on the provided Origin, falling back to the first option if Origin cannot be found in the list.


jcmoraisjr:
#477

Issue #798, opened 2021-06-09
Stable server IDs for consistent-hash balancing


arodland:
**What are you trying to do**

I have a service that I'm migrating to k8s. It currently uses haproxy with `balance uri` and `hash-type consistent` to distribute traffic in a way that sends requests for related pieces of content to the same backend server, which increases cache performance by a lot. This mapping should remain stable across reloads and when servers are added and removed (only the portion of the hash ring corresponding to the new/removed server should change, the whole list shouldn't shuffle).

For this, it's not good enough just to sort the list and let haproxy assign sequential IDs, because an addition or removal in the middle of the list disturbs everything after it. So, our existing system assigns a unique ID to each server and passes that as the `id` in the `server` line, which haproxy uses as that server's hash seed.

I'd like to be able to do something similar with haproxy-ingress.

**What HAProxy Ingress should do or how it should behave differently**

Provide a way to set the `id` field of a `server` based on some stable property of the pod backing it (UUID, or name, or IP). HAProxy `id` is a 32-bit value, so a hash truncated to 32 bits would seem to be appropriate.

Since the `id` must be unique, in the case of a collision, I would suggest using a stable [open-addressing](https://en.wikipedia.org/wiki/Open_addressing) technique by taking the servers in lexicographic order, and if a slot is already taken, rehashing or probing forward until a free slot is found. For reasonable numbers of servers, collisions should happen extremely rarely (<0.1% chance of a collision for up to 2900 servers), so this technique is only a *tiny* bit less stable than one that uses persistent unique IDs.

jcmoraisjr:
Added in `v0.13.0-beta.2`. Closing.

PR #801, opened 2021-06-17
Stable IDs for consistent-hash load balancing

arodland:
Implementation for #798. Create a way to give servers stable `id` values in the HAProxy config to facilitate content-based balancing configs.


docs/content/en/docs/configuration/keys.md
```diff
@@ -910,6 +911,16 @@ HAProxy Ingress won't refuse to change the default naming if dynamic update is `
 
 ---
 
+## Backend server ID assignment
+
+| Configuration key          | Scope     | Default      | Since |
+|----------------------------|-----------|--------------|-------|
+| `assign-backend-server-id` | `Backend` | `false     ` | TODO  |
+
+When `true`, each backend server will receive an `id` in HAProxy config based on the Kubernetes UID of the pod backing it. When using a hash-based [`balance-algorithm`](#balance-algorithm) (for example `uri` or `source`) together with consistent hashing, this will maintain the stability of assignments when pods are added or removed — that is, a given URI component or source IP will mostly keep hashing to the same server. When this seetting is `false`, an addition or deletion in the server list may disturb the hash assignments of some or all of the remaining servers.
```
jcmoraisjr, line 920: please add also a few words about this config making dynamic update to not work - on a rolling update, the target IPs will change (which can be dynamically updated) but also the server id (which cannot).

pkg/converters/ingress/ingress.go
```diff
@@ -683,6 +686,18 @@ func (c *converter) syncChangedEndpointCookies() {
...
+	for _, backend := range c.haproxy.Backends().Items() {
+		c.syncBackendEndpointHashes(backend)
+	}
+}
+
+func (c *converter) syncChangedEndpointHashes() {
+	for _, backend := range c.haproxy.Backends().ItemsAdd() {
+		c.syncBackendEndpointHashes(backend)
+	}
+}
```
jcmoraisjr, line 699: what if these funcs were merged with `syncEndpointCookies()` and `syncChangedEndpointCookies()` (removing the `Cookie` suffix)?

pkg/converters/ingress/ingress.go
```diff
@@ -948,6 +963,50 @@ func (c *converter) syncBackendEndpointCookies(backend *hatypes.Backend) {
...
+			hash = hasher.Sum32() & 0x7fffffff
+		} else {
+			hash = 1
+			c.logger.Error("error calculating hash value for pod %s; ID assignment won't be stable: %v", ep.TargetRef, err)
+		}
+		for { // If the ID is already used, linearly probe to find one that's not
+			_, exists := usedPUIDS[hash]
+			if hash != 0 && !exists {
+				break
+			}
```
jcmoraisjr, line 1002: what's the meaning of `hash != 0`, maybe you meant `hash != 1` which means error calling `GetPod()`?

rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -789,6 +789,7 @@ backend {{ $backend.ID }}
         {{- "" }} weight {{ $ep.Weight }}
         {{- if and ($backend.CookieAffinity) ($ep.CookieValue) }} cookie {{ $ep.CookieValue }}{{ end }}
         {{- if $ep.SourceIP }} source {{ $ep.SourceIP }}{{ end }}
+        {{- if ne (int64 $ep.PUID) 0 }} id {{ $ep.PUID }}{{ end }}
```
jcmoraisjr, line 792: this can be simplified to `{{- if $ep.PUID }} id {{ $ep.PUID }}{{ end }}`

jcmoraisjr (commented):
Thanks for the contribution! Following a few suggestions and doubts.

jcmoraisjr:
Also about docs, please add also this config key here, along with `drain-support`: https://github.com/jcmoraisjr/haproxy-ingress/blob/master/docs/content/en/docs/configuration/command-line.md#--disable-pod-list

jcmoraisjr:
And finally about tests there are two of them needed by this functionality.

1. template, the easiest, just copy eg the code below to a new item of the test case and adjust to use the PUID

https://github.com/jcmoraisjr/haproxy-ingress/blob/128e09a9f2de5682361ce8527ef52deea6b02b7d/pkg/haproxy/instance_test.go#L656-L664

2. ingress converter isn't that funny but this one can be used as a baseline, it also creates pods, configures an annotation and check what happens. you'll probably want to add PUID in the endpointMock, and you'll probably want to add `omitempty`.

https://github.com/jcmoraisjr/haproxy-ingress/blob/128e09a9f2de5682361ce8527ef52deea6b02b7d/pkg/converters/ingress/ingress_test.go#L277-L317

https://github.com/jcmoraisjr/haproxy-ingress/blob/128e09a9f2de5682361ce8527ef52deea6b02b7d/pkg/converters/helper_test/marshal.go#L42-L47

docs/content/en/docs/configuration/keys.md
jcmoraisjr: ps

- remove the trailing spaces in the `false` quote;
- since v0.13 - this feature will be merged.

arodland: Okay. So I should recommend that anyone using this should disable `dynamic-scaling`, the same as with `backend-server-naming` above?

docs/content/en/docs/configuration/keys.md
```diff
@@ -910,6 +911,16 @@ HAProxy Ingress won't refuse to change the default naming if dynamic update is `
 
 ---
 
+## Backend server ID assignment
```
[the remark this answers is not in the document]
arodland: OK.

docs/content/en/docs/configuration/keys.md
```diff
@@ -288,210 +288,211 @@ TCP port number.
...
+| [`affinity`](#affinity)                                 | affinity type                           | Backend |                    |
+| [`agent-check-addr`](#agent-check)                      | address for agent checks                | Backend |                    |
+| [`agent-check-interval`](#agent-check)                  | time with suffix                        | Backend |                    |
+| [`agent-check-port`](#agent-check)                      | backend agent listen port               | Backend |                    |
+| [`agent-check-send`](#agent-check)                      | string to send upon agent connection    | Backend |                    |
+| [`allowlist-source-range`](#allowlist)                  | Comma-separated IPs or CIDRs            | Path    |                    |
+| [`app-root`](#app-root)                                 | /url                                    | Host    |                    |
+| [`assign-backend-server-id`](#assign-backend-server-id) | [true\|false]                           | Backend | `false`            |
```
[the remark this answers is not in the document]
arodland: OK.

pkg/converters/ingress/ingress.go
arodland: Sounds fine :)

pkg/converters/ingress/ingress.go
arodland: No, `0` is an invalid value for `id` in HAProxy (conveniently it also gives us a value to use for "don't set the ID" in the template). So if we have to linearly probe forward from 2,147,483,647, we should go to 1, not 0.

rootfs/etc/templates/haproxy/haproxy.tmpl
arodland: OK. I don't know this template language, so I was copying from a nearby example :)

docs/content/en/docs/configuration/keys.md
jcmoraisjr: Hmm I don't think so, only a note, a new phrase in an existent paragraph is enough, just to remind about this behavior. The backend naming is another matter due to the ability to wrongly name servers after a dynamic update.

jcmoraisjr: ps - with "not work" I mean a rolling update will make haproxy reload instead of dynamically update the endpoints.

arodland:
Oh okay, so no need for the user to do anything, just that dynamic updates
won't be possible.

docs/content/en/docs/configuration/keys.md
arodland: okay, so the user doesn't have to do anything, because the ingress recognizes that field as non-updatable... we just want to warn them. Correct?

jcmoraisjr: yep, a dynamic update only happens if the diff in the old and new state can be applied, a distinct PUID value is enough to go to the reload route.

docs/content/en/docs/configuration/keys.md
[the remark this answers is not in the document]
arodland: addressed.

docs/content/en/docs/configuration/keys.md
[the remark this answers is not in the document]
arodland: addressed.

docs/content/en/docs/configuration/keys.md
arodland: Okay, improved.

pkg/converters/ingress/ingress.go
arodland: Commented.

rootfs/etc/templates/haproxy/haproxy.tmpl
arodland: Done.

arodland:
Added the template test. Attempted the other test, but it's failing — I'm doing something wrong so PUID isn't being populated at all... and also I can't figure out how to do it by parsing some annotation, so I just set the field directly (which still doesn't work). Please help.

jcmoraisjr:
First of all thanks for the time spent on this! It seems pretty good.

Currently `backend.Server.AssignID` is only used to let the `syncBackendEndpointHashes()` know that an assignment is desired, the real state used by the template is the `endpoint.PUID`. You can remove the `AssignID` attribute and read the annotation state like this:

```diff
-	if !backend.Server.AssignID {
+	mapper := c.backendAnnotations[backend]
+	if mapper != nil && !mapper.Get(ingtypes.BackAssignBackendServerID).Bool() {
```

Regarding the tests, the ingress part uses a non trivial path and have some challenges to be properly tested. Below is a diff of the changes to fix it:

```diff
diff --git a/pkg/converters/helper_test/marshal.go b/pkg/converters/helper_test/marshal.go
index 0133e530..886a085a 100644
--- a/pkg/converters/helper_test/marshal.go
+++ b/pkg/converters/helper_test/marshal.go
@@ -88,7 +88,7 @@ func marshalBackends(weight bool, habackends ...*hatypes.Backend) []backendMock
 	for _, b := range habackends {
 		endpoints := []endpointMock{}
 		for _, e := range b.Endpoints {
-			endpoint := endpointMock{IP: e.IP, Port: e.Port, Drain: e.Weight == 0}
+			endpoint := endpointMock{IP: e.IP, Port: e.Port, Drain: e.Weight == 0, PUID: e.PUID}
 			if weight {
 				endpoint.Weight = e.Weight
 			}
diff --git a/pkg/converters/ingress/ingress_test.go b/pkg/converters/ingress/ingress_test.go
index 7f8038bb..95776361 100644
--- a/pkg/converters/ingress/ingress_test.go
+++ b/pkg/converters/ingress/ingress_test.go
@@ -323,7 +323,11 @@ func TestSyncServerIDs(t *testing.T) {
 	c := setup(t)
 	defer c.teardown()

-	_, ep := c.createSvc1("default/echo", "8080", "172.17.1.101,172.17.1.102,172.17.1.103")
+	svc, ep := c.createSvc1("default/echo", "8080", "172.17.1.101,172.17.1.102,172.17.1.103")
+	ep.Subsets[0].Addresses[0].TargetRef.Name = "echo-1"
+	ep.Subsets[0].Addresses[1].TargetRef.Name = "echo-2"
+	ep.Subsets[0].Addresses[2].TargetRef.Name = "echo-3"
+
 	c.cache.PodList = make(map[string]*api.Pod)
 	c.cache.PodList["default/echo-1"] = c.createPod1("default/echo-1", "172.17.1.101", "http:8080")
 	c.cache.PodList["default/echo-2"] = c.createPod1("default/echo-2", "172.17.1.102", "http:8080")
@@ -335,15 +339,11 @@ func TestSyncServerIDs(t *testing.T) {
 	c.cache.PodList["default/echo-2"].UID = "costarring"
 	c.cache.PodList["default/echo-3"].UID = "liquid"

-	c.cache.Changed.EndpointsNew = []*api.Endpoints{ep}
+	svc.Annotations = map[string]string{"ingress.kubernetes.io/" + ingtypes.BackAssignBackendServerID: "true"}
 	c.Sync(
 		c.createIng1("default/echo", "echo.example.com", "/", "echo:8080"),
 	)

-	// TODO this doesn't work.
-	be := c.hconfig.Backends().FindBackend("default", "echo", "8080")
-	be.Server.AssignID = true
-
 	c.compareConfigBack(`
 - id: default_echo_8080
   endpoints:
@@ -2596,6 +2596,7 @@ func (u *updaterMock) UpdateHostConfig(host *hatypes.Host, mapper *annotations.M
 func (u *updaterMock) UpdateBackendConfig(backend *hatypes.Backend, mapper *annotations.Mapper) {
 	backend.Server.MaxConn = mapper.Get(ingtypes.BackMaxconnServer).Int()
 	backend.BalanceAlgorithm = mapper.Get(ingtypes.BackBalanceAlgorithm).Value
+	backend.Server.AssignID = mapper.Get(ingtypes.BackAssignBackendServerID).Bool()
 	for _, path := range backend.Paths {
 		config := mapper.GetConfig(path.Link)
 		path.MaxBodySize = config.Get(ingtypes.BackProxyBodySize).Int64()
```

arodland:
Tests pass!

pkg/converters/ingress/ingress.go
```diff
@@ -952,7 +952,8 @@ func (c *converter) syncBackendEndpointCookies(backend *hatypes.Backend) {
 }
 
 func (c *converter) syncBackendEndpointHashes(backend *hatypes.Backend) {
-	if !backend.Server.AssignID {
+	mapper := c.backendAnnotations[backend]
+	if mapper != nil && !mapper.Get(ingtypes.BackAssignBackendServerID).Bool() {
```
arodland, line 956: Actually shouldn't this be `if mapper == nil || !mapper.Get(ingtypes.BackAssignBackendServerID).Bool()` ?

arodland: @jcmoraisjr

jcmoraisjr: ouch, yes, makes sense. in fact this `mapper != nil` is a "just in case", it should never return nil for a known backend, but your proposal is in fact the correct way to implement this.

arodland: All set.

jcmoraisjr:
Thank you Andrew! It seems good, merging now.

Issue #803, opened 2021-06-22
Request for better feedback around --configmap= option


toothbrush:
Hi! Thanks for maintaining haproxy-ingress! 👍 I ran into a small annoyance while setting up a new instance, so i thought i'd share.

**Description of the problem**

Setting up a new instance of haproxy-ingress, things didn't work as expected; it turned out it wasn't reading my configmap.  It was difficult to track down the problem – it turned out i was setting `--configmap=wrongnamespace/nonexistent-myconfigmap` but `haproxy-ingress` accepted it silently.

Additionally, the `--configmap=...` option isn't mentioned on https://haproxy-ingress.github.io/docs/configuration/command-line/, adding to my confusion.  It is mentioned here though: 

https://github.com/jcmoraisjr/haproxy-ingress/blob/42034000bc7e4516696fc7f107b6a2d9b82588a3/examples/custom-configuration/README.md#L29-31

**Expected behavior**

At least i would expect a log line output, or better yet a fatal crash if the user passes `--configmap=...` pointing at a ConfigMap that cannot be read.  This would've pointed me straight to the issue.

Additionally, it could be nice to output to the console something like "Successfully loaded configuration from ConfigMap your/configmap." so that you know that's working.

**Steps to reproduce the problem**

1. Call haproxy-ingress with `--configmap=nonexistent`.

**Environment information**

HAProxy Ingress version: `v0.12.3`

jcmoraisjr:
Hi thanks for reporting this. We really need to spend a few more time with the command-line options doc.

Regarding the configmap reading, this is in fact a historical and annoying misbehavior. I'll add an error log in the maintenance branches so we don't break deployments that uses this as a feature, and will crash the application on new versions.

jcmoraisjr:
`v0.10.9`, `v0.11.8` and `v0.12.6` will now log if the global `--configmap` cannot be found, and `v0.13.0-beta.2` and newer will crash. Closing.

PR #804, opened 2021-06-23
Ensure that configured global ConfigMap exists

jcmoraisjr:
Launch is actually copying the global configmap name from the command-line straight to the lister without checking its existence. The lister will watch configmap events and will only wake up if the name matches with the one provided in the command-line, so currently a mistyped name is silently ignored.

From now the configmap name is checked in the bootstrap process and will crash if not found or if the controller doesn't have permission to read it.

This should be merged as far as v0.10, but v0.12 and older should only log error instead, due to preserve backward compatibility with the current behavior.

### DIFF
--- a/pkg/common/ingress/controller/launch.go
+++ b/pkg/common/ingress/controller/launch.go
@@ -252,6 +252,19 @@ func NewIngressController(backend ingress.Controller) *GenericController {
 
 	ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
#803

toothbrush:
Thank you for the quick turnaround on this one! 🙌

Issue #805, opened 2021-06-23
Missing command-line options doc


jcmoraisjr:
**Description of the problem**

Some of the available command-line options aren't documented

**Expected behavior**

Document the following command-line options:

* [x] apiserver-host
* [x] configmap
* [x] disable-node-list
* [x] election-id
* [x] force-namespace-isolation
* [x] health-check-path
* [x] report-node-internal-ip-address
* [x] sync-period
* [x] update-status-on-shutdown
* [x] v
* [x] validate-config
* [x] version

jcmoraisjr:
#803

jcmoraisjr:
Added in `v0.12.6` doc. Closing.

PR #806, opened 2021-06-23
docs: Add all command-line options to list.

toothbrush:
Hey, i don't know if this is correct or helpful, but i thought i'd try to go through the code and add all the command-line flags i can find.

I wasn't sure about the description on some of them, feel free to amend or tell me to do it better 😊

### DIFF
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -30,6 +30,7 @@ The following command-line options are supported:
 | [`--disable-api-warnings`](#disable-api-warnings)       | [true\|false]              | `false`                 | v0.12 |
 | [`--disable-pod-list`](#disable-pod-list)               | [true\|false]              | `false`                 | v0.11 |
 | [`--election-id`](#election-id)                         | identifier                 | `ingress-controller-leader` |   |
+| [`--force-namespace-isolation`](#force-namespace-isolation) | [true\|false]          | `false`                 |       |
 | [`--healthz-port`](#stats)                              | port number                | `10254`                 |       |
 | [`--ingress-class`](#ingress-class)                     | name                       | `haproxy`               |       |
 | [`--kubeconfig`](#kubeconfig)                           | /path/to/kubeconfig        | in cluster config       |       |
@@ -181,6 +182,15 @@ Disables in memory pod list and also pod watch for changes. Pod list and watch i
 ## --election-id
 
 The ID to be used for electing ingress controller leader.  Defaults to `ingress-controller-leader`.
+
+---
+
+## --force-namespace-isolation
+
+Whether to force namespace isolation.  This flag is required to avoid the reference of secrets,
+configmaps or the default backend service located in a different namespace than specified in the
+flag `--watch-namespace` (which defaults to all namespaces, so you will probably want to set that
+flag, too).
 
 ---
 
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -51,6 +51,7 @@ The following command-line options are supported:
 | [`--update-status-on-shutdown`](#update-status-on-shutdown) | [true\|false]          | `true`                  |       |
 | [`--validate-config`](#validate-config)                 | [true\|false]              | `false`                 |       |
 | [`--verify-hostname`](#verify-hostname)                 | [true\|false]              | `true`                  |       |
+| [`--version`](#version)                                 | [true\|false]              | `false`                 |       |
 | [`--wait-before-shutdown`](#wait-before-shutdown)       | seconds as integer         | `0`                     | v0.8  |
 | [`--wait-before-update`](#wait-before-update)           | duration                   | `200ms`                 | v0.11 |
 | [`--watch-gateway`](#watch-gateway)                     | [true\|false]              | `false`                 | v0.13 |
@@ -474,6 +475,12 @@ match the hostname are discarded and a warning is logged into the ingress contro
 
 Use `--verify-hostname=false` argument to bypass this validation. If used, HAProxy will provide
 the certificate declared in the `secretName` ignoring if the certificate is or is not valid.
+
+---
+
+## --version
+
+Show release information about the ingress controller.
 
 ---
 
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -45,6 +45,7 @@ The following command-line options are supported:
 | [`--sort-backends`](#sort-backends)                     | [true\|false]              | `false`                 |       |
 | [`--sort-endpoints-by`](#sort-endpoints-by)             | [endpoint\|ip\|name\|random] | `endpoint`            | v0.11 |
 | [`--stats-collect-processing-period`](#stats)           | time                       | `500ms`                 | v0.10 |
+| [`--sync-period`](#sync-period)                         | time                       | `10m`                   |       |
 | [`--tcp-services-configmap`](#tcp-services-configmap)   | namespace/configmapname    | no tcp svc              |       |
 | [`--verify-hostname`](#verify-hostname)                 | [true\|false]              | `true`                  |       |
 | [`--wait-before-shutdown`](#wait-before-shutdown)       | seconds as integer         | `0`                     | v0.8  |
@@ -370,6 +371,13 @@ Options:
 * `--healthz-port`: Defines the port number haproxy-ingress should listen to. Defaults to `10254`.
 * `--profiling`: Configures if the profiling URI should be enabled. Defaults to `true`.
 * `--stats-collect-processing-period`: Defines the interval between two consecutive readings of haproxy's `Idle_pct`, used to generate `haproxy_processing_seconds_total` metric. haproxy updates Idle_pct every `500ms`, which makes that the best configuration value, and it's also the default if not configured. Values higher than `500ms` will produce a less accurate collect. Change to 0 (zero) to disable this metric.
+
+---
+
+## --sync-period
+
+Configures the default resync period of the Kubernetes client's informer factory. Defaults to 10
+minutes.
 
 ---
 
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -20,6 +20,7 @@ The following command-line options are supported:
 | [`--acme-track-tls-annotation`](#acme)                  | [true\|false]              | `false`                 | v0.9  |
 | [`--allow-cross-namespace`](#allow-cross-namespace)     | [true\|false]              | `false`                 |       |
 | [`--annotations-prefix`](#annotations-prefix)           | prefix list without `/`    | `haproxy-ingress.github.io,ingress.kubernetes.io` | v0.8  |
+| [`--apiserver-host`](#apiserver-host)                   | address of K8s API server  |                         |       |
 | [`--backend-shards`](#backend-shards)                   | int                        | `0`                     | v0.11 |
 | [`--buckets-response-time`](#buckets-response-time)     | float64 slice           | `.0005,.001,.002,.005,.01` | v0.10 |
 | [`--controller-class`](#ingress-class)                  | suffix                     | `""`                    | v0.12 |
@@ -99,6 +100,15 @@ command-line option is used.
 Annotations with other prefix or without any prefix are ignored. This allows to use HAProxy Ingress
 with other ingress controllers that shares ingress and service resources without conflicting each
 other.
+
+---
+
+## --apiserver-host
+
+Allows you to specify an explicit host for the Kubernetes API server, in the format of
+`protocol://address:port`, e.g., `http://localhost:8080`.  If this value isn't specified, the
+assumption is that the binary is running inside a Kubernetes cluster and local discovery will be
+attempted.
 
 ---
 
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -145,6 +145,8 @@ with `file://` can be used, containing both certificate and private key in PEM f
 
 A self-signed fake certificate is used if not declared, the secret or the file is not found.
 
+---
+
 ## --disable-api-warnings
 
 Since v0.12.4
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -22,7 +22,7 @@ The following command-line options are supported:
 | [`--annotations-prefix`](#annotations-prefix)           | prefix list without `/`    | `haproxy-ingress.github.io,ingress.kubernetes.io` | v0.8  |
 | [`--backend-shards`](#backend-shards)                   | int                        | `0`                     | v0.11 |
 | [`--buckets-response-time`](#buckets-response-time)     | float64 slice           | `.0005,.001,.002,.005,.01` | v0.10 |
-| [`--controller-class`](#ingress-class)                  | suffix                     | ``                      | v0.12 |
+| [`--controller-class`](#ingress-class)                  | suffix                     | `""`                    | v0.12 |
 | [`--default-backend-service`](#default-backend-service) | namespace/servicename      | haproxy's 404 page      |       |
 | [`--default-ssl-certificate`](#default-ssl-certificate) | namespace/secretname       | fake, auto generated    |       |
 | [`--disable-api-warnings`](#disable-api-warnings)       | [true\|false]              | `false`                 | v0.12 |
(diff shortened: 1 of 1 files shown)


toothbrush:
This should address #805, at least somewhat.

docs/content/en/docs/configuration/command-line.md
```diff
@@ -376,6 +448,27 @@ See also:
...
+Indicates whether the ingress controller should update the Ingress status IP/hostname when the
+controller is being stopped. Defaults to `true`.
+
+---
+
+## --validate-config
+
+Determines whether the resulting configuration files should be validated when a dynamic update was
+applied. Default value is `false`, which means the validation will only happen when HAProxy needs to
+be reloaded.
```
toothbrush, line 468: Should i describe the behaviour if a dynamic update causes invalid configuration?  Will it just ignore the update, or crash?

docs/content/en/docs/configuration/command-line.md
```diff
@@ -376,6 +448,27 @@ See also:
 
 ---
 
+## --update-status
+
+Indicates whether the ingress controller should update the Ingress status IP/hostname. Defaults to `true`.
```
toothbrush, line 453: I copied this description from the CLI, but i'm unclear what it means exactly.  As a result this description could probably be improved.

docs/content/en/docs/configuration/command-line.md
```diff
@@ -315,12 +380,19 @@ Configures an endpoint with statistics, debugging and health checks. The followi
...
+* `--health-check-path`: Defines the URL to be used as a health check for the default server.  Defaults to `/healthz`.
 * `--healthz-port`: Defines the port number haproxy-ingress should listen to. Defaults to `10254`.
 * `--profiling`: Configures if the profiling URI should be enabled. Defaults to `true`.
 * `--stats-collect-processing-period`: Defines the interval between two consecutive readings of haproxy's `Idle_pct`, used to generate `haproxy_processing_seconds_total` metric. haproxy updates Idle_pct every `500ms`, which makes that the best configuration value, and it's also the default if not configured. Values higher than `500ms` will produce a less accurate collect. Change to 0 (zero) to disable this metric.
 
 ---
 
+## --sync-period
+
+Specifies how often to relist and confirm cloud resources. Defaults to 10 minutes.
```
toothbrush, line 392: This description is also a bit mysterious to me 😅

docs/content/en/docs/configuration/command-line.md
```diff
@@ -152,6 +195,21 @@ Disables in memory pod list and also pod watch for changes. Pod list and watch i
 
 ---
 
+## --election-id
+
+The ID to be used for electing ingress controller leader.  Defaults to `ingress-controller-leader`.
```
toothbrush, line 200: Should we mention where this ID is used / stored?

docs/content/en/docs/configuration/command-line.md
```diff
@@ -152,6 +195,21 @@ Disables in memory pod list and also pod watch for changes. Pod list and watch i
...
+
+---
+
+## --force-namespace-isolation
+
+Whether to force namespace isolation.  This flag is required to avoid the reference of secrets,
+configmaps or the default backend service located in a different namespace than specified in the
+flag `--watch-namespace` (which defaults to all namespaces, so you will probably want to set that
+flag, too).
+
```
toothbrush, line 210: Perhaps mention the need to set `--disable-node-list`, too.

docs/content/en/docs/configuration/command-line.md
```diff
@@ -144,6 +180,13 @@ deprecation. The default behavior is to log all API server warnings.
 
 ---
 
+## --disable-node-list
+
+Disable querying nodes. If `--force-namespace-isolation` is true, this should also be set.  Defaults
```
toothbrush, line 185: > If `--force-namespace-isolation` is true, this should also be set.

Why is that?

toothbrush:
@jcmoraisjr, i added a bunch of questions – if you want you can informally reply and i'd be happy to elaborate and fix up the descriptions in the documentation.  If this is useful of course. 😄

jcmoraisjr:
Hi Paul, thanks for the time invested on this! The launch functionality is one of the last features we inherited from the old times of k8s/ingress as a dependency (all the current pkg/common/* source files) and wasn't ported to the new controller yet. Launch is responsible for parse and show the command-line options, and some of them we simply don't use. Please give me a few more days, I'll deprecate what we don't use, revise all the inline doc using your questions as a guideline, and will notify here as soon as it's done and PR'ed.

toothbrush:
No problem, give me a ping when i can update this doc.  Thanks again for taking the time to address all of this! 🥬 🍌

jcmoraisjr:
I've just added a PR that tries to fix some alignment issues and also some bad descriptions to the help output of command-line options. I hope most of your questions were answered, I'll comment the ones that maybe wasn't and leave here the current output for convenience. Don't hesitate to ping here if you need anything.

```
$ go run pkg/main.go -h
Usage of :
      --acme-check-period duration                 Time between checks of invalid or expiring certificates (default 24h0m0s)
      --acme-election-id string                    Prefix of the election ID used to choose the acme leader (default "acme-leader")
      --acme-fail-initial-duration duration        The initial time to wait to retry sign a new certificate after a failure. The
                                                   time between retries will grow exponentially until 'acme-fail-max-duration' (default 5m0s)
      --acme-fail-max-duration duration            The maximum time to wait after failing to sign a new certificate (default 8h0m0s)
      --acme-secret-key-name string                Name and an optional namespace of the secret which will store the acme account
                                                   private key. If a namespace is not provided, the secret will be created in the
                                                   same namespace of the controller pod (default "acme-private-key")
      --acme-server                                Enables acme server. This server is used to receive and answer challenges from
                                                   Lets Encrypt or other acme implementations.
      --acme-token-configmap-name string           Name and an optional namespace of the configmap which will store acme tokens
                                                   used to answer the acme challenges. If a namespace is not provided, the secret
                                                   will be created in the same namespace of the controller pod (default "acme-validation-tokens")
      --acme-track-tls-annotation                  Enable tracking of ingress objects annotated with 'kubernetes.io/tls-acme'
      --allow-cross-namespace                      Defines if the ingress controller can reference resources of another
                                                   namespaces. Cannot be used if force-namespace-isolation is true
      --alsologtostderr                            log to standard error as well as files
      --annotations-prefix string                  Defines a comma-separated list of annotation prefix for ingress and service (default "haproxy-ingress.github.io,ingress.kubernetes.io")
      --apiserver-host string                      The address of the Kubernetes Apiserver to connect to in the format of
                                                   protocol://address:port, e.g., http://localhost:8080. If not specified, the
                                                   assumption is that the binary runs inside a Kubernetes cluster and local
                                                   discovery is attempted.
      --backend-shards int                         Defines how much files should be used to configure the haproxy backends
      --buckets-response-time float64Slice         Configures the buckets of the histogram used to compute the response time of
                                                   the haproxy's admin socket. The response time unit is in seconds. (default [0.000500,0.001000,0.002000,0.005000,0.010000])
      --configmap string                           Name of the ConfigMap that contains the custom configuration to use
      --controller-class string                    Defines an alternative controller name this controller should listen to. If
                                                   empty, this controller will listen to ingress resources whose controller's
                                                   IngressClass is 'haproxy-ingress.github.io/controller'. Non empty values add a
                                                   new /path, eg controller-class=staging will make this controller look for
                                                   'haproxy-ingress.github.io/controller/staging'
      --default-backend-service string             Service used to serve a 404 page for the default backend. Takes the form
                                                   namespace/name. The controller uses the first node port of this Service for the
                                                   default backend.
      --default-ssl-certificate string             Name of the secret that contains a SSL certificate to be used as
                                                   default for a HTTPS catch-all server
      --disable-api-warnings                       Declare to disable warnings from the API server.
      --disable-node-list                          DEPRECATED: This flag used to disable node listing due to missing permissions.
                                                   Actually node listing isn't needed and it is always disabled
      --disable-pod-list                           Defines if HAProxy Ingress should disable pod watch and in memory list. Pod
                                                   list is mandatory for drain-support (should not be disabled) and optional for
                                                   blue/green.
      --election-id string                         Election id to be used for status update. (default "ingress-controller-leader")
      --force-namespace-isolation                  Force namespace isolation. This flag is required to avoid the reference of
                                                   secrets, configmaps or the default backend service located in a different
                                                   namespace than the specified in the flag --watch-namespace.
      --health-check-path string                   Defines the URL to be used as health check inside in the default server. (default "/healthz")
      --healthz-port int                           port for healthz endpoint. (default 10254)
      --ignore-ingress-without-class               DEPRECATED, this option is ignored. Use --watch-ingress-without-class
                                                   command-line option instead to define if ingress without class should be
                                                   tracked.
      --ingress-class string                       Name of the IngressClass to route through this controller. (default "haproxy")
      --kubeconfig string                          Path to kubeconfig file with authorization and master location information.
      --log_backtrace_at traceLocation             when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                             If non-empty, write log files in this directory
      --logtostderr                                log to standard error instead of files
      --master-socket string                       Defines the master CLI unix socket of an external HAProxy running in
                                                   master-worker mode. Defaults to use the embedded HAProxy if not declared.
      --max-old-config-files int                   Maximum old haproxy timestamped config files to allow before being cleaned up.
                                                   A value <= 0 indicates a single non-timestamped config file will be used
      --profiling                                  Enable profiling via web interface host:port/debug/pprof/ (default true)
      --publish-service string                     Service fronting the ingress controllers. Takes the form namespace/name. The
                                                   controller will set the endpoint records on the ingress objects to reflect
                                                   those on the service.
      --rate-limit-update float32                  Maximum of updates per second this controller should perform. Default is 0.5,
                                                   which means wait 2 seconds between Ingress updates in order to add more changes
                                                   in a single reload (default 0.5)
      --reload-strategy string                     Name of the reload strategy. Options are: native or reusesocket (default "reusesocket")
      --report-node-internal-ip-address            Defines if the nodes IP address to be returned in the ingress status should be
                                                   the internal instead of the external IP address
      --sort-backends                              Defines if backend's endpoints should be sorted by name. This option has less
                                                   precedence than --sort-endpoints-by if both are declared.
      --sort-endpoints-by string                   Defines how to sort backend's endpoints. Allowed values are: 'endpoint' - same
                                                   k8s endpoint order (default); 'name' - server/endpoint name;
                                                   'ip' - server/endpoint IP and port; 'random' - shuffle endpoints on every
                                                   haproxy reload
      --stats-collect-processing-period duration   Defines the interval between two consecutive readings of haproxy's Idle_pct.
                                                   haproxy updates Idle_pct every 500ms, which makes that the best configuration
                                                   value. Change to 0 (zero) to disable this metric. (default 500ms)
      --stderrthreshold severity                   logs at or above this threshold go to stderr (default 2)
      --sync-period duration                       Configures the default resync period of Kubernetes' informer factory. (default 10m0s)
      --tcp-services-configmap string              Name of the ConfigMap that contains the definition of the TCP services to
                                                   expose. The key in the map indicates the external port to be used. The value is
                                                   the name of the service with the format namespace/serviceName and the port of
                                                   the service could be a number of the name of the port. The ports 80 and 443 are
                                                   not allowed as external ports. This ports are reserved for the backend
      --update-status                              Indicates if the controller should update the 'status' attribute of all the
                                                   Ingress resources that this controller is tracking. (default true)
      --update-status-on-shutdown                  Indicates if the ingress controller should update the Ingress status
                                                   IP/hostname when the controller is being stopped. (default true)
  -v, --v Level                                    log level for V logs
      --validate-config                            Define if the resulting configuration files should be validated when a dynamic
                                                   update was applied. Default value is false, which means the validation will
                                                   only happen when HAProxy needs to be reloaded. If validation fails, HAProxy
                                                   Ingress will log the error and set the metric 'haproxyingress_update_success'
                                                   as failed (zero)
      --verify-hostname                            Defines if the controller should verify if the provided certificate is valid,
                                                   ie, it's SAN extension has the hostname. (default true)
      --version                                    Shows release information about the Ingress controller
      --vmodule moduleSpec                         comma-separated list of pattern=N settings for file-filtered logging
      --wait-before-shutdown int                   Define time controller waits until it shuts down when SIGTERM signal was
                                                   received
      --wait-before-update duration                Amount of time to wait before start a reconciliation and update haproxy, giving
                                                   the time to receive all/most of the changes of a batch update. (default 200ms)
      --watch-gateway                              Watch and parse resources from the Gateway API
      --watch-ingress-without-class                Defines if this controller should also listen to ingress resources that doesn't
                                                   declare neither the kubernetes.io/ingress.class annotation nor the
                                                   <ingress>.spec.ingressClassName field. Defaults to false
      --watch-namespace string                     Namespace to watch for Ingress. Default is to watch all namespaces
pflag: help requested
exit status 2
```

docs/content/en/docs/configuration/command-line.md
jcmoraisjr: Added a few more words in the new output

docs/content/en/docs/configuration/command-line.md
jcmoraisjr: Yeah same here. This value is passed straight to the k8s client's informer factory and it's in fact a bit of mysterious to me as well what it's really doing. Need to dig a bit more in the code. Anyway I changed the description to something a bit more concrete even though it still doesn't explain the functionality.

docs/content/en/docs/configuration/command-line.md
jcmoraisjr: This will in the end create a configmap that stores an internal state used to know who is the current leader and if it's alive. The id is used to name this configmap, and the leader is used to update ingress states. I'd leave it as it is (which is different from the previous version of the -h output?) because we already have a new refactor which will reuse this leader elector to the acme functionality, so we'll have a new chance to improve this description.

docs/content/en/docs/configuration/command-line.md
jcmoraisjr: No need to add it, just deprecated the `--disable-node-list` option.

docs/content/en/docs/configuration/command-line.md
jcmoraisjr: You can remove the `--disable-node-list` altogether. Planning a new command-line parsing refactor during v0.14 development which should simplify all this cluster/namespaced scoped configurations, so the description will be improved as well.

toothbrush:
Wow, that looks excellent, thanks – i'll have a look at this later and update the PR.  Thank you!

docs/content/en/docs/configuration/command-line.md
toothbrush: Updated based on new CLI help texts.

docs/content/en/docs/configuration/command-line.md
toothbrush: > I'd leave it as it is (which is different from the previous version of the -h output?)

Yeah, in a few cases i've tried to editorialise a little bit, but if you prefer, i'm happy to put the CLI help text verbatim here.

toothbrush:
Alright, i've made some small adjustments based on your answers.  Let me know if there's anything else i need to do! 👍

docs/content/en/docs/configuration/command-line.md
```diff
@@ -36,11 +42,17 @@ The following command-line options are supported:
...
+| [`--report-node-internal-ip-address`](#report-node-internal-ip-address) | [true\|false] | `false`              |       |
 | [`--sort-backends`](#sort-backends)                     | [true\|false]              | `false`                 |       |
 | [`--sort-endpoints-by`](#sort-endpoints-by)             | [endpoint\|ip\|name\|random] | `endpoint`            | v0.11 |
 | [`--stats-collect-processing-period`](#stats)           | time                       | `500ms`                 | v0.10 |
+| [`--sync-period`](#sync-period)                         | time                       | `600s`                  |       |
 | [`--tcp-services-configmap`](#tcp-services-configmap)   | namespace/configmapname    | no tcp svc              |       |
+| [`--update-status`](#update-status)                     | [true\|false]              | `true`                  |       |
+| [`--update-status-on-shutdown`](#update-status-on-shutdown) | [true\|false]          | `true`                  |       |
+| [`--validate-config`](#validate-config)                 | [true\|false]              | `false`                 |       |
```
jcmoraisjr, line 53: Add `--v` option with something like this in the description:

Configures the log verbosity. `1` is the default value and adds only errors, warnings and a few update events. `2` is a good balance between low verbosity and rich details about controller's events.

docs/content/en/docs/configuration/command-line.md
```diff
@@ -36,11 +42,17 @@ The following command-line options are supported:
...
 | [`--verify-hostname`](#verify-hostname)                 | [true\|false]              | `true`                  |       |
+| [`--version`](#version)                                 | [true\|false]              | `false`                 |       |
```
jcmoraisjr, line 55: `--version` was missing?

jcmoraisjr (commented):
Just a few considerations and I'm done. Btw a pretty good job! Thanks for starting this review.

docs/content/en/docs/configuration/command-line.md
```diff
@@ -20,14 +20,20 @@ The following command-line options are supported:
...
 | [`--default-backend-service`](#default-backend-service) | namespace/servicename      | haproxy's 404 page      |       |
 | [`--default-ssl-certificate`](#default-ssl-certificate) | namespace/secretname       | fake, auto generated    |       |
 | [`--disable-api-warnings`](#disable-api-warnings)       | [true\|false]              | `false`                 | v0.12 |
 | [`--disable-pod-list`](#disable-pod-list)               | [true\|false]              | `false`                 | v0.11 |
+| [`--election-id`](#election-id)                         | identifier                 | `ingress-controller-leader` |   |
+| [`--force-namespace-isolation`](#force-namespace-isolation) | [true\|false]          | `false`                 |       |
+| [`--health-check-path`](#stats)                         | path                       | `/healthz`              |       |
 | [`--healthz-port`](#stats)                              | port number                | `10254`                 |       |
+| [`--ignore-ingress-without-class`](#ingress-class)      | [true\|false]              | `false`            | deprecated |
```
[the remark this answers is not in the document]
toothbrush: No problem, done.

docs/content/en/docs/configuration/command-line.md
toothbrush: Seems it was! I mean, it's not very important either.  But with these things i guess i naturally err on the side of preferring to mention everything rather than having "hidden" options :).

jcmoraisjr:
lgtm, thanks! merging.

Issue #807, opened 2021-06-23
HAProxy backend is marked as DOWN even though cfg has working Pods listed as servers


mkubaczyk:
**Description of the problem**

HAProxy is reporting specific backend to be DOWN even though two servers are listed in the configuration file for that backend. I confirmed these two servers were pointing at existing, working Pods of a service.

Having a chance while this happened for us, I quickly run stats on port 1936 and found out that:

<img width="1670" alt="Screenshot 2021-06-23 at 11 38 22" src="https://user-images.githubusercontent.com/12384536/123092451-c4c6a600-d42a-11eb-92a7-d368f8f73f27.png">

even though at the same time HAProxy had a configuration of:

```/etc/haproxy # cat haproxy5-backend324.cfg 
  # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# #
# #   HAProxy Ingress Controller
# #   --------------------------
# #   This file is automatically updated, do not edit
# #
#
backend connect_audit-report_6280
    mode http
    balance leastconn
    timeout queue 10s
    timeout server 300s
    http-request set-header X-Original-Forwarded-For %[hdr(x-forwarded-for)] if { hdr(x-forwarded-for) -m found }
    http-request del-header x-forwarded-for
    option forwardfor
    server srv007 10.147.65.36:6280 weight 1 check inter 2s
    server srv008 10.147.65.94:6280 weight 1 check inter 2s
    server srv003 127.0.0.1:1023 disabled weight 1 check inter 2s
    server srv004 127.0.0.1:1023 disabled weight 1 check inter 2s
    server srv006 127.0.0.1:1023 disabled weight 1 check inter 2s
    server srv005 127.0.0.1:1023 disabled weight 1 check inter 2s
    server srv002 127.0.0.1:1023 disabled weight 1 check inter 2s
    server srv001 127.0.0.1:1023 disabled weight 1 check inter 2s
```
This means srv007 and srv008 is pointing at working Pods but HAProxy does not see them according to stats output above.

It actually lasted for some time according to prometheus metric:

![Untitled](https://user-images.githubusercontent.com/12384536/123093437-fbe98700-d42b-11eb-945b-e2962f4b091b.png)

and only after manual forced restart of one Pod of our service it caused HAProxy to change its location in config to srv003 and hence treat it like UP:

[...]

PR #810, opened 2021-06-28
Reload haproxy if a backend server cannot be found

jcmoraisjr:
The controller maintain and evolve an in memory model that reflects the haproxy state. This model is used to rebuild the configuration files and also to name haproxy backends and servers when running a dynamic update. If for any reason the model and the haproxy instance loose their synchronization, backends or servers might not be found and a dynamic update would fail. This update parses the haproxy response, if it returns a `No such server` or any other non recognized message, haproxy will be fully reloaded in order to resynchronize with the model.


jcmoraisjr:
#807

PR #811, opened 2021-06-28
Add log of reload error on every reconciliation

jcmoraisjr:
Improves the visibility of a configuration issue. Currently only reload attempts would lead to a log error if the current state builds an invalid configuration file. This change adds an error log on every reconciliation, being dynamic or not, informing the timestamp when the configuration was broken for the first time, if the configuration is in fact broken. A broken configuration leads to new state not being correctly applied, and should also promote an outage if the controller or haproxy crashes, or the controller restarts.


jcmoraisjr:
#807

mkubaczyk (approved):

Issue #817, opened 2021-07-07
ignoring auth URL when host setted without dots


Dryymoon:
**Description of the problem**

We used custom auth proxy microservice.
Before and in release 0.13.0-snapshot.2 all work fine.
But after release 0.13.0-snapshot.3 released - Haproxy ignoring auth url and auth proxy dont work.

Partial config of target restricted service
```
...
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: haproxy
    ingress.kubernetes.io/auth-url: "http://ms-authentication/v1/haproxy-firewall"
...
```
Error:
```
backend.go:159] ignoring auth URL with an invalid domain on ingress 'default/some-restricted-service': lookup ms-authentication/v1: no such host
```

```ms-authentication``` - real name of custom auth service, service deployed inside kubernetes

**Expected behavior**

Auth proxy works without errors,
All requests to target service should be inspected by auth proxy

**Current behavior**
Haproxy Ingress controller hit error in console and all request to target service go without any authentification by auth service

**Environment information**

HAProxy Ingress version: 
0.13.0-snapshot.2 - works well
0.13.0-snapshot.3, 0.13.0-beta.1 - not work, got error

Ingress objects:

```
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: haproxy
    ingress.kubernetes.io/auth-url: "http://ms-authentication/v1/haproxy-firewall"
```

jcmoraisjr:
Hi thanks for this reproducer! This was a corner case where a http proto was used, you entered `<domain>/<path>` but `<namespace>/<name>` pattern was extracted. You can work around this in the current version adding a domain suffix, eg `ms-authentication.<your-namespace>`. We'll fix this in the next tag.

jcmoraisjr:
Fixed in `v0.13.0-beta.2`. Also, requests will always be denied if an attempt to configure external authentication fail.

PR #818, opened 2021-07-08
Fix auth-url parsing if hostname misses a dot

jcmoraisjr:
`auth-url` accepts http/s or service protocol, and the later accepts namespaced service names, if the configuration allows, since `v0.13-snapshot.3`. Namespaced names has a slash just like URIs, so parsing the whole URL is somewhat tricky. A domain name without dots was being matched as a namespace and concatenated with the first dir of the path, leading to a lookup error. This update fixes this behavior and also adds a few more unit tests covering more URL formats.

### DIFF
--- a/pkg/converters/ingress/utils/utils.go
+++ b/pkg/converters/ingress/utils/utils.go
@@ -22,7 +22,7 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
#817

PR #819, opened 2021-07-08
Always deny requests of failed auth configurations

jcmoraisjr:
The attempt to configure auth external or oauth means that the backend should be protected. However the current implementation is leaving the backend unprotected if a misconfiguration is found - which can be an external failure like a dns lookup. This update ensures that an attempt to protect a backend will in fact protect it, and misconfigurations or failures during the configuration parsing means an always-deny behavior.


jcmoraisjr:
#817

PR #825, opened 2021-07-25
Added section for AuditLog sidecar for ModSecurity daemonset

sealneaward:
In our customization of the ModSecurity agent, we have certain rules triggered to write to the auditlog and ignore being printed to the modsecurity-spoa container STDOUT. From my research, there doesn't appear to be a way to configure ModSecurity to also print the audit log file to STDOUT, but that might be better anyways because it would make it harder to read a file being edited with two different log sources.

We need to view what's going into the audit log file through container STDOUT. When we created the daemonset, we added a sidecar container to print that file to STDOUT for further ingestion.

Even though it is simple, it is extremely useful if someone was to seperate the destinations of logs for different rules and actions.


jcmoraisjr:
Lgtm, note however that this is the old examples page (yep I need to clean up things). You can find the sources used to build the doc site in `/docs/content/en/docs/examples/`. Note also that the modsecurity example has a manifest, you can find it in `/docs/static/resources/`, add the new one there, and the final path will be `https://haproxy-ingress.github.io/resources/filename.yaml`

sealneaward:
I updated from review, curious why the static resources aren't tracked. There isn't any line for it in the .gitignore file. @jcmoraisjr any reason for this?

jcmoraisjr:
> curious why the static resources aren't tracked

`/docs/.gitignore` seems to have broken config. Remove the `resources/` line just to track the new file, but without adding the `.gitignore` in the commit. I'll have a look on it.

sealneaward:
Updated with the resources file

jcmoraisjr:
Lgtm, thanks! Merging.

PR #826, opened 2021-07-29
Added docs on adding ClusterIP service for ModSecurity and removing NodeSelector

sealneaward:
Adding more to the ModSecurity docs to show the configuration of ModSecurity as a ClusterIP service wihout a NodeSelector.
This way you can have the daemonset schedule new pods whenever the cluster scales its nodes, without having to label the new nodes or adding additional endpoints for the new pods.


jcmoraisjr:
Your proposal have some good points which in some level makes the first deployment useless. What if the first deployment was changed to the proposed behavior instead of guide another change in the end of the doc? It could even be made as ordinary deployments instead. Note that you can create a service issuing `$ kubectl expose deployment modsecurity-spoa --port=12345` without the need of creating a new manifest - this makes easier to the user to change in order to fit his/her needs.

sealneaward:
I can write the docs to replace the [first deployment](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/docs/content/en/docs/examples/modsecurity.md#deploying-agent). Did you want to see docs on how to expose the `deployment` option and the `daemonset` option?

The `kubectl expose` command only works for deployments, where you will have to create a service with a manifest for daemonsets. Did you want to see new docs to include a deployment option with the `kubectl expose` svc and a daemonset option with the svc manifest?

jcmoraisjr:
The first version of the doc was written with a dedicated nodes for modsecurity in mind, which I think it isn't the canonical way of deploying workloads in Kubernetes, even the ones that demand more load. Hence the idea to use a deployment which can be easily scaled fronting with a service.

Anyway it's more like suggestions than a strong opinion or a guideline. My opinion is that the doc can be evolved to a single way of deploying modsecurity, preferably one that's used most, and the starting of the doc can shortly describe other ways to accomplish this, just like it's already doing. This makes thing easier for one that's starting with modsecurity and with the controller.

sealneaward:
I replaced the default config to use a deployment and kubectl expose to create the ClusterIP service.
Changed all related docs to do the same (config, auditlog sidecar).
Let me know if you want to see other changes.

sealneaward:
@jcmoraisjr any chance on another review?

jcmoraisjr:
Sure, sorry about taking so long (again). I want to run on a proper environment in order to validate the changes, but I'm investigating a bug for days which is taking all of my free time and it's postponing the v0.13 final tag. I already have some suggestions but want to send a single review in order to value your time. In the worst case I'll update this pr this weekend.

sealneaward:
No problem. If you have more pressing issues, then no rush.

docs/static/resources/modsecurity-deployment-auditlog-sidecar.yaml
```diff
@@ -0,0 +1,42 @@
...
+            emptyDir: {}
+          containers:
+          - name: modsecurity-spoa
+            image: quay.io/jcmoraisjr/modsecurity-spoa
+            args:
+            - -n
+            - "1"
+            ports:
+            - containerPort: 12345
+              hostPort: 12345
```
jcmoraisjr, line 30: We don't need `hostPort` anymore. Remember to remove from the `modsecurity-deployment.yaml` as well.

docs/static/resources/modsecurity-deployment-auditlog-sidecar.yaml
```diff
@@ -0,0 +1,42 @@
...
+            volumeMounts:
+            - name: varlog
+              mountPath: /var/log
+          - name: audit-log
+            image: busybox
+            args: [/bin/sh, -c, 'tail -n+1 -f /var/log/modsec_audit.log']
+            volumeMounts:
+            - name: varlog
+              mountPath: /var/log
+          hostNetwork: true
```
jcmoraisjr, line 42: `hostNetwork` should also be removed from both manifests.

jcmoraisjr (commented):
Hi, finally took the time to run de example and found the issues below.

sealneaward:
Sorry about that, made all of the edits from memory without actually applying them on a cluster.
Updated from review.

docs/content/en/docs/examples/modsecurity.md
```diff
@@ -26,31 +26,33 @@ adjust the steps to fit your need.
...
-Check if the agent is up and running:
 
+You can now create the service that provides a ClusterIP address for the HAProxy ConfigMap.
 ```
-$ kubectl -n ingress-controller get pod -lrun=modsecurity-spoa -owide
-NAME                     READY     STATUS    RESTARTS   AGE       IP               NODE
-modsecurity-spoa-pp6jz   1/1       Running   0          7s        192.168.100.99   192.168.100.99
+$ kubectl -n ingress-controller expose deployment modsecurity-spoa --port=12345 --type=ClusterIP
+```
+
```
jcmoraisjr, line 50: Missing a `service/modsecurity-spoa exposed` below this expose, just to follow the convention of the doc.

docs/content/en/docs/examples/modsecurity.md
```diff
@@ -136,8 +137,10 @@ Check the agent logs:
...
 $ kubectl -n ingress-controller get pod -lrun=modsecurity-spoa
-NAME                     READY     STATUS    RESTARTS   AGE
-modsecurity-spoa-5g5h2   1/1       Running   0          1h
+NAME                                READY   STATUS    RESTARTS   AGE
+modsecurity-spoa-6f757ffd88-9qt2f   1/1     Running   0          11m
+modsecurity-spoa-6f757ffd88-vwtzr   1/1     Running   0          11m
+modsecurity-spoa-6f757ffd88-q4rvm   1/1     Running   0          11m
 ...
 
 $ kubectl -n ingress-controller logs --tail=10 modsecurity-spoa-5g5h2
```
jcmoraisjr, line 146: Change this pod name to `modsecurity-spoa-6f757ffd88-9qt2f`.

docs/static/resources/modsecurity-deployment-auditlog-sidecar.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      run: modsecurity-spoa
+  template:
+    metadata:
+      labels:
+        run: modsecurity-spoa
+        spec:
```
jcmoraisjr, line 18: Need to remove 4 spaces from the whole `spec` field.

jcmoraisjr (commented):
Yeah just a few more changes I didn't see in the first run. After that it should be fine to merge.

sealneaward:
Update from review, let me know if there is anything else.

docs/static/resources/modsecurity-deployment-auditlog-sidecar.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      run: modsecurity-spoa
+  template:
+    metadata:
+      labels:
+        run: modsecurity-spoa
+      spec:
```
jcmoraisjr, line 18: spec should be inside template, sibling of metadata.

jcmoraisjr (commented):
spec is still not aligned, and that's it.

sealneaward:
Yeah, my bad. Been copying over changes I made from terraform to yaml.
Updated from reivew.

jcmoraisjr:
Lgtm, thanks! Merging now.

PR #828, opened 2021-08-01
Fix set ssl cert end-of-command

jcmoraisjr:
haproxy recognizes the end of an instruction after a line break. This however doesn't work with set ssl cert due to the multi line payload, so two line breaks should be sent. set ssl cert was broken after the socket refactor since two consecutive commands - set ssl cert and commit ssl cert - are sent in the same connection. This update adds two line breaks after the first command, and also improves a manual test with set/commit ssl cert.


jcmoraisjr:
Added to v0.13 as well.

PR #833, opened 2021-08-04
Gateway API: when using v1alpha1, certificateRef.group now accepts "core"

### LINKED ISSUE #830: haproxy-ingress expects an empty certificateRef group, but the Gateway API v1alpha1 requires a non-empty group
I am using haproxy-ingress v0.13.0-beta.2 with the v1alpha1 Gateway API CRDs (more specifically, v0.3.0).

I noticed that ingress-controller only accepts a certificateRef of the form:

```yaml
certificateRef:
   name: haproxy-tls
   kind: Secret
   group: ""
```

That's due to haproxyingress's [gateway.go](https://github.com/jcmoraisjr/haproxy-ingress/blob/23e24182b243020976a5582af655c76f3d4fa6a5/pkg/converters/gateway/gateway.go#L433-L435):

```go
func (c *converter) readCertRef(namespace string, certRef *gatewayv1alpha1.LocalObjectReference) (crtFile convtypes.CrtFile, err error) {
	if certRef.Group != "" {
		return crtFile, fmt.Errorf("unsupported Group '%s'", certRef.Group) // 🔥
	}
	...
}
```

which leads to haproxy-ingress skipping the listener:

```console
gateway.go:407] skipping certificate reference on Gateway 'default/echoserver': unsupported Group 'core' 
```

Unfortunately, the Gateway API CRDs `v1alpha1` (at v0.3.0) requires the `group` to be not empty as per the [local_object_reference_types.go](https://github.com/kubernetes-sigs/gateway-api/blob/dac9bed9f56d27f27bdac67926f49159c99af048/apis/v1alpha1/local_object_reference_types.go#L26) type.

For example, the following:

```yaml
kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.3.0" | kubectl apply -f -
kubectl apply -f- <<EOF
apiVersion: networking.x-k8s.io/v1alpha1
kind: Gateway
metadata:
  name: example-haproxy
spec:
  gatewayClassName: haproxy
  listeners:
  - hostname: haproxy.mael-valais-gcp.jetstacker.net
    protocol: HTTPS
    port: 443
    routes:
      kind: HTTPRoute
      selector:
        matchLabels:
          gateway: example-haproxy
    tls:
      mode: Terminate
      certificateRef:
        name: haproxy-tls
        kind: Secret
        group: ""
EOF
```

leads to the following error:

```
The Gateway "example-haproxy" is invalid: spec.listeners.tls.certificateRef.group: Invalid value: "": spec.listeners.tls.certificateRef.group in body should be at least 1 chars long
```

Note that `v1alpha2` (not released yet) has a fix for this (https://github.com/kubernetes-sigs/gateway-api/pull/562), but this fix was not backported to the `v1alpha1` CRDs.

maelvls:
The change was required due to the Gateway API v1alpha1 requiring a non-empty string in `certificateRef.group`. With v1alpha2, `""` is now a valid value as per https://github.com/kubernetes-sigs/gateway-api/pull/562.

Fixes #830.

### DIFF
--- a/pkg/converters/gateway/gateway.go
+++ b/pkg/converters/gateway/gateway.go
@@ -430,11 +430,15 @@ func (c *converter) applyCertRef(gwSource, routeSource *Source, hosts []*hatypes
 }
 
 func (c *converter) readCertRef(namespace string, certRef *gatewayv1alpha1.LocalObjectReference) (crtFile convtypes.CrtFile, err error) {
-	if certRef.Group != "" {
-		return crtFile, fmt.Errorf("unsupported Group '%s'", certRef.Group)
+	// In v1alpha1, the `group` cannot be left empty and is expected to be equal
+	// to "core" for the Secret kind. In v1alpha2, the `group` can be left empty
+	// and "core" is implied when the group is left empty. The discussion is
+	// available at https://github.com/kubernetes-sigs/gateway-api/pull/562.
+	if certRef.Group != "" && certRef.Group != "core" {
+		return crtFile, fmt.Errorf("unsupported Group '%s', supported groups are 'core' and ''", certRef.Group)
 	}
 	if certRef.Kind != "" && strings.ToLower(certRef.Kind) != "secret" {
-		return crtFile, fmt.Errorf("unsupported Kind '%s'", certRef.Kind)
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Sorry about taking so long to revise and thanks for sharing and patching this issue with us! Lgtm, merging now.

PR #841, opened 2021-08-27
docs: add modsec resource limits to controls V2 memory consumption

sealneaward:
From our deployment/daemonset of ModSecurity, we have been noticing the ModSecurity pods hogging memory after our performance tests finish. There is no garbage collection done and the memory consumed only increases over time.

We believe this is due to documented memory leak issues with V2 of ModSecurity.
Our kubernetes native solution to this is to provide memory allocations to the deployment, to force restarts of the pods instead of hogging memory on nodes.
We've ran performance tests against this change to the deployment and there is no performance regression.

### DIFF
--- a/docs/content/en/docs/examples/modsecurity.md
+++ b/docs/content/en/docs/examples/modsecurity.md
@@ -33,6 +33,11 @@ $ kubectl create -f https://haproxy-ingress.github.io/resources/modsecurity-depl
 deployment.apps/modsecurity-spoa created
 ```
 
+{{% alert title="Note" %}}
+This deployment configures a small amount of requests and limits resources,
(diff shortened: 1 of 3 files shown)


jcmoraisjr:
Sounds fair. Please add also a note in the example doc so one will be advised that there is a small requests and limits configuration. Something like this just after the first `kubectl create` is enough:

```
{{% alert title="Note" %}}
This deployment configures a small amount of requests and limits resources,
remember to adjust them before moving to production.
{{% /alert %}}
```

sealneaward:
Added the alert. Let me know if you want to see anything else @jcmoraisjr

jcmoraisjr:
LGTM thanks! Merging now.

Issue #840, opened 2021-08-25
ConfigMap based TCP service endpoints don't update


bootc:
**Description of the problem**

When using ConfigMap based TCP, service endpoints changes do not update HAProxy.

That is, if I have a service exposed using ConfigMap based TCP and I delete some of the pods that run this service, the old endpoint addresses are never removed from `haproxy.cfg` and their replacements are never added.

**Expected behavior**

As endpoints change the `haproxy.cfg` should update to include the new endpoints.

**Steps to reproduce the problem**

1. Configure an example Service + Deployment.
2. Examine the addresses in `haproxy.cfg`.
3. Delete some of the pods.
4. Note how the old addresses remain in `haproxy.cfg` and the new addresses are absent.

**Environment information**

HAProxy Ingress version: `v0.13.1`
Helm chart version: `0.13.1`
Using the "external" HAProxy with the Helm chart, which runs HAProxy as a sidecar.

Command-line options:

```
--configmap=bootc-system/haproxy-ingress \
--ingress-class=haproxy \
--master-socket=/var/run/haproxy/master.sock \
--sort-backends \
--tcp-services-configmap=bootc-system/haproxy-ingress-tcp \
--ignore-ingress-without-class=true \
--wait-before-shutdown=30
```

Global options:

```
bind-ip-addr-http: '[::]'
bind-ip-addr-tcp: '[::]'
healthz-port: "10253"
https-log-format: default
prometheus-port: "9101"
stats-port: "1936"
syslog-endpoint: stdout
syslog-format: raw
tcp-log-format: default
```

Ingress objects:

N/A

TCP ConfigMap:

```
"8140": puppet/puppetserver:8140
```

jcmoraisjr:
Hi thanks for reporting. This is an unfortunate behavior added when Gateway API parsing was included. ConfigMap based TCP services will only parse again if a global config is changed or when the ConfigMap with the TCP services is updated. I'll issue a new tag shortly that should fix this.

jcmoraisjr:
Fixed in v0.13.2. Closing.

PR #842, opened 2021-08-28
Fix endpoint update of configmap based tcp services

jcmoraisjr:
Legacy ConfigMap based TCP services has its own endpoint list that doesn't track partial parsing. The entry point was changed when Gateway API was added, it was was incorrectly configured to only run when a full sync is needed, or when the ConfigMap was changed, leading to new state not being reflected in the configuration. This update ensures that ConfigMap based TCP services is always parsed when something changes in the cluster.


jcmoraisjr:
#840

Issue #838, opened 2021-08-13
oauth2-proxy configuration insecure


mkkot-onegini:
The configuration provided here:
https://github.com/jcmoraisjr/haproxy-ingress/blob/master/examples/auth/oauth/app.yaml

works as expected when oauth2-proxy service is available on the cluster. If it gets removed for some reason (human error, misconfiguration), the haproxy will allow connection to secured service without any authentication and authorization.

```
I0812 14:22:10.153076       7 controller.go:281] Starting HAProxy update id=436
W0812 14:22:10.153561       7 ingress.go:143] skipping backend config of ingress 'app': service namespace/oauth2-proxy was not found
```

The haproxy should not allow unauthenticated traffic to the service.

**Steps to reproduce the problem**

1. Deploy the example configuration from https://github.com/jcmoraisjr/haproxy-ingress/tree/master/examples/auth/oauth
2. Remove oauth2-proxy service
3. Try to reach the echoservice via ingress

**Environment information**

HAProxy Ingress version: v0.10.9

jcmoraisjr:
v0.13 has a new auth external and oauth implementation that doesn't have this issue. v0.12 and olders were already fixed and a new release should be issued shortly.

jcmoraisjr:
All supported versions updated. Closing.

PR #843, opened 2021-08-28
Always deny requests if oauth is misconfigured

jcmoraisjr:
An attempt to enable oauth should end up with a proper configuration, or the request should always fail in order to protect the backend service against misconfigurations. This change ensures that an attempt to configure oauth will lead to a proper configuration, or all requests will be denied.

This already works this way on v0.13, should be merged between v0.12 and v0.10.


jcmoraisjr:
#838

Issue #831, opened 2021-08-02
TLS cert verification failing when ca.crt only contains root cert and tls.crt contains intermediate and leaf cert


jpiper:
**Description of the problem**

When my ingress references TLS secret that has a tls.key, tls.crt and ca.crt, with the ca.crt containing the root cert, and the tls.crt containing the intermediate and leaf cert, the ingress controller says

> syncing a non ca/crt secret httpbin/httpbin-test-rw-tls
> using default certificate due to an error reading secret 'httpbin-test-rw-tls' on ingress 'httpbin/httpbin-test-rw': secret 'httpbin/httpbin-test-rw-tls' have neither ca.crt nor tls.crt/tls.key pair

When I delete the ca.crt from the TLS object and then edit the ingress to force a reconciliation it all loads fine.

**Expected behavior**

I expect the certs to load.

**Environment information**

HAProxy Ingress version: `v0.12.6`

Command-line options:

```
        - --default-backend-service=$(POD_NAMESPACE)/ingress-default-backend
        - --configmap=$(POD_NAMESPACE)/haproxy-ingress
        - --sort-backends
        - --reload-strategy=reusesocket
```

Global options:

```
    http-port: "10199"
    https-port: "10200"
    ssl-ciphers: ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305
    ssl-options: force-tlsv12
    syslog-endpoint: localhost:514
    timeout-client: 60s
    timeout-http-request: 60s```

Ingress objects:

```
<redacted>
```

**Comment by jpiper** [comment]
ah, so I noticed you never print the actually error here https://github.com/jcmoraisjr/haproxy-ingress/blob/master/pkg/common/ingress/controller/backend_ssl.go#L45

I changed this to: `glog.V(3).Infof("syncing a non ca/crt secret %v (reason: %v)", key, err)` and then compiled/deployed and saw that the issue is now:

```
I0802 23:52:18.689318       7 backend_ssl.go:45] syncing a non ca/crt secret httpbin/httpbin-test-rw-tls (reason: unexpected error creating pem file: failed to verify certificate chain: 
	x509: certificate signed by unknown authority
)
```

[...]

PR #845, opened 2021-08-29
Fix validation if ca is used with crt and key

jcmoraisjr:
An undocumented but inherited behavior from the so called ingress controller common code base allows to use a ca.crt key along with crt and key. However a crt validation was failing due to the missing intermediate crts in the validation. There is no behavioral change when using crt and key without a ca.


jcmoraisjr:
#831

PR #846, opened 2021-08-29
Feature/allowlist behind reverse proxy

DCkQ6:
Added http request set-src directive to backend definition.

This feature is required to control access to specific paths while ingress is hidden behind reverse proxy. It is ment to use only with trusted headers, where any header modification is strictly controlled by reverse proxy and ingress itself is in separated - air gapped - network.


rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -512,6 +512,9 @@ backend {{ $backend.ID }}
 {{- range $e1 := short 10 $allow.Exception }}
     acl allow_exception_src{{ $i }} src{{ range $e := $e1 }} {{ $e }}{{ end }}
 {{- end }}
+{{- with $allow.SourceHeader }}
```
jcmoraisjr, line 515: Change to `if` instead just to follow the convention.

docs/content/en/docs/configuration/keys.md
```diff
@@ -677,6 +678,7 @@ See also:
 | `allowlist-source-range` | `Path` |         | v0.12 |
 | `denylist-source-range`  | `Path` |         | v0.12 |
 | `whitelist-source-range` | `Path` |         |       |
+| `allowlist-source-header`| `Path` |         | v0.1x |
```
jcmoraisjr, line 681: This can be merged to v0.13, so please change to v0.13.2 here.

jcmoraisjr (commented):
Thanks for the contribution! Following some minor changes.

rootfs/etc/templates/haproxy/haproxy.tmpl
DCkQ6: done :+1:

docs/content/en/docs/configuration/keys.md
DCkQ6: done :+1:

jcmoraisjr:
LGTM, merging now, thanks!

PR #849, opened 2021-09-04
docs: update api group of rbac config

chitoku-k:
Due to the deprecation of beta Ingress APIs, they should also been removed/modified in the way supported in Kubernetes v1.22+. Also, `ingressclasses` has been missing since some time haproxy-ingress started to integrate with them.
https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

Thanks in advance!

### DIFF
--- a/docs/haproxy-ingress.yaml
+++ b/docs/haproxy-ingress.yaml
@@ -41,6 +41,7 @@ rules:
       - list
       - watch
   - apiGroups:
+      - extensions
       - networking.k8s.io
     resources:
       - ingressclasses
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Hi, these are some old configuration files that's being still referenced, unfortunately. So it seems good to update it and thanks for the contribution. In this case, please maintain the extensions api group because some supported versions still need it.

chitoku-k:
OK, fixed!

jcmoraisjr:
LGTM, thanks! Merging.

Issue #854, opened 2021-09-14
Config-backend config map key not working as expected in v0.12.9


rexhinvorpsi:
**Description of the problem**
Hi we switched our haproxy version from 0.12.3 to 0.12.9 to remidiate the HTTP/2 vulnerability. 
After that the config-backend snippet is not set in the haproxy.cfg even though it is set in the config-map correctly. We traced down to the last working version being 0.12.5.

**Expected behavior**
backend snippets to be set in haproxy.cfg

**Environment information**

HAProxy Ingress version: `v0.12.9`

jcmoraisjr:
Sorry about the annoying regression. I can confirm that backend snippet configuration is broken since #820, planning to tag a new version later this week. Please confirm if only ConfigMap based (global) configuration snippet is broken, as far as I can see, annotation based config snippets are still working.

jcmoraisjr:
Fixed in all supported versions. Closing.

PR #856, opened 2021-09-15
Fix global config-backend snippet config

jcmoraisjr:
The mitigation of CVE-2021-25740 added a command-line option to allow a sysadmin to partially or completely remove backend configuration snippet. This option added a new code where there was only a straightforward assignment. This code fails on how to check if a configuration has a source and, instead of adjust the logging message, the whole assignment was being skipped. However a global configuration by definition doesn't have a source, leading to backend configuration snippet being skipped if not configured via ingress or service annotation.

This fix should be merged up to v0.10.


jcmoraisjr:
#854

PR #857, opened 2021-09-23
Add --ingress-class-precedence to allow IngressClass taking precedence over annotation

mkubaczyk:
When defined as true, `--controller-class-precedence` with give a precedence to controller class over kubernetes.io/ingress.class annotation. This will be especially useful while migrating from one approach to another having zero downtime in mind. It can be removed once the support of ingress annotation is fully disabled.

I am here referring to situation when both IngressClassName and ingress annotation are defined in Ingress which results in
```W0923 07:50:41.725353       7 cache.go:715] ingress prometheus/prometheus has conflicting ingress class configuration, using annotation reference (false)``` in haproxy ingress controller logs.

### DIFF
--- a/docs/content/en/docs/configuration/command-line.md
+++ b/docs/content/en/docs/configuration/command-line.md
@@ -36,6 +36,7 @@ The following command-line options are supported:
 | [`--health-check-path`](#stats)                         | path                       | `/healthz`              |       |
 | [`--healthz-port`](#stats)                              | port number                | `10254`                 |       |
 | [`--ingress-class`](#ingress-class)                     | name                       | `haproxy`               |       |
+| [`--ingress-class-precedence`](#ingress-class)          | [true\|false]              | `false`                 | v0.13.5 |
 | [`--kubeconfig`](#kubeconfig)                           | /path/to/kubeconfig        | in cluster config       |       |
 | [`--master-socket`](#master-socket)                     | socket path                | use embedded haproxy    | v0.12 |
 | [`--max-old-config-files`](#max-old-config-files)       | num of files               | `0`                     |       |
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by mkubaczyk** [comment]
I haven't add reviewers so requesting a review from @jcmoraisjr :-)

**Comment by jcmoraisjr** [comment]
Thanks for the contribution and sorry about taking so long to comment.

I'd suggest you two main changes:

```golang
		ingressClassPrecedence = flags.Bool("ingress-class-precedence", false,
			`Defines if IngressClass resource should take precedence over
kubernetes.io/ingress.class annotation if both are defined and conflicting.`)
```

This is a suggested change in the name and description with two benefits: better fits the purpose of the functionality and also avoids to change the whole struct, which leads to conflict when merging or cherry-picking.

The other suggestion is to try to add the warning message if cfg is true - ie using the functionality will only change the precedence in the case of a conflict, but still warning when the conflict happens.

Please let me know if you need anything, I'll answer faster this time =)

mkubaczyk:
@jcmoraisjr sounds reasonable. I updated the code. Please take a look to make sure I have not skipped some condition. :-)

jcmoraisjr:
I ended up with this piece of code, it'll only add a new logic, compared with the original code, if the config is true. Please check if it makes sense to you:

```golang
	// c.cfg.IngressClassPrecedence as `true` gives precedence to IngressClass
	// if both class and annotation are configured and they conflict
	if hasAnn {
		if hasClass && fromAnn != fromClass {
			if c.cfg.IngressClassPrecedence {
				c.logger.Warn("ingress %s/%s has conflicting ingress class configuration, "+
					"using ingress class reference because of --ingress-controller-precedence enabled (%t)",
					ing.Namespace, ing.Name, fromClass)
				return fromClass
			}
			c.logger.Warn("ingress %s/%s has conflicting ingress class configuration, using annotation reference (%t)",
				ing.Namespace, ing.Name, fromAnn)
		}
		return fromAnn
	}
```

You can also add the starting version, this pr will be cherry-picked to v0.13:

```
| [`--ingress-class-precedence`](#ingress-class)          | [true\|false]              | `false`                 | v0.13.5 |
```

Adjust the alignment in the `Configuration` struct, both in controller (declaration) and launch (initialization) files.

And finally, try to squash all the resulting commits into a single one.

Thanks!

mkubaczyk:
@jcmoraisjr should be fine now

jcmoraisjr:
Great, lgtm, thanks! Merging now.

mkubaczyk:
@jcmoraisjr are you planing to release some new minor version that would also contain this change? :-)

jcmoraisjr:
Hi, sorry about the delay in the response. v0.13.5 was just released and has this feature merged.

PR #861, opened 2021-10-09
remove setting vary origin header always when multiple origins are set

michal800106:
it overwrites another settings header vary from annotations. We should remember to set manually vary origin when we
have multiple origins.

### DIFF
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -135,14 +135,16 @@ func TestBackends(t *testing.T) {
     http-response set-header Access-Control-Allow-Origin  "http://d11.local" if { var(txn.pathID) path01 path02 } !cors_allow_origin0
     http-response set-header Access-Control-Allow-Methods "GET, PUT, POST, DELETE, PATCH, OPTIONS" if { var(txn.pathID) path01 path02 }
     http-response set-header Access-Control-Allow-Headers "DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization" if { var(txn.pathID) path01 path02 }
-    http-response set-header Vary Origin if { var(txn.pathID) path01 path02 }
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
Hi, could you elaborate a bit more? Afaics this header is only defined if cors-allow-origin declares more than one header, in this case it should be added, doc [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#the_http_response_headers). Maybe change it to add-header instead?

jcmoraisjr:
Nice, thanks! This should only be a problem if you use cors with more than one allow-origin, is this your use case?

michal800106:
Yeah exactly it is our case. We are vary happy that haproxy-ingrex added possibility for using multiple allow-origin domains. But we need also possibility for caching per language and for this purpose we use a vary header.

michal800106:
I checked with add-header vary Origin: the response has multiple vary instances:
```
vary: accept-language
vary: Origin
```
On stack-overflow I found this link
https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2
and I'm not sure is it O.K. to have separated values?

jcmoraisjr:
Try this one:

```
    http-response set-header Vary %[res.hdr(Vary)],Origin
        {{- "" }} if { res.hdr(Vary) -m found }
        {{- if $pathIDs }} { var(txn.pathID) {{ $pathIDs }} }{{ end }}
    http-response set-header Vary Origin
        {{- "" }} if ! { res.hdr(Vary) -m found }
        {{- if $pathIDs }} { var(txn.pathID) {{ $pathIDs }} }{{ end }}
```

I'll double check if we have a less verbose option.

jcmoraisjr:
Let's see what [people say here](https://www.mail-archive.com/haproxy@formilux.org/msg41302.html) and then please edit this same pr - you found the issue, you have the credits. =) Remember to try to squash the new commit with this one and check if this will break the test (it should).

michal800106:
O.K. I've added already you proposed code for this PR. I will change it if on mailing list arrive better option. I've squashed commits as you said. I've fixed tests.

jcmoraisjr:
No feedback from the mailing list, but it seems good in my opinion, so merging now. Thanks!

PR #867, opened 2021-11-16
Fix typo: s/distict/distinct

juggernaut:


jcmoraisjr:
Thanks! Merging. Note that the change will be applied in the doc only in the next release.

Issue #709, opened 2020-12-07
Support arm64 - raspberry pi


ieugen:
I tried deploy haproxy ingress chart on a raspberry pi cluster for demos and I got 

```
exec user process caused: exec format error
```
It would be nice if there are instructions on how to use this chart with arm systems.
I believe just the right image is required.

nicolass67:
Hi ieugen,

I had the same issue. 
The haproxy ingress can be build on raspberry pi. I'm doing it on rpi4. It takes around 5-10 minutes of build time. 

My procedure: 
`https://github.com/jcmoraisjr/haproxy-ingress/releases/tag/v0.13.3`
`git checkout tags/v0.13.3 && cd haproxy-ingress`
`docker build -f builder/Dockerfile -t haproxy-ingress-arm:v0.13.3 .`

For the moment I didn't had time to test the image, but it's already a good beginning that it builds. 

Regards,

jcmoraisjr:
arm64 is supported since v0.13, please let me know if you can use the generated image.

nicolass67:
Hi Joao, 
I use a RPi 4 with the official Raspberry Pi OS (based on debian + update to deb 11). Kubernetes is 1.22.2. 
It seems that this OS is this arch : `Linux XXXXXX armv7l GNU/Linux`

When using the Helm Chart for haproxy-ingress, the pod doesnt start : 
`haproxy-ingress   haproxy-ingress-tg9s9                   0/1     ImagePullBackOff `

Looking at the events : 
`Failed to pull image "quay.io/jcmoraisjr/haproxy-ingress:v0.13.3": rpc error: code = Unknown desc = no matching manifest for linux/arm/v7 in the manifest list entries`

Maybe a simple retag of arm64 could do the trick ? 
As said I manage to build the docker file on this machine. I just need time to test it. 

Regards,

jcmoraisjr:
Ah ok this doesn't seem to be supported yet, we actually only build arm/v8:

https://github.com/jcmoraisjr/haproxy-ingress/blob/420969f1065f97468f26d0ad8c8cb542e36db9f0/.github/workflows/image.yaml#L37-L42

Every new platform takes about 10 minutes to generate the image, but I'll evaluate to add some other widely used ones.

nicolass67:
@jcmoraisjr : I built the image on the rpi4, pushed it into a private registry and used the helm chart to use this image : it works (tests: ingresses/ingressClases etc ...). 
[...]

PR #870, opened 2021-12-06
Add new target platforms

jcmoraisjr:
Add target platform images for `linux/arm/v7`, `linux/arm/v6` and `linux/s390x`.


jcmoraisjr:
#709

PR #874, opened 2021-12-26
Bump dependencies

mrueg:
This change bumps golang to 1.16.12 and updates dependencies in go.mod.


jcmoraisjr:
Great, thank you! We should have a v0.14 tag in a month or so, I'll leave this open so we can merge an up to date client-go and golang versions.

mrueg:
> Great, thank you! We should have a v0.14 tag in a month or so, I'll leave this open so we can merge an up to date client-go and golang versions.



Do you want me to upgrade to go 1.17 and k8s 1.23?

jcmoraisjr:
Both golang and Kubernetes releases have a faster step if compared with haproxy ingress, so yes go 1.17 and k8s 1.23 seems to be a more reasonable options.

jcmoraisjr:
Hi, we're really late with v0.14 and maybe we'll issue the first tag in the next few days. Please confirm if you can squash all the commits into a single one and rebase on top of master, I'll merge it just after that. Btw thanks for maintaining this branch up to date.

mrueg:
@jcmoraisjr working on fixing conflicts and squashing

builder/Dockerfile
```diff
@@ -12,16 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.16.5-alpine AS builder
+FROM golang:1.17.6-alpine AS builder
 
 ARG GIT_TAG
-
+WORKDIR /src
```
mrueg, line 18: @jcmoraisjr small additional change, it's preferred over using cd

jcmoraisjr: sure, sounds good.

jcmoraisjr:
My local env suggested me to move `gopkg.in/go-playground/assert.v1` to the second require argument, can you confirm this?

mrueg:
> My local env suggested me to move `gopkg.in/go-playground/assert.v1` to the second require argument, can you confirm this?

Yes, you're correct. I fixed it! Thanks for catching that :)

jcmoraisjr:
LGTM, thanks! Merging now.

Issue #873, opened 2021-12-20
working config map on v0.11 regarding backend-config snippet gets error on v0.13


KamranSwish:
**Description of the problem**

After migrating to v0.13 from v0.11, I see some warrning like bellow:

**W1220 11:55:26.839441       7 backend.go:541] skipping configuration snippet on global config: keyword '' not allowed**

for each line on config-backend multiline configmap configuration. those ignored configuration snippets contain stick-table and some other ACLs which are needed for applications to work properly.
some facts:
 1. ingress controller restart due to missing stick table in one backend(this is a secondary issue, if we fix other this will automatically be solved)
 2. the config-backend multiline content is not added to haproxy.cfg backends section.
 3. I tried to move multiline config-backend snippet to ingress annotation or ingressclass configmap. the same results but no success. I do not have clear idea how to apply single configmap (containing ACLs and stick-table entries) to all backends!
 4. I use terraform deploy ingress controller configuration

**Expected behavior**

the new version supports old config-backend multiline syntax or there I could find some clear documents what changes should I do!

**Steps to reproduce the problem**

1. using new v0.13 ingress controller image on ingress controller deployment.
2. no changes to old configmap

**Environment information**

HAProxy Ingress version: `v0.13`

Command-line options:

```
      - args:
        - --configmap=$(POD_NAMESPACE)/haproxy-ingress
        - --ingress-class=haproxy
        - --sort-backends
        - --default-ssl-certificate=file:///etc/folder/certs/app.pem```

Global options:

```
  backend-server-naming: pod
  config-backend: |
    option h1-case-adjust-bogus-server
    stick-table type string len 32 size 1m expire 300s

    acl ...
    acl ...
    .
    .
    .
    stick match ...
    stick match ...
    stick match ...
    stick store-request ...
    stick store-response ...
    stick store-response ...
  config-defaults: |
    ...
  config-frontend: |-
    ...
  config-global: |-
    ... 

```

Ingress objects:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/maxconn-server: "1024"
    kubernetes.io/ingress.class: haproxy
```

KamranSwish:
I see some restrictions has been added in the following pull request, 

https://github.com/jcmoraisjr/haproxy-ingress/pull/820/commits/b9b3f12f6dedce370dd8d88951cd9b45728d0e80

I just want some one helps me where should I move the definisions!
[...]

PR #876, opened 2021-12-27
Add disableKeywords only if defined

jcmoraisjr:
`--disable-config-keywords` command-line option allows to define a comma separated list of blacklisted keywords. If not declared, an empty string is used as an input to `Split()`. This empty string is however converted in an array of one element with an empty string, and such empty keyword matches with empty lines, denying configurations with empty lines. This update ignores empty entries when parsing the custom snippet, and also `TrimSpace()` all the options in order to properly match with the custom snippet tokenizer.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strings"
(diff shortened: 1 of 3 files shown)


jcmoraisjr:
#873

PR #887, opened 2022-01-16
build: update to Go 1.17

Juneezee:
This PR introduces three changes:

1. Update the `go` directive in `go.mod` file by running `go mod tidy -go=1.17`.
    **Note 1:** This does not prevent users with earlier Go versions from successfully building packages from this module.

    **Note 2:** The additional `require` directive is used to record indirect dependencies for Go 1.17 or higher, see https://go.dev/ref/mod#go-mod-file-go.

2. Update GitHub Actions to Go 1.17

3. Update Dockerfile to use `golang:1.17.6-alpine`

Go 1.16 will soon reach its end-of-life when Go 1.18 is released in February 2022.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,9 +11,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/setup-go@v1
+    - uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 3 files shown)


jcmoraisjr:
Hi thanks for the contribution! It seems a dupe of #874 which will be merged just before the first v0.14 tag.

Issue #888, opened 2022-01-17
Domain attribute in Set-Cookie header should not have multiple possible values


hbobenicio:
Hi, @jcmoraisjr . Hope you're doing well!

**Description of the problem**

Sorry if not very descriptive or precise, I'm not an k8s/ingress expert. I'll try my best.

https://github.com/jcmoraisjr/haproxy-ingress/blob/5887cc3aeb9682668f0b241a9b0b14375ab4c844/rootfs/etc/templates/haproxy/haproxy.tmpl#L676

[haproxy.tmpl](https://github.com/jcmoraisjr/haproxy-ingress/blob/5887cc3aeb9682668f0b241a9b0b14375ab4c844/rootfs/etc/templates/haproxy/haproxy.tmpl#L676) is rendering multiple values for the domain attribute, which results in multiple domain attributes set for the Set-Cookie response header, when using *Session Cookies* and trying to *share* them. Headers like this may then be returned by the ingress:

```
set-cookie: INGRESSCOOKIE=2efcdeadcafe287c; path=/; domain=foo.org; domain=foo.com; domain=www.foo.com; HttpOnly
```

The thing is that multiple attributes with the same key (in this case, "domain"), happens to *just override* previous values. Servers should not rely on this either. This is what I found about it:

> "[...] To maximize compatibility with user agents, servers SHOULD NOT produce two attributes with the same name in the
   same set-cookie-string. [...]"
>
> https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1

> "[...] If the cookie-attribute-list contains an attribute with an
>        attribute-name of "Domain":
>
>           Let the domain-attribute be the attribute-value of the last
>           attribute in the cookie-attribute-list with an attribute-name
>           of "Domain".
>
> https://datatracker.ietf.org/doc/html/rfc6265#section-5.3

In the above example, user agents should then only send INGRESSCOOKIE to the last value, www.foo.com.

If accessing "foo.org" or "foo.com" with Google Chrome it warns us about this response header defect like this:

```
"This attempt to set a cookie via a Set-Cookie header was blocked because its Domain attribute was invalid with regards to the current host url"
```

Then the Cookie Sharing between multiple domains does not work.

In fact, the correct way if one wants to define the *domain* Set-Cookie attribute is giving just one value. And in the context of sharing it to different domains, only subdomains (in my understanding) are allowed. This is what I found.

> "The Domain attribute specifies those hosts to which the cookie will
[...]

PR #889, opened 2022-01-18
Add session-cookie-domain configuration key

jcmoraisjr:
Modern browsers doesn't support more than one `domain` attribute in the set-cookie header, deprecating the `session-cookie-domain` configuration key that's used to share a cookie between more than one domain. The new key `session-cookie-domain` allows the user to configure which value the only domain attribute should have.


jcmoraisjr:
#888

Issue #892, opened 2022-01-23
CVE-2020-29652, CVE-2021-38561


rgherta:
**Description of the problem**

Pasting from the helm charts vulnerability scan

<html>
<body>

  | ID | SEVERITY | PACKAGE | VERSION | FIXED IN
-- | -- | -- | -- | -- | --
  | CVE-2020-29652 | HIGH | golang.org/x/crypto | v0.0.0-20201002170205-7f63de1d35b0 | v0.0.0-20201216223049-8b5274cf687f
  | CVE-2021-38561 | UNKNOWN | golang.org/x/text | v0.3.6 | 0.3.7

</body>
</html>

golang: crypto/ssh: crafted authentication request can lead to nil pointer dereference
A nil pointer dereference in the golang.org/x/crypto/ssh component through v0.0.0-20201203163018-be400aefbc4c for Go allows remote attackers to cause a denial of service against SSH servers.

**Expected behavior**

We probably need to build with the new version of crypto and text

**Steps to reproduce the problem**

https://artifacthub.io/packages/helm/haproxy-ingress/haproxy-ingress?modal=security-report

**Environment information**

HAProxy Ingress version: `v0.13.6`

jcmoraisjr:
Finally tagged and released in v0.13.7. Thanks. Closing.

PR #893, opened 2022-01-23
fixes for #892

rgherta:


jcmoraisjr:
Hi thanks for the contribution! Following a couple of suggestions:

- Remove and recreate go.sum with `rm go.sum && go mod tidy` so unused entries are removed
- `docker-build` makes more sense so it's preferable to change the Makefile

On top of that I'd also suggest you to bring these two changes on two distinct PR, this will make my life easier when building the release notes.

PR #905, opened 2022-02-21
Add haproxy 2.5 support for external haproxy

jcmoraisjr:
Creates a new `show proc` parser compatible with haproxy 2.2 to 2.5. Added in a safe way to allow merging as far as controller v0.12 - version 2.4 and below will continue to use the old parser, minimizing the risk of adding an unknown bug from the new one.

Should be merged as far as v0.12, which is the first version that implements external haproxy and parses `show proc` output.

### DIFF
--- a/pkg/haproxy/socket/socket.go
+++ b/pkg/haproxy/socket/socket.go
@@ -233,6 +233,9 @@ type Proc struct {
 	PID     int
 	RPID    int
 	Reloads int
+	Failed  int
+	Uptime  string
+	Version string
 }
 
 // HAProxyProcs reads and converts `show proc` from the master CLI to a ProcTable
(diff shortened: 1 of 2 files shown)


sifr0x:
Is 2.5 support ready to merge?
If not I'll be happy to help you on this PR.

jcmoraisjr:
Hi yes, this should be fine to be merged, missing only the last revision and a new exploratory test. I needed to give some priority to other tasks and could evolve the controller in the last weeks (months?) as much as I want. Planning to add this functionality as far as v0.12 and tag new releases within the next week. Any help is very much appreciated, so if you have the skills to build and run the controller locally, you could have a look if 2.5 works without any issue with this pr. Otherwise, no problem, a new v0.12..v0.14 tag should be available shortly.

Issue #906, opened 2022-03-18
For default/sink ingress paths HAproxy expecting a trailing slash


arun11299:
HAProxy ingress version : 0.13.6

Kubernetes : GKE
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5-gke.1805", GitCommit:"6be6bf42dd3e740d8515b47dee3c9fc3ddf602b2", GitTreeState:"clean", BuildDate:"2022-02-18T05:04:07Z", GoVersion:"go1.16.7b7", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.23) and server (1.21) exceeds the supported minor version skew of +/-1
```
haproxy configuration. configmap:
```
Name:         haproxy-configmap
Namespace:    ar-lp
Labels:       app.kubernetes.io/managed-by=Helm
Annotations:  meta.helm.sh/release-name:  ar-lp
              meta.helm.sh/release-namespace: default

Data
====
bind-https:
----
:8443
config-backend:
----
timeout server 500s

config-frontend:
----
http-response replace-header Set-Cookie (.*) \1;\ Secure

bind-http:
----
:8080
```

Ingress config:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/backends: '{"k8s-be-32289--cb238fb942e4da41":"HEALTHY","k8s1-cb238fb9-ar-lp-ar-lp-grafana-8300-2ab57ed8":"UNHEALTHY","k8s1-cb238fb9-ar-lp-ar-lp-launchpad-8008-c5e04f6b":"UNHEALTHY","k8s1-cb238fb9-ar-lp-ar-lp-launchpad-8009-e5256e99":"UNHEALTHY","k8s1-cb238fb9-ar-lp-ar-lp-launchpad-8443-63aa6161":"UNHEALTHY"}'
    ingress.kubernetes.io/forwarding-rule: k8s2-fr-595g39yp-ar-lp-rw-ingress-eqa0flxs
[...]

PR #908, opened 2022-03-21
Fix match of prefix pathtype if using default host

jcmoraisjr:
HAProxy's subdir match type is used to implement the `Prefix` pathType. This match type however doesn't implement exactly what is expected from the `Prefix` type if a single slash is used as the pattern - ingress expects that `/bar` would match, but it doesn't, matching only `/bar/` instead. This works well with declared hostnames because HAProxy Ingress builds the pattern concatenating hostname and path, but this doesn't happen with the default hostname where the path is configured alone.

This update adds a prefix in the default hostname patterns, just like it's already happening with declared hostnames, in order to them to behave in the same way.


jcmoraisjr:
#906

PR #910, opened 2022-03-22
remove initial whitespaces from haproxy template

ironashram:
In conjunction with https://github.com/haproxy-ingress/charts/pull/48 improves custom haproxy template usage


jcmoraisjr:
Thanks. Once these white spaces are the result of an attempt of a kind of an art, I'd suggest instead to change the first space into a hashtag. The `map.tmpl` file had a similar issue in the past and now it's *designed* like this:

https://github.com/jcmoraisjr/haproxy-ingress/blob/c934e1d6306341ca6dd77a925c5c6b359dab00c6/rootfs/etc/templates/map/map.tmpl#L1-L8

Issue #881, opened 2022-01-07
add SHA-2 to ingress.kubernetes.io/auth-tls-cert-header


Yosique:
According to https://www.ietf.org/rfc/rfc9155.txt SHA1 and MD5 should no longer be used.
In https://github.com/haproxy/haproxy/issues/123 support for SHA-2 was added for haproxy.
Please add `X-SSL-Client` headers for the SHA-2 family.

jcmoraisjr:
Added to v0.14.0-alpha.2, thanks for the suggestion. Closing.

PR #911, opened 2022-03-23
Add ssl-fingerprint-sha2-bits configuration key

jcmoraisjr:
`ssl-fingerprint-sha2-bits` defines the number of bits of the SHA-2 fingerprint of the client certificate of a mTLS request. The header `X-SSL-Client-SHA2` is only added if the configuration is declared.


jcmoraisjr:
#881

PR #914, opened 2022-04-05
Only inspect SSL handshake for SNI routing for SSL passthrough, fixes #912

### LINKED ISSUE #912: TLS termination with SNI-based routing on TCP port always delays connection by 5 seconds

Enabling TLS termination and SNI-based routing on a TCP service (using the `tcp-service-port` annotation) with a specific host-rule causes a delay of 5 seconds regardless of the value of the server name transmitted by the client.

The culprit appears to be this line in the configuration template:
https://github.com/jcmoraisjr/haproxy-ingress/blob/b30446862cb1e29efbc12fcddda8fef5a2f47f08/rootfs/etc/templates/haproxy/haproxy.tmpl#L1031

The delay comes from the `tcp-request inspect-delay 5s` instruction, but the delay should *not* be hit as the variable match on `req.ssl_fc_sni` should occur long before the delay is over.

As the frontend at that point has a `bind [...] ssl [...]` instruction, the `req.ssl_hello_type` value will not be set in this case, see the [HAProxy configuration manual](https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#7.3.5-req.ssl_hello_type) for details. I think the fix is as simple as removing that line, since we know our connection is TLS-enabled at that point, but I didn't want to submit a PR yet in case there's something I'm missing there.

I've amended the template in my own instance and the connection was established immediately when initiated with the correct SNI information, but closed if the host didn't match the configured one. So this one-line change would definitely solve my particular case.

**Expected behavior**

The connection to the appropriate backend (based on SNI) should be established immediately after the TLS handshake.

1. Create an ingress with `tcp-service-port`, a `tls` section and a `rule` matching the desired host
   * Do this for e.g. a simple echo server
2. Connect to the service using e.g. `openssl s_client`
3. Observe a delay of 5 seconds before the backend sends an answer

**Environment information**

HAProxy Ingress version: `v0.13.6`

Ingress objects:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: tcp-echo
  annotations:
    kubernetes.io/ingress.class: haproxy
    haproxy-ingress.github.io/tcp-service-port: 12345
spec:
  tls:
    - hosts:
        - tcp-echo.example.com
  rules:
    - host: tcp-echo.example.com
      http:
        paths:
          - pathType: Prefix
            path: /
            backend:
              service:
                name: tcp-echo-service
                port:
                  number: 2701
```

PaleoCrafter:
If the TCP port is bound as SSL-enabled, i.e. terminating the SSL connection, `req.ssl_hello_type` is not set and the inspection delay would block the request unnecessarily.

The solution is to only perform the inspection and add the delay if we're passing through the SSL connection and still want to route based on SNI. The delay can be optional in this case as the `ssl_fc_sni` value is immediately available at that point.

### DIFF
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -2037,9 +2037,7 @@ frontend _front_tcp_7011
 frontend _front_tcp_7012
     bind :7012 ssl crt /ssl/7012.pem
     mode tcp
-    tcp-request inspect-delay 5s
     tcp-request content set-var(req.tcpback) ssl_fc_sni,lower,map_str(/etc/haproxy/maps/_tcp_sni_7012__exact.map)
-    tcp-request content accept if { req.ssl_hello_type 1 }
     use_backend %[var(req.tcpback)] if { var(req.tcpback) -m found }
 frontend _front_tcp_7013
     bind :7013
--- a/rootfs/etc/templates/haproxy/haproxy.tmpl
+++ b/rootfs/etc/templates/haproxy/haproxy.tmpl
@@ -1010,7 +1010,9 @@ frontend {{ $proxy_name }}
 
 {{- /*------------------------------------*/}}
 {{- if $tcpport.SNIMap.HasHost }}
+{{- if $tls.TLSFilename }}
     tcp-request inspect-delay 5s
+{{- end }}
 {{- range $match := $tcpport.SNIMap.MatchFiles }}
     tcp-request content set-var(req.tcpback) {{ if $tls.TLSFilename }}ssl_fc_sni{{ else }}req.ssl_sni{{ end }},lower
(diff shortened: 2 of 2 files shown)


jcmoraisjr:
Thank you for the contribution! Seeing the error in the unit test I realized that I've asked you to patch the template in the wrong way, we should use ` {{- if not $tls.TLSFilename }}` because the inspect should only be made if we do not have a crt file, i.e. we don't offload the encrypted connections. So my suggestion is to invert the logic here and also remove the inspect and hello-type from the corresponding config in the unit test.

PaleoCrafter:
Yep, I should've thought a little longer about what I was doing there :sweat_smile: I'll be amending the changes and fixing the unit tests momentarily!

jcmoraisjr:
LGTM thanks! Merging now.

Issue #823, opened 2021-07-13
Allow configuring Cache-Control header in autogenerated responses


Cynocracy:
**What are you trying to do**

Hi! First off, thanks for the solution, we're using it to great success!

In our setup, we're using CloudFlare cache on top of HAProxy -> backend, and we have browser caching enabled.
The issue, aiui, with sending no-cache, is that this doesn't impact the browser cache, which will treat the request as valid (and storeable).

**What HAProxy Ingress should do or how it should behave differently**
Given that a change from no-cache -> no-store could be backward incompatible, it would be nice to have a config option that allows for changing the cache control header on HAProxy generated responses (lua and non-lua ones even, using error_pages, would be even better!)

Happy to brainstorm about how to expose that option, but to me it seems like a per-frontend option (or global).
Also happy to help contribute the feat :)

Cynocracy:
We're successfully working around this by volume mounting an alternative services.lua, fwiw, but that feels fragile (we might not keep up with the status codes, or the interaction between files might change)

jcmoraisjr:
Hi, thanks for the fr and sorry about taking so long to comment. Yes, lets make it happen, I'm already planning to implement a simple way to override the default response data for all possible http status code we manage, which I think will somewhat help in this issue. I'll draw a bit more the solution as soon as I have more details, but should be something like a new configmap entry per status code, configured as a multi-line content with the raw http output. Does it seem to help on your use case?

Cynocracy:
That sounds like it would work for our use case, thank you!

jcmoraisjr:
Finally added to v0.14.0-alpha.2. Closing.

PR #915, opened 2022-04-05
Add http-response configuration keys

jcmoraisjr:
Add option to overwrite any response payload issued by haproxy, or configured by haproxy ingress. This update might lead to backward incompatibility if a deployment already customize the Lua script with the haproxy ingress generated responses.

We're using two approaches here. One of them is implementing services in Lua, this gives some flexibility to issue codes that haproxy doesn't support like 495. The other approach is using HAProxy errorfile, which is the proper way to overwrite responses generated internally by HAProxy and hard to move to a Lua script, like 500 or 503. All of them however uses the same interface, so users don't need to bother if the overwrite will be made via errorfile or Lua script.

The implementation was also made in such a way that a response code can be changed, this is useful in scenarios that the client should be redirected via `302 Found` if there isn't a server available (503) or if the request is denied (403).


jcmoraisjr:
#823

PR #927, opened 2022-06-20
Add cors-allow-origin-regex annotation

### LINKED ISSUE #853: Support wildcards in `ingress.kubernetes.io/cors-allow-origin`
**What are you trying to do**

I'm making use of `ingress.kubernetes.io/cors-allow-origin` across multiple-domains and it would handy if we could support something like:

`ingress.kubernetes.io/cors-allow-origin: "*.example.com, *.example2.com"`

I think perhaps the current option of `*` is not restrictive enough for my needs...but at the same time being able to use wildcards for different domains would save a lot of management overhead (we have lots of hosts...).

Related: https://github.com/jcmoraisjr/haproxy-ingress/issues/477

wolf-cosmose:
* add `haquote` function for quoting strings in haproxy config template
* add `cors-allow-origin-regex` annotaton
* don't allow empty origins in `cors-allow-origin` to avoid generating broken haproxy configs

This should fix #853 

The `cors-allow-origin-regex` annotation still requires some value in `cors-allow-origin` for sending when the origin doesn't match, as noted in the docs. Would be nice if the default wasn't `*`, which totally bypasses any of the regexes, but I guess that'd break backwards compatibility. Maybe we should have a warning for that?

I tried to update the docs, tests, etc. but I may've missed something, please let me know if I did.


pkg/converters/ingress/annotations/backend.go
```diff
@@ -515,16 +515,23 @@ func (c *updater) buildBackendCors(d *backData) {
...
+		allowOriginRegex := config.Get(ingtypes.BackCorsAllowOriginRegex).Value
+
 		if enabled {
 			path.Cors = hatypes.Cors{
 				Enabled:          enabled,
 				AllowCredentials: config.Get(ingtypes.BackCorsAllowCredentials).Bool(),
 				AllowHeaders:     config.Get(ingtypes.BackCorsAllowHeaders).Value,
 				AllowMethods:     config.Get(ingtypes.BackCorsAllowMethods).Value,
 				AllowOrigin:      strings.Split(config.Get(ingtypes.BackCorsAllowOrigin).Value, ","),
+				AllowOriginRegex: []string{},
```
jcmoraisjr, line 527: All the changes in this method can be simplified to something like

```golang
		...
		if enabled {
			var allowOriginRegex []string
			if regex := config.Get(ingtypes.BackCorsAllowOriginRegex).Value; regex != "" {
				allowOriginRegex = strings.Split(regex, " ")
			}
```

and then using that var straight in the struct declaration. nil or empty slices in golang is pretty much the same, you don't need to worry about its initialization.

pkg/converters/ingress/annotations/backend_test.go
```diff
@@ -1209,6 +1209,7 @@ const (
 )
 
 var corsDefaultOrigin = []string{"*"}
+var corsDefaultOriginRegex = []string{}
```
jcmoraisjr, line 1212: Now the default value is nil instead of an empty slice, you can remove this declaration and all the items where it's used.

pkg/converters/ingress/defaults.go
```diff
@@ -50,6 +50,7 @@ func createDefaults() map[string]string {
 		types.BackCorsAllowHeaders:       "DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization",
 		types.BackCorsAllowMethods:       "GET, PUT, POST, DELETE, PATCH, OPTIONS",
 		types.BackCorsAllowOrigin:        "*",
+		types.BackCorsAllowOriginRegex:   "",
```
jcmoraisjr, line 53: No need to initialize with zero value.

pkg/haproxy/template/funcmap.go
```diff
@@ -57,6 +58,11 @@ func createFuncMap() gotemplate.FuncMap {
 			}
 			return out
 		},
+		"haquote": func(s string) string {
+			// put in single quotes
+			// escape single quotes inside as ' "'" ' (without the spaces)
+			return "'" + strings.ReplaceAll(s, "'", "'\"'\"'") + "'"
```
jcmoraisjr, line 64: Maybe `'"'"'` with backquote around makes it less unclear?

rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -737,11 +737,14 @@ backend {{ $backend.ID }}
...
 {{- if and $cors.Enabled $cors.AllowOrigin }}
-{{- $dynamicOrigin := gt (len $cors.AllowOrigin) 1 }}
+{{- $dynamicOrigin := or $cors.AllowOriginRegex (gt (len $cors.AllowOrigin) 1) }}
 {{- if $dynamicOrigin }}
 {{- range $a1 := short 5 $cors.AllowOrigin }}
     acl cors_allow_origin{{ $i }} var(txn.hdr_origin{{ $i }}) -m str{{ range $a := $a1 }} {{ $a }}{{ end }}
 {{- end }}
+{{- range $a1 := short 5 $cors.AllowOriginRegex }}
+    acl cors_allow_origin{{ $i }} var(txn.hdr_origin{{ $i }}) -m reg{{ range $a := $a1 }} {{ $a |haquote }}{{ end }}
+{{- end }}
```
jcmoraisjr, line 747: Templating is tricky and breaks very easily, I'd love to see a unit test for this change.

Have a look at this snippet:

https://github.com/jcmoraisjr/haproxy-ingress/blob/5677e338eebef8de5bdde6dcbb3d51c55eb1acb3/pkg/haproxy/instance_test.go#L239-L267

There are some other similar snippets above it with cors tests. Take one of them as an example to create a new one just to cover the new option, even better if you provide at least 6 regex as input.

jcmoraisjr (commented):
Great addition, thanks! Following a few suggestions.

rootfs/etc/templates/haproxy/haproxy.tmpl
wolf-cosmose: Thanks for pointing me to the right place, added a few tests there.

pkg/haproxy/template/funcmap.go
wolf-cosmose: good idea, should be good now

pkg/converters/ingress/annotations/backend_test.go
wolf-cosmose: ok, removed

Issue #932, opened 2022-06-24
panic: interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *v1.ConfigMap


monkeymorgan:
**Description of the problem**

During a period where the API server was overloaded (but still functional) and after a leader election in etcd (possible red herring) a large number of instances (150 out of 175) started to panic / crash / fail readiness checks ( /healthz) / restart.

The key log message is : 

panic: interface conversion: interface {} is cache.DeletedFinalStateUnknown, not *v1.ConfigMap

and the key parts of the stacktrace are: 

/go/pkg/mod/k8s.io/client-go@v0.19.13/tools/cache/controller.go
/go/pkg/mod/k8s.io/client-go@v0.19.13/tools/cache/shared_informer.go:779
https://github.com/kubernetes/client-go/blob/v0.19.13/tools/cache/shared_informer.go#L779

It seems that processorListener deleteNotification returns cache.DeletedFinalStateUnknown which is unexpected and causes a panic / lock-up. 

In-flight requests are not completed, and the clients of the in-flight requests timeout.

The readiness probes fail to establish a connection : 

Readiness probe failed: Get "http://10.12.228.150:10253/healthz": dial tcp 10.12.228.150:10253: connect: connection refused

k8s then starts new pods to replace these ones. In our case, this made the load on the API server rocket, as 150 new proxies attempted to list 500+ resources (ingresses / endpoints / secrets / configmaps / pods). I think these namespaces may have 'old' configmaps too which are not cleaned up.

Curiously, only one out of many (similar) deployments collapsed. Others continued to work fine (and emit traces about slow api server syncs) despite the load on the API server rising to the point where it was 95% 5xxs.

**Expected behavior**

Total resilience in the face of all API server issues ! (I know, a lot to ask) 

To reduce the cascade effect, I wonder would it be possible to add the option for label selectors to more tightly control the lister-watchers (observers) ? e.g. in the case of configmap, and in our case we have a 1:1 relationship from proxy to pod deployment so we could ideally only watch pods of interest. This namespace has circa 1000 pods, and 1000s are not unusual.

**Steps to reproduce the problem**

This could be tricky ... I expect it's a race cond. But the only prior errors are:

client-go/leaderelection.go:325
error retrieving resource lock A-NAMESPACE/ingress-controller-leader-haproxy-ingress-redacted-prod: etcdserver: leader changed

1.  Load the api server 
[...]

PR #934, opened 2022-06-26
Check type assertion on all informers

jcmoraisjr:
Delete event from informers might return `cache.DeletedFinalStateUnknown` instead of the actual object if the object is removed from the api while the watch is closed, e.g. during an outage. Informers that need to check the state of a removed object must check the type, otherwise a panic would be issued. This update implements a safe typecast on the delete event of all informers that were missing it.


jcmoraisjr:
#932

PR #940, opened 2022-08-01
Documents the expected format for --configmap key

lafolle:
If this format is not followed following error is thrown:

`
2022/08/01 08:50:43 unable to parse static config: error reading global ConfigMap 'haproxy-ingress': an empty namespace may not be set when a resource name is provided
`

that actually is thrown by k8.

I'm not sure if it is a common knowledge to always use namespace when specifying a configmap. Ignore this if that is the case.
If not, then i guess the format can also be enforced at the time of parsing `--configmap` arg?

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -71,7 +71,7 @@ The following sections describe in a few more details about configuration strate
 
 ConfigMap key/value options are read in the following conditions:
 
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
It seems a fair addition, although it's already in the command-line doc as well. Lgtm.

Regarding the error, this is due to the new golanglint version that enforces v0.19 formatting. Is it possible to you to rebase this commit on top of master so we have a green pipeline?

jcmoraisjr:
Nevermind, fixed now, merging. Thanks!

Issue #939, opened 2022-07-27
Critical and High Vulnerabilities in 0.13.8


rohank2002:
Attached is CSV which contains all the information on the vulnerabilities found.

[haproxy_ingress_vulnerabilities.csv](https://github.com/jcmoraisjr/haproxy-ingress/files/9202984/haproxy_ingress_vulnerabilities.csv)

We probably need to build with the new version of alpine 3.16.1

**Steps to reproduce the problem**

https://artifacthub.io/packages/helm/haproxy-ingress/haproxy-ingress/0.13.8?modal=security-report

**Environment information**

HAProxy Ingress version: v0.13.8

jcmoraisjr:
Ack, thanks for the info. All of the exploitable vulnerabilities come from haproxy dependencies, and we use a pre-compiled version. Other than that v0.13's upstream is already eol so we cannot simply wait for a new compilation. Maybe we'll need to compile it ourselves. Currently the fastest way to get rid of the CVEs is to use an external haproxy deployment, doc [here](https://haproxy-ingress.github.io/docs/examples/external-haproxy/), and an image with updated dependencies.

jcmoraisjr:
Fun fact, haproxy 2.3.21 was released today, although eol, and all the exploitable components are up to date. I'll cherry pick some other fixes and tag a new version shortly.

rohank2002:
Thank you  @jcmoraisjr.

jcmoraisjr:
v0.13.9 released which fixes all the vulnerabilities according to [quay.io report](https://quay.io/repository/jcmoraisjr/haproxy-ingress?tab=tags).

Artifacthub however still [reports](https://artifacthub.io/packages/helm/haproxy-ingress/haproxy-ingress/0.13.9?modal=security-report) that the image is still vulnerable to CVE2022-1996, which sounds strange to me because, according to [this report](https://nvd.nist.gov/vuln/detail/CVE-2022-1996), 2.16 is the first fixed version of the v2 branch. This is an indirect dependency, hard to upgrade across branches, and we as users of the direct dependency can only wait for a new version that updates the indirect one - sometimes reporting or even sending pr upstream ourselves.

Anyway a few more steps regarding secure images were added to our pipeline and release process. The only issue from the Artifacthub report isn't exploitable from the controller, and it seems not to be vulnerable anymore. So I'm closing this issue.

jcmoraisjr:
[...]

PR #943, opened 2022-08-06
Update client-go from v0.23.8 to v0.23.9 and indirect dependencies

jcmoraisjr:
Makes a vanilla update of the k8s dependencies and also an update to the latest indirect dependencies in order to fix some security issues. The indirect dependencies update should be adjusted and cherry picked to v0.13.


jcmoraisjr:
#939

PR #945, opened 2022-08-16
Migrate to new versions / off deprecated packages.

mrueg:
@jcmoraisjr some small migrations to a newer version of the dependency for masterminds/sprig as well as removing the archived pkg/errors package.


jcmoraisjr:
lgtm, thanks! merging now.

Issue #938, opened 2022-07-13
ExternalName Service without ports no longer works


merusso:
**Description of the problem**

HAProxy Ingress no longer supports ExternalName Services that do not include a port. When an Ingress points to an ExternalName Service without a port, the backend is not configured, resulting in an HTTP 404 error. The HA Proxy Ingress logs include the following:

> W0713 17:17:03.966403       9 ingress.go:524] skipping backend config of ingress 'default/httpbin': port not found: '80'

In v0.7.6, this config was supported. Support was broken in some later version of HAProxy Ingress (v0.8?).

**Expected behavior**

Since ExternalName Services do not require a port, I expected this config to create a backend configured using the same port specified on the Ingress. From my observation, this was the behavior in v0.7.6.

**Steps to reproduce the problem**

1. Create an ExternalName Service without `ports`
2. Create an Ingress pointing to this Service
3. Attempt to hit the URL exposed by the Ingress, e.g. https://httpbin.mydomain

**Environment information**

HAProxy Ingress version: `v0.13.8`

Command-line options:

```
--configmap=system/haproxy-ingress
--ingress-class=haproxy
--sort-backends
--default-backend-service=system/haproxy-ingress-default-backend
--rate-limit-update=10
--watch-ingress-without-class=true
```

Global options:

```
balance-algorithm: roundrobin
config-frontend: |
  maxconn 20000
  capture request header Host len 32
  capture request header Cookie len 32
  capture request header User-Agent len 32
  http-request redirect scheme https if !{ ssl_fc }
forwardfor: ifmissing
healthz-port: "10253"
ssl-always-add-https: "true"
ssl-redirect: "false"
stats-port: "1936"
syslog-endpoint: localhost:514
timeout-server: 600s
timeout-server-fin: 600s
```

Ingress object:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: httpbin
spec:
  rules:
    - host: httpbin.mydomain
      http:
        paths:
          - backend:
              service:
                name: httpbin
                port:
                  number: 80
            path: /
            pathType: ImplementationSpecific
```

Non-working Service object:

```
apiVersion: v1
kind: Service
metadata:
  name: httpbin
spec:
  type: ExternalName
  externalName: httpbin.org
```

Working Service object:

```
apiVersion: v1
kind: Service
metadata:
  name: httpbin
spec:
  type: ExternalName
  externalName: httpbin.org
  ports:
[...]

PR #946, opened 2022-09-06
Add support for service external name without port

jcmoraisjr:
A backend server is configured with the port number declared in the
service. However, services of type ExternalName doesn't need to have
a port declared, and since the v0.8 refactor, HAProxy Ingress refuses
to configure backends for services without port.

This update checks if the service is of type ExternalName and if the
ingress referencing this service uses a numeric port number. If true,
the port number declared in the ingress will be used as the port to
connect to the backend.


jcmoraisjr:
#938

PR #954, opened 2022-10-26
Make WAF errors fail closed by default. Add waf-fail-closed=false to override

mac-chaffee:
Full issue is described here: https://github.com/haproxy/spoa-modsecurity/issues/3

**Note: This is a breaking change from 0.14 to 0.15**

This PR changes the default behavior (that we copied from the upstream spoa-modsecurity docs) of allowing requests through if modsecurity errors out or times out. Now, only a valid response from modsecurity (`txn.modsec.code=0`) will allow requests through.

Users can go back to the old behavior by setting `waf-fail-closed: false` which would mean that e.g. a flood of requests could overload modsecurity-spoa and trigger timeouts, which would allow requests through without being scanned by modsecurity.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -2856,7 +2856,7 @@ See also:
 | Configuration key | Scope  | Default | Since |
 |-------------------|--------|---------|-------|
 | `waf`             | `Path` |         |       |
-| `waf-fail-closed` | `Path` | `true`  | v0.15 |
+| `waf-fail-closed` | `Path` | `true`  | v0.14 |
 | `waf-mode`        | `Path` | `deny`  | v0.9  |
 
 
@@ -2866,7 +2866,7 @@ to validate requests. Currently the only supported value is `modsecurity`.
 This configuration has no effect if the ModSecurity endpoints are not configured.
 
 The `waf-fail-closed` key determines whether an error from the WAF (such as a timeout) should
-cause the request to be denied ("true") or allowed ("false"). The default value of "true" means that errors or timeouts will cause requests to be denied.
+cause the request to be denied ("true") or allowed ("false"). In v0.14, this defaults to "false" for backwards compatibility. In v0.15, this defaults to "true" which means that errors or timeouts will cause requests to be denied.
 
 The `waf-mode` key defines whether the WAF should be `deny` or `detect` for that Backend.
 If the WAF is in `detect` mode the requests are passed to ModSecurity and logged, but not denied.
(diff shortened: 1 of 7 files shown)


docs/content/en/docs/configuration/keys.md
```diff
@@ -2855,16 +2856,20 @@ See also:
 | Configuration key | Scope  | Default | Since |
 |-------------------|--------|---------|-------|
 | `waf`             | `Path` |         |       |
+| `waf-fail-closed` | `Path` | `true`  | v0.15 |
```
jcmoraisjr, line 2859: we'll cherry pick to v0.14 so you can update the starting version here.

docs/content/en/docs/configuration/keys.md
```diff
@@ -2855,16 +2856,20 @@ See also:
...
 | `waf-mode`        | `Path` | `deny`  | v0.9  |
 
+
 Defines which web application firewall (WAF) implementation should be used
 to validate requests. Currently the only supported value is `modsecurity`.
 
 This configuration has no effect if the ModSecurity endpoints are not configured.
 
+The `waf-fail-closed` key determines whether an error from the WAF (such as a timeout) should
+cause the request to be denied ("true") or allowed ("false"). The default value of "true" means that errors or timeouts will cause requests to be denied.
```
jcmoraisjr, line 2869: add something like "true since v0.15". I'll update this doc and the default value in v0.14 to false.

rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -582,7 +582,11 @@ backend {{ $backend.ID }}
 {{- range $i, $waf := $wafCfg.Items }}
 {{- if eq $waf.Mode "deny" }}
 {{- range $pathIDs := $wafCfg.PathIDs $i }}
+    {{- if $waf.FailClosed }}
+    http-request deny if !{ var(txn.modsec.code) -m int eq 0 }
+    {{- else }}
     http-request deny if { var(txn.modsec.code) -m int gt 0 }
+    {{- end }}
```
jcmoraisjr, line 589: please unindent these three `{{commands}}` removing the spaces on the left.

jcmoraisjr (commented):
Thanks for this PR! See a few updates below.

docs/content/en/docs/configuration/keys.md
mac-chaffee: I think I understand what you're saying: v0.14 will default to false, and v0.15 will default to true, presumably to make it easier for users to migrate. Does my updated sentence look okay now?

jcmoraisjr: Sure it's fine, thanks!

PR #957, opened 2022-11-02
Enable Load Server State feature for external haproxy

andrejbaran:
This should make the [Load Server State](https://haproxy-ingress.github.io/docs/configuration/keys/#load-server-state) feature work for setups where external haproxy is used.

### DIFF
--- a/pkg/haproxy/instance.go
+++ b/pkg/haproxy/instance.go
@@ -729,21 +729,23 @@ func (i *instance) waitWorker() error {
 }
 
 func (i *instance) retrieveServersState() (string, error) {
-	if state, err := i.conns.Admin().Send(nil, "show servers state"); err != nil {
+	state, err := i.conns.Admin().Send(nil, "show servers state")
+	if err != nil {
 		return "", fmt.Errorf("failed to retrieve servers state from external haproxy; %w", err)
-	} else {
-		return state[0], nil
 	}
+
+	return state[0], nil
 }
 
 func (i *instance) persistServersState() error {
-	if state, err := i.retrieveServersState(); err != nil {
+	state, err := i.retrieveServersState()
+	if err != nil {
 		return err
-	} else {
-		stateFilePath := filepath.Join(i.config.Global().LocalFSPrefix, "/var/lib/haproxy/state-global")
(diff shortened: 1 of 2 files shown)


andrejbaran:
First of all @jcmoraisjr huge thank you for starting and maintaining such a great project!

Apologies for dropping this without a prior issue but I figured that is's a fairly simple and small change. Let me know if there is anything I can do to make this happen.

I also have a backport ready for v0.13 (which we are using) (probably 0.12 too) branch since the socket logic has been reorganized for 0.14.

pkg/haproxy/instance.go
```diff
@@ -722,3 +727,24 @@ func (i *instance) waitWorker() error {
...
+		return state[0], nil
+	}
+}
+
+func (i *instance) persistServersState() error {
+	if state, err := i.retrieveServersState(); err != nil {
+		return err
+	} else {
+		stateFilePath := filepath.Join(i.config.Global().LocalFSPrefix, "/var/lib/haproxy/state-global")
+		if err := os.WriteFile(stateFilePath, []byte(state), os.ModePerm); err != nil {
```
andrejbaran, line 744:
```suggestion
		if err := os.WriteFile(stateFilePath, []byte(state), 0o666); err != nil {
```

jcmoraisjr:
Hi, thanks for supporting this controller! Please have a look in the lint complains, in short you can remove some of the `else` statements just after a return. Other than that please send the link to your v0.13 or v0.12 patches, couldn't find on your fork. No need to create another pr, just pushing them to your fork is enough. I'll use them to make the cherry pick to the older versions preserving the author of the commit.

andrejbaran:
Hey @jcmoraisjr,

Thanks for the quick response and apologies for the delay. Linter should be happy now.
Here's the patch for 0.13 https://github.com/ftshp/haproxy-ingress/commit/8fac467aaff1baeb6a241926619890df0d1539a0

jcmoraisjr:
Thanks! Lgtm. Just a qq, wondering if `persistServerState()` in v0.13 should be inside `if LoadServerState` as well?

jcmoraisjr:
Great, thanks! Merging now.

PR #933, opened 2022-06-25
Add controller-runtime support

jcmoraisjr:
This update adds support to Kubernetes' controller-runtime. v0.15 will start the controller with controller-runtime by default, but can be changed to the legacy runtime using command-line option. v0.16 will drop the legacy runtime support.

All the current runtime features will be migrated, even if in an incompatible way - and if so, properly documented in the release notes.

Resource parser and configuration builder were marginally changed, just to fit a few interface adjusts, making this upgrade with a low risk to generate hidden bugs and compatibility changes.

New architecture at a glance is:

- `pkg/controller` - continues with the runtime configuration
- `pkg/controller/launch` - bootstraps all the configurations and services needed to start the runtime manager
- `pkg/controller/config` - static, command-line based config options
- `pkg/controller/services` - configures, starts and manages all the internal services
- `pkg/controller/reconciler` - watch, filter, group and dispatch notification about resource changes
- `pkg/controller/legacy` - legacy runtime configuration
- `pkg/common` - legacy runtime component

pkg/controller/legacy and pkg/common will be removed in v0.16.

TODO:

* [x] design
* [x] config parser
* [x] core services
* [x] status updater service
* [x] acme services
* [x] converter interface
* [x] instance interface
* [x] cache interface
* [x] resource watcher
* [x] resource handlers
* [x] resource filters
* [x] logging parameters
* [x] listening probes, metrics, profiling
* [x] exploratory tests
* [x] update related doc

PR #959, opened 2022-11-10
Adds support for EndpointSlices API in master

lafolle:
Adds support for endpoint slices api to fetch endpoints.

Corresponding clusterrole changes https://github.com/haproxy-ingress/charts/pull/66

### DIFF
--- a/pkg/common/ingress/controller/launch.go
+++ b/pkg/common/ingress/controller/launch.go
@@ -296,6 +296,11 @@ Actually node listing isn't needed and it is always disabled`)
 			`DEPRECATED, this option is ignored. Use --watch-ingress-without-class
 command-line option instead to define if ingress without class should be
 tracked.`)
+
+		enableEndpointSlicesAPI = flags.Bool("enable-endpointslices-api", false,
+			`Enables EndpointSlices API and disables watching Endpoints API. Only enable
+			in
+k8s >=1.21+`)
 	)
 
 	logLevel := new(klog.Level)
@@ -378,6 +383,10 @@ tracked.`)
 		klog.Info("running embedded haproxy, mode is master-worker")
 	} else {
 		klog.Info("running embedded haproxy, mode is daemon")
+	}
+
+	if *enableEndpointSlicesAPI {
+		klog.Infof("watching endpointslices - --enable-endpointslices-api is true")
 	}
 
 	if !(*reloadStrategy == "native" || *reloadStrategy == "reusesocket" || *reloadStrategy == "multibinder") {
@@ -575,6 +584,7 @@ tracked.`)
 		BackendShards:            *backendShards,
 		SortEndpointsBy:          sortEndpoints,
 		UseNodeInternalIP:        *useNodeInternalIP,
+		EnableEndpointSlicesAPI:  *enableEndpointSlicesAPI,
 	}
 
 	ic := newIngressController(config)
--- a/pkg/controller/legacy/cache.go
+++ b/pkg/controller/legacy/cache.go
@@ -33,6 +33,7 @@ import (
 	"time"
 
 	api "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	networking "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -143,8 +144,18 @@ func createCache(
 		cfg.ForceNamespaceIsolation,
 		!cfg.DisablePodList,
 		cfg.ResyncPeriod,
+		cfg.EnableEndpointSlicesAPI,
 	)
 	return cache
+}
+
+func (c *k8scache) GetEndpointSlices(service *api.Service) ([]*discoveryv1.EndpointSlice, error) {
+	serviceNameLabel := map[string]string{"kubernetes.io/service-name": service.Name}
+	selector, err := buildLabelSelector(serviceNameLabel)
+	if err != nil {
+		return nil, err
+	}
+	return c.listers.endpointSliceLister.EndpointSlices(service.Namespace).List(selector)
 }
 
 func (c *k8scache) RunAsync(stopCh <-chan struct{}) {
@@ -840,6 +851,8 @@ func (c *k8scache) Notify(old, cur interface{}) {
 			ch.NeedFullSync = true
 		case *api.Endpoints:
 			ch.EndpointsNew = append(ch.EndpointsNew, cur)
+		case *discoveryv1.EndpointSlice:
+			ch.EndpointSlicesUpd = append(ch.EndpointSlicesUpd, cur)
 		case *api.Service:
 			svc := cur
 			if old == nil {
@@ -961,6 +974,9 @@ func (c *k8scache) SwapChangedObjects() *convtypes.ChangedObjects {
 	}
 	for _, ep := range ch.EndpointsNew {
 		addChanges(convtypes.ResourceEndpoints, eventUpdate, ep.Namespace, ep.Name)
+	}
+	for _, eps := range ch.EndpointSlicesUpd {
+		addChanges(convtypes.ResourceEndpoints, eventUpdate, eps.Namespace, eps.Labels["kubernetes.io/service-name"])
 	}
 	for _, svc := range ch.ServicesDel {
 		addChanges(convtypes.ResourceService, eventDel, svc.Namespace, svc.Name)
--- a/pkg/converters/configmap/tcpservices.go
+++ b/pkg/converters/configmap/tcpservices.go
@@ -36,18 +36,20 @@ type TCPServicesConverter interface {
 // NewTCPServicesConverter ...
 func NewTCPServicesConverter(options *convtypes.ConverterOptions, haproxy haproxy.Config, changed *convtypes.ChangedObjects) TCPServicesConverter {
 	return &tcpSvcConverter{
-		logger:  options.Logger,
-		cache:   options.Cache,
-		haproxy: haproxy,
-		changed: changed,
+		logger:                  options.Logger,
+		cache:                   options.Cache,
+		haproxy:                 haproxy,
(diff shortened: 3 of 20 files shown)


monkeymorgan:
Note that @lafolle is not currently working on this (The 'in progress' comment might make you think there's more to come). He's away for the foreseeable, but I can try and answer any follow ups.

jcmoraisjr:
Thanks for the update. I'll do my best to add it in the first v0.15 snapshot and I'll also try to add it in the first v0.14 stable or maybe in the first patch of that branch.

pkg/common/ingress/controller/launch.go
```diff
@@ -296,6 +296,10 @@ Actually node listing isn't needed and it is always disabled`)
 			`DEPRECATED, this option is ignored. Use --watch-ingress-without-class
 command-line option instead to define if ingress without class should be
 tracked.`)
+
+		enableEndpointSlicesAPI = flags.Bool("enable-endpointslices-api", true,
+			`Enables EndpointSlices API and disables watching Endpoints API. Only enable
+			in k8s >=1.21+`)
```
jcmoraisjr, line 302: Make this defaults to `false`.

pkg/controller/listers.go
```diff
@@ -205,14 +217,29 @@ func (l *listers) RunAsync(stopCh <-chan struct{}) {
...
+			l.ingressInformer.HasSynced,
+			l.endpointSliceInformer.HasSynced,
+			l.serviceInformer.HasSynced,
+			l.secretInformer.HasSynced,
+			l.configMapInformer.HasSynced,
+			l.podInformer.HasSynced,
+		)
+
+	} else {
+		go l.endpointInformer.Run(stopCh)
```
jcmoraisjr, line 233: Remove the endpointInformer.Run() outside this if/else, which should panic if endpointslices is enabled.

pkg/converters/configmap/tcpservices.go
```diff
@@ -44,10 +44,11 @@ func NewTCPServicesConverter(options *convtypes.ConverterOptions, haproxy haprox
...
 type tcpSvcConverter struct {
-	logger  types.Logger
-	cache   convtypes.Cache
-	haproxy haproxy.Config
-	changed *convtypes.ChangedObjects
+	logger               types.Logger
+	cache                convtypes.Cache
+	haproxy              haproxy.Config
+	changed              *convtypes.ChangedObjects
+	useEndpointSlicesAPI bool
```
jcmoraisjr, line 51: Missing initialization in the New...() constructor above.

pkg/converters/types/interfaces.go
```diff
@@ -99,11 +103,12 @@ const (
...
 
-	ResourceConfigMap ResourceType = "ConfigMap"
-	ResourceService   ResourceType = "Service"
-	ResourceEndpoints ResourceType = "Endpoints"
-	ResourceSecret    ResourceType = "Secret"
-	ResourcePod       ResourceType = "Pod"
+	ResourceConfigMap      ResourceType = "ConfigMap"
+	ResourceService        ResourceType = "Service"
+	ResourceEndpoints      ResourceType = "Endpoints"
+	ResourceEndpointSlices ResourceType = "EndpointSlices"
```
jcmoraisjr, line 109: I think you don't need this. Let me describe in a nutshell how tracking currently works:

- resource informer (`listers.go`) notifies the cache that a resource changed (create/update/delete)
- cache adds the resource identifier (ns+name) in an array of changed objects
- converter reads that array in order to find what changed

on the other side:

- object references (ing references svc, svc ref. endpoint, ing ref. secret, etc.) are tracked via `tracker.RefName()` (see below)
- `tracker.QueryLinks()` knows who should be synced again due to a resource update (see below)

`tracker.RefName()`:

https://github.com/jcmoraisjr/haproxy-ingress/blob/b87f94b9149bad3cef902e7e058c62ad2d83e02e/pkg/converters/ingress/ingress.go#L780-L783

`tracker.QueryLinks()`:

https://github.com/jcmoraisjr/haproxy-ingress/blob/b87f94b9149bad3cef902e7e058c62ad2d83e02e/pkg/converters/ingress/ingress.go#L219-L226

That said, what we need to do is to create a relation between svc and ep, so svc (hence, its ingress) will be synced when ep changes. The problem with endpointslices is that you cannot track the resource name because controller manager can create new endpointslices with unpredictable names.

My idea here is, instead, reuse all the former endpoint types - resource type and array of changed resources - and update it whenever an endpointslice is changed. Note that, instead of using its name, use it's annotation, in order to have a predictable name, btw the same name pattern the former ep already uses (it matches the svc name). You can test this by simply scaling up (provided you have free slots) or down, you should see that the current approach isn't changing backend servers, but the former ep does, and that proposed refactor should do it as well.

pkg/converters/types/options.go
```diff
@@ -22,25 +22,26 @@ import (
...
+	DefaultBackend          string
+	DefaultCrtSecret        string
+	FakeCrtFile             CrtFile
+	FakeCAFile              CrtFile
+	AnnotationPrefix        []string
+	DisableKeywords         []string
+	AcmeTrackTLSAnn         bool
+	TrackInstances          bool
+	HasGateway              bool
+	EnableEndpointSlicesAPI bool
```
jcmoraisjr, line 44: This wasn't initialized, here:

https://github.com/jcmoraisjr/haproxy-ingress/blob/b87f94b9149bad3cef902e7e058c62ad2d83e02e/pkg/controller/controller.go#L156-L175

jcmoraisjr (commented):
Thank you for the contribution! Sorry about taking so much time to revise it.

See a few comments below.

jcmoraisjr:
Folks, needed to merge #933, sorry. You can see some conflicts but all of them should be pretty simple to fix. The main issue however is that all the lister, informer and cache code base need to be written for the new controller, but you don't need to worry about that, I can take care of that if you prefer. Just run the controller with `HAPROXY_INGRESS_RUNTIME=LEGACY`, or adjust the code below, to make your local test.

https://github.com/jcmoraisjr/haproxy-ingress/blob/df00660d95445eb92dda9a34ac1e83f47091fee0/pkg/main.go#L32-L36

pkg/converters/types/interfaces.go
lafolle: Thanks for the explanation Joao. I've updated to reuse endpoint datastructures.

lafolle:
Linter seems to be failing with new golangci-lint 1.52.2 while it [passes](https://github.com/jcmoraisjr/haproxy-ingress/actions/runs/4338829538/jobs/7575853755?pr=995) for 1.51.2. Two options: 1) fix haproxy-ingress to satisfy the new linter 2) pin golangci-lint to 1.51.x.

pkg/converters/configmap/tcpservices.go
lafolle: done

pkg/controller/legacy/cache.go
```diff
@@ -962,6 +975,9 @@ func (c *k8scache) SwapChangedObjects() *convtypes.ChangedObjects {
 	for _, ep := range ch.EndpointsNew {
 		addChanges(convtypes.ResourceEndpoints, eventUpdate, ep.Namespace, ep.Name)
 	}
+	for _, eps := range ch.EndpointSlicesUpd {
+		addChanges(convtypes.ResourceEndpoints, eventUpdate, eps.Namespace, eps.Annotations["kubernetes.io/service-name"])
+	}
```
jcmoraisjr, line 980: This should be `eps.Labels`, sorry about my previous comment when I suggested annotation instead.

pkg/converters/types/options.go
jcmoraisjr, line 44: Use `EnableEPSlices` instead, this helps to avoid changing all the lines of the struct declaration, which in turn helps to avoid conflict when cherry-picking fixes and improvements between branches.

pkg/converters/utils/services.go
```diff
@@ -73,27 +73,92 @@ type Endpoint struct {
 }
 
 // CreateEndpoints ...
-func CreateEndpoints(cache types.Cache, svc *api.Service, svcPort *api.ServicePort) (ready, notReady []*Endpoint, err error) {
+func CreateEndpoints(cache types.Cache, svc *api.Service, svcPort *api.ServicePort, useEndpointSlices bool) (ready, notReady []*Endpoint, err error) {
```
jcmoraisjr, line 76: This func seems more like a builder/factory now, consider the snippet below as a way to split specific behavior to specific new funcs.

```go
	switch {
	case svc.Spec.Type == api.ServiceTypeExternalName:
		ready, err := createEndpointsExternalName(cache, svc, svcPort)
		return ready, nil, err
	case useEndpointSlices:
		endpoints, err := cache.GetEndpointSlices(svc)
		if err != nil {
			return nil, nil, err
		}
		return createEndpointSlices(endpoints, svcPort)
	default:
		endpoints, err := cache.GetEndpoints(svc)
		if err != nil {
			return nil, nil, err
		}
		return createEndpoints(endpoints, svcPort)
	}
```

```go
func createEndpoints(endpoints *api.Endpoints, svcPort *api.ServicePort) (ready, notReady []*Endpoint, err error)
```

```go
func createEndpointSlices(endpointSlices []*discoveryv1.EndpointSlice, svcPort *api.ServicePort) (ready, notReady []*Endpoint, err error)
```

jcmoraisjr (commented):
Thank you! See below a few more suggestions, and also:

- please remove `revive` from `.golangci.yaml`, this should fix the lint errors. I'll revisit the lint config and also the errors in a future pr;
- add the new option in `command-line.md` doc file;
- try to squash all the commits into a single one.

After that I'm done and we can merge. I'll prepare a cherry-pick for v0.14 later, as well as the v0.15 controller implementation. Thanks for all the effort put in this controller, that's very much appreciated.

jcmoraisjr:
LGTM thanks! Merging now, v0.14 cherry-pick and v0.15 controller implementation coming soon.

PR #960, opened 2022-11-11
Adds support for EndpointSlices API in 0.13 release

lafolle:
_In progress_


monkeymorgan:
Very interested in the timelines for this @jcmoraisjr - is this on your roadmap/radar for 0.13 or 0.14 ?
I see the https://github.com/jcmoraisjr/haproxy-ingress/pull/959 for master. 
We're currently running many instances of hpx-ingress with pod counts well in excess of 1000 and would like to update to k8s 1.22 soon at which point this becomes a real necessity.

Full disclosure: I work with the author of these two drafts / we have similar code patched in. But obviously would like to get back to using a released version. Without putting pressure on (since this is open source after all!) we'd be curious to know roughly where it lands in the priorities.

jcmoraisjr:
@lafolle @monkeymorgan first of all a big thank you for choosing, testing, using, patching and sharing, this is of an invaluable help.

@monkeymorgan in the next few days I'd like to see #959 reviewed and merged, meaning that it'll be part of v0.15, which is the current version in master. I don't have any problem to see it merged to v0.14 as well. I did a quick look on the changes and it seems good overall, probably have just a small suggestions as soon as it gets ready for review. Please let me know you all if you need any help to fix some of the tests.

Regarding v0.13 I'm a bit worried about updating. Although it is a flagged option, so it's an opt in feature, it changes quite a few parts of the code, from listers to converters, in order to make it happen. It's not a strong opinion yet but tending not to merge atm - v0.13 unfortunately has a history of flakes and regressions, it was somewhat hard to stabilize it. Maybe you folks can have a look on v0.14 and see how it behaves on your env? That'd be great for us.

jcmoraisjr:
Folks we cannot add this feature to v0.13 unfortunately, but we'll however do it on v0.14. #959 already takes care of that. Closing this one.

PR #961, opened 2022-11-14
Improve dashboard compatibility and cpu display

mac-chaffee:
* Changed the panel for displaying CPU usage to specifically select haproxy. The old panel would show CPU usage for every container in the cluster without the extra label I added.
* The `cluster` and `hostname` labels seem to be specific to other people's Prometheus instances since I don't have those labels. As a result, I rely on the "allValue" to still see stuff in the dashboard, but the old value of `.+` for hostname meant that nothing would appear even if I did select "All". Changing it to `.*` to match the other "allValue"s means the hostname is no longer required.


jcmoraisjr:
Hi, couldn't find your commit anymore, so I recreated it on #1223. This change is already applied on Grafana dashboards as well. Thank you, and sorry about the long delay on applying it.

PR #963, opened 2022-11-16
Revert "Merge pull request #954 from mac-chaffee/fail-closed"

### LINKED ISSUE #962: waf-fail-closed=true rejects all requests

See https://github.com/haproxy/spoa-modsecurity/issues/3#issuecomment-1316092212

While testing, I discovered that this comment is not entirely true:
https://github.com/haproxy/spoa-modsecurity/blob/3c895f3e7dd291dba19d57ba054b277e6fb80ca4/spoa.c#L93

When spoa-modsecurity approves a request, it actually sets `txn.modsec.code` to -1, meaning my changes in #954 aren't correct.

But there's a bigger problem: you can't actually tell the difference between spoa-modsecurity throwing certain errors and approving a request since the response is -1 either way.

**Expected behavior**

waf-fail-closed=true still lets in valid requests.

This might end up being hard to fix since we might need the upstream spoa-modsecurity to change the behavior of the return code. Maybe just a revert of #954 is the best bet for now. Thoughts?

**Environment information**

HAProxy Ingress version: master branch

mac-chaffee:
This reverts commit ff2953bdb6346e944a42e4fbd1e87ef727c8433d, reversing changes made to 316b7ed13d28d7f5836450ad93287712cf0acf47.

Fixes #962 

Long-term fix likely involves changes to how the upstream spoa-modsecurity sets the return code.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -503,7 +503,6 @@ The table below describes all supported configuration keys.
 | [`username`](#security)                              | haproxy user name                       | Global  | `haproxy`          |
 | [`var-namespace`](#var-namespace)                    | [true\|false]                           | Host    | `false`            |
 | [`waf`](#waf)                                        | "modsecurity"                           | Path    |                    |
-| [`waf-fail-closed`](#waf)                            | [true\|false]                           | Path    | `true` |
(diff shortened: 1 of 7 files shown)


jcmoraisjr:
Lgtm, thanks for that. Further comments on #962

PR #964, opened 2022-11-16
Coraza support

mac-chaffee:
This PR adds a new config value, `modsecurity-use-coraza`, which changes the WAF/SPOE configuration to support https://github.com/corazawaf/coraza-spoa

Specifically, this config matches the "experimental" branch of coraza-spoa. Slight changes in the args may be required in the future as the API stabilizes, so we do not provide default args in the source code (only in the docs).

To use, just set these values in the helm chart:

```
controller:
  image:
    repository: <must use an image based on this branch>
    tag: <must use an image based on this branch>
  config:
    modsecurity-use-coraza: "true"
    modsecurity-endpoints: 127.0.0.1:12345
    modsecurity-args: "app=hdr(host) id=unique-id src-ip=src src-port=src_port dst-ip=dst dst-port=dst_port method=method path=path query=query version=req.ver headers=req.hdrs body=req.body"
    waf: modsecurity
  extraVolumes:
    - name: coraza-config
      configMap:
        name: coraza-config
  extraContainers:
    - name: coraza-spoa
      image: <you must supply your own image based on the "experimental" branch with this PR applied https://github.com/corazawaf/coraza-spoa/pull/36 >
      ports:
      - containerPort: 12345
        name: spop
        protocol: TCP
      resources:
        limits:
          memory: 500Mi
        requests:
          cpu: 1000m
          memory: 250Mi
      volumeMounts:
      - name: coraza-config
        mountPath: /config.yaml
        subPath: config.yaml
        readOnly: true
      livenessProbe:
        tcpSocket:
          port: 12345
        failureThreshold: 3
        initialDelaySeconds: 30
        periodSeconds: 5
        successThreshold: 1
        timeoutSeconds: 4
```

The above values will mount a ConfigMap for Coraza's config.yaml, which must be created separately. Here is that configmap:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: coraza-config
  namespace: haproxy-ingress
data:
  config.yaml: |
    bind: 127.0.0.1:12345
    default_application: default_app
    applications:
      default_app:
        include:
          - /etc/coraza-spoa/coraza.conf
          - /etc/coraza-spoa/crs-setup.conf
          - /etc/coraza-spoa/rules/*.conf

        transaction_ttl: 60000
        transaction_active_limit: 100000
        log_level: debug
        log_file: /dev/stdout

```
Note the `default_application` line, that is from this PR: https://github.com/corazawaf/coraza-spoa/pull/36
That PR is needed because I set the `app` arg to be the `Host` header, so you need a default_application to configure WAF rules for all hosts. Otherwise, you'd need a different config block for every single Ingress hostname in your cluster.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -415,6 +415,7 @@ The table below describes all supported configuration keys.
 | [`modsecurity-timeout-hello`](#modsecurity)          | time with suffix                        | Global  | `100ms`            |
 | [`modsecurity-timeout-idle`](#modsecurity)           | time with suffix                        | Global  | `30s`              |
 | [`modsecurity-timeout-processing`](#modsecurity)     | time with suffix                        | Global  | `1s`               |
+| [`modsecurity-use-coraza`](#modsecurity)             | [true\|false]                           | Global  | `false`               |
 | [`nbproc-ssl`](#nbproc)                              | number of process                       | Global  | `0`                |
 | [`nbthread`](#nbthread)                              | number of threads                       | Global  |                    |
 | [`no-tls-redirect-locations`](#ssl-redirect)         | comma-separated list of URIs            | Global  | `/.well-known/acme-challenge` |
@@ -1997,6 +1998,7 @@ See also:
 | `modsecurity-timeout-idle`       | `Global` | `30s`   |       |
 | `modsecurity-timeout-processing` | `Global` | `1s`    |       |
 | `modsecurity-timeout-server`     | `Global` | `5s`    | v0.10 |
+| `modsecurity-use-coraza`         | `Global` | `false` | v0.14 |
 
 
 Configure modsecurity agent. These options only have effect if `modsecurity-endpoints`
@@ -2019,10 +2021,11 @@ The following keys are supported:
 * `modsecurity-timeout-idle`: Defines the maximum time to wait before close an idle connection. Default value is `30s`.
 * `modsecurity-timeout-processing`: Defines the maximum time to wait for the whole ModSecurity processing. Default value is `1s`.
 * `modsecurity-timeout-server`: Defines the maximum time to wait for an agent response. Configures the haproxy's timeout server. Defaults to `5s` if not configured.
+* `modsecurity-use-coraza`: Defines whether the generated SPOE config should include Coraza-specific values. In order to use Coraza instead of Modsecurity, you must set this to "true" and also set `modsecurity-args` based on the instructions in the [coraza-spoa repository](https://github.com/corazawaf/coraza-spoa). A full example can be found [here]({{% relref "../examples/modsecurity#using-coraza-instead-of-modsecurity" %}}).
 
 See also:
 
-* [example]({{% relref "../examples/modsecurity" %}}) page.
+* [modsecurity example]({{% relref "../examples/modsecurity" %}}) page.
 * [`waf`](#waf) configuration key.
 * https://www.haproxy.org/download/2.0/doc/SPOE.txt
 * https://docs.haproxy.org/2.4/configuration.html#9.3
@@ -2858,7 +2861,7 @@ See also:
 | `waf-mode`        | `Path` | `deny`  | v0.9  |
 
 Defines which web application firewall (WAF) implementation should be used
-to validate requests. Currently the only supported value is `modsecurity`.
+to validate requests. Currently the only supported value is `modsecurity`, which also supports Coraza endpoints when `modsecurity-use-coraza` is set to "true".
 
 This configuration has no effect if the ModSecurity endpoints are not configured.
 
--- a/docs/content/en/docs/examples/modsecurity.md
+++ b/docs/content/en/docs/examples/modsecurity.md
@@ -43,8 +43,8 @@ Check if the agent is up and running:
 
 ```
 $ kubectl -n ingress-controller get deployment modsecurity-spoa
-NAME                     READY     UP-TO-DATE   AVAILABLE  AGE  
-modsecurity-spoa         3/3       3            3          7s   
+NAME                     READY     UP-TO-DATE   AVAILABLE  AGE
+modsecurity-spoa         3/3       3            3          7s
 ```
 
 
@@ -186,3 +186,28 @@ modsecurity-spoa-6596c6b444-cht27   2/2     Running   0          14m
 modsecurity-spoa-6596c6b444-kw2tr   2/2     Running   0          14m
 modsecurity-spoa-6596c6b444-mkndw   2/2     Running   0          14m
 ```
+
+## Using Coraza instead of ModSecurity
+
+Since the maintainers of ModSecurity are dropping support in 2024, [OWASP has created a replacement called Coraza](https://coreruleset.org/20211222/talking-about-modsecurity-and-the-new-coraza-waf/) which is a drop-in replacement for ModSecurity.
+
+In order to use Coraza, the process is essentially the same as described in the above sections, with a few exceptions. First, in the haproxy-ingress ConfigMap, add the following two additional keys:
+
+```yaml
+  modsecurity-use-coraza: true
+  modsecurity-args: "app=hdr(host) id=unique-id src-ip=src src-port=src_port dst-ip=dst dst-port=dst_port method=method path=path query=query version=req.ver headers=req.hdrs body=req.body"
+```
+
+You may require different `modsecurity-args` depending on your Coraza config and your version of coraza-spoa. Check the [coraza-spoa README](https://github.com/corazawaf/coraza-spoa) for full details.
+
+Second, you'll need to change the spoa-modsecurity container image to a coraza-spoa image and create a configmap to hold the Coraza config.yaml. A complete example with all the changes can be found [here](/resources/coraza-deployment.yaml).
+
+{{% alert title="Warning" color="warning" %}}
+The coraza-spoa image that we provide in the above example is based on [an experimental branch of coraza-spoa](https://github.com/corazawaf/coraza-spoa/pull/36). For production environments, it would be best to wait until the experimental changes are merged and [an official image is released](https://github.com/corazawaf/coraza-spoa/issues/37).
+{{% /alert %}}
+
+### Troubleshooting Coraza
+
+* Ensure that the `bind` port in Coraza's config.yaml matches the port in `modsecurity-endpoints`.
+* Pay close attention to the `modsecurity-args` (specifically the `app` arg) and make sure they are set according to the coraza-spoa docs since the exact args may vary depending on your version of coraza-spoa.
+* For more complex issues, it may help to set `log_level: debug` in Coraza.
--- a/docs/static/resources/coraza-deployment.yaml
+++ b/docs/static/resources/coraza-deployment.yaml
@@ -1,0 +1,73 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    run: coraza-spoa
+  name: coraza-spoa
+  namespace: ingress-controller
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      run: coraza-spoa
+  template:
+    metadata:
+      labels:
+        run: coraza-spoa
+    spec:
+      containers:
+      - name: coraza-spoa
+        # NOTE: Built based on this PR: https://github.com/corazawaf/coraza-spoa/pull/36
+        # An official coraza-spoa image will be released by the upstream project soon.
+        image: quay.io/jcmoraisjr/coraza-spoa:experimental
+        ports:
+        - containerPort: 12345
+          name: spop
+          protocol: TCP
+        resources:
+          limits:
+            cpu: 200m
+            memory: 150Mi
+          requests:
+            cpu: 200m
+            memory: 150Mi
+        livenessProbe:
+          failureThreshold: 3
+          initialDelaySeconds: 30
+          periodSeconds: 5
+          successThreshold: 1
+          tcpSocket:
+            port: 12345
+          timeoutSeconds: 4
+      volumeMounts:
+      - name: coraza-config
+        mountPath: /config.yaml
+        subPath: config.yaml
+        readOnly: true
+    volumes:
+    - name: coraza-config
+      configMap:
+        name: coraza-config
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coraza-config
+  namespace: ingress-controller
+data:
+  # Check the official documentation: https://github.com/corazawaf/coraza-spoa
+  config.yaml: |
+    bind: :12345
+    default_application: default_app
+    applications:
+      default_app:
+        include:
+          - /etc/coraza-spoa/coraza.conf
+          - /etc/coraza-spoa/crs-setup.conf
+          - /etc/coraza-spoa/rules/*.conf
+
+        transaction_ttl: 60000
+        transaction_active_limit: 100000
+        log_level: info
+        log_file: /dev/stdout
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -4801,6 +4801,8 @@ func TestModSecurity(t *testing.T) {
 		modsecExp       string
 		modsecAgentArgs []string
 		modsecAgentExp  string
+		modsecUseCoraza bool
+		modsecOtherExp  string
 	}{
 		{
 			waf:        "modsecurity",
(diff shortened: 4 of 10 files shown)


jcmoraisjr:
Great job. I still need to have a closer look and want to run it myself on a local env, but for now I have two suggestions:

* please rebase on top of master since I saw you built this pr on top of your previous one, which is already merged;
* do not change where defaults is stored along with all the other changes, but instead let's postpone this move to another pr (btw why did you change?)

sealneaward:
@jcmoraisjr I can help with some testing too.

mac-chaffee:
Ah sorry forgot to mention the defaults.go thing. If I move the file back to where it was, I get an import cycle error:

```
$ go test -tags cgo ./...
# github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy
package github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy
	imports github.com/jcmoraisjr/haproxy-ingress/pkg/converters/ingress: import cycle not allowed in test
```

The cycle is "instance_test.go" imports "pkg/converters/ingress" and "pkg/converters/ingress/ingress.go" imports "pkg/haproxy", which is where instance_test.go is.

Seems either defaults.go needs to move or instance_test.go needs to move so neither gets accidentally imported when the parent folder is imported. Open to other ideas though!

sealneaward:
Ok, so I set this up on my cluster, and verified the configuration suggested works. I am able to intercept attacks and read them in the coraza sidecar container.

```
Loading 1 applications
time="2022-11-21T19:21:07Z" level=info msg="spoe: listening on [::]:9000"
{"level":"error","ts":1669058760.188159,"msg":"[client \"10.100.0.172\"] Coraza: Warning. OS File Access Attempt [file \"/etc/coraza-spoa/rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf\"] [line \"0\"] [id \"930120\"] [rev \"\"] [msg \"OS File Access Attempt\"] [data \"Matched Data: etc/passwd found within ARGS:p: /etc/passwd\"] [severity \"critical\"] [ver \"OWASP_CRS/4.0.0-rc1\"] [maturity \"0\"] [accuracy \"0\"] [tag \"application-multi\"] [tag \"language-multi\"] [tag \"platform-multi\"] [tag \"attack-lfi\"] [tag \"paranoia-level/1\"] [tag \"OWASP_CRS\"] [tag \"capec/1000/255/153/126\"] [tag \"PCI/6.5.4\"] [hostname \"\"] [uri \"/?p=/etc/passwd\"] [unique_id \"260e3d15-dd04-41b7-a634-45d17f0a8390\"]\n"}
```

One thing I noticed is that it is not passing the hostname to the logs, but I don't know if that needs to be addressed in the SPOA code vs the ingress controller code.
Otherwise, looks good. This is very exciting.

mac-chaffee:
@sealneaward ah good catch, I will investigate that

docs/content/en/docs/configuration/keys.md
```diff
@@ -2852,10 +2852,11 @@ See also:
...
 
-| Configuration key | Scope  | Default | Since |
-|-------------------|--------|---------|-------|
-| `waf`             | `Path` |         |       |
-| `waf-mode`        | `Path` | `deny`  | v0.9  |
+| Configuration key | Scope    | Default | Since |
+|-------------------|----------|---------|-------|
+| `waf`             | `Path`   |         |       |
+| `waf-mode`        | `Path`   | `deny`  | v0.9  |
+| `use-coraza`      | `Global` | `false` | v0.14 |
```
jcmoraisjr, line 2859: This seems a bit misplaced here, leading to conflict with every new waf implementation. What about renaming to `modsecurity-use-coraza` and moving to the modsecurity group? A few words here talking about Coraza should be enough, e.g. something like "Currently the only supported value is `modsecurity`, which also supports Coraza endpoints.".

pkg/converters/ingress/types/defaults.go
```diff
@@ -0,0 +1,120 @@
...
+const (
+	defaultSSLOptions = "no-sslv3 no-tlsv10 no-tlsv11 no-tls-tickets"
+	// TLS up to 1.2
+	defaultSSLCiphers = "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384"
+	// TLS 1.3
+	defaultSSLCipherSuites = "TLS_AES_128_GCM_SHA256:TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256"
+	// Ref: https://github.com/haproxy/spoa-modsecurity/blob/3c895f3e7dd291dba19d57ba054b277e6fb80ca4/README#L70
+	defaultModsecurityArgs = "unique-id method path query req.ver req.hdrs_bin req.body_size req.body"
+	// Ref: https://github.com/corazawaf/coraza-spoa/blob/179cf897147e165ccc1b8974888d8029737b7af5/doc/config/coraza.cfg#L14
+	DefaultCorazaArgs = "app=hdr(host) id=unique-id src-ip=src src-port=src_port dst-ip=dst dst-port=dst_port method=method path=path query=query version=req.ver headers=req.hdrs body=req.body"
```
jcmoraisjr, line 28: This can be changed to private.

pkg/haproxy/instance_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/kylelemons/godebug/diff"
 	yaml "gopkg.in/yaml.v2"
 
+	ingtypes "github.com/jcmoraisjr/haproxy-ingress/pkg/converters/ingress/types"
```
jcmoraisjr, line 31: This adds a layer violation, haproxy package doesn't know about converters. Removing this will also fix the circular reference. See also the comment below about importing the const.

pkg/haproxy/instance_test.go
```diff
@@ -4912,7 +4939,13 @@ func TestModSecurity(t *testing.T) {
 		globalModsec.Endpoints = test.endpoints
 		globalModsec.Timeout.Connect = "1s"
 		globalModsec.Timeout.Server = "2s"
-		globalModsec.Args = test.modsecAgentArgs
+		if test.modsecAgentArgs != nil {
+			globalModsec.Args = test.modsecAgentArgs
+		} else {
+			globalModsec.Args = strings.Split(ingtypes.DefaultCorazaArgs, " ")
```
jcmoraisjr, line 4945: You can add arbitrary strings here, it doesn't need to be the real configuration. A simple copy/paste from the real data is fine.

pkg/haproxy/instance_test.go
```diff
@@ -4912,7 +4939,13 @@ func TestModSecurity(t *testing.T) {
 		globalModsec.Endpoints = test.endpoints
 		globalModsec.Timeout.Connect = "1s"
 		globalModsec.Timeout.Server = "2s"
-		globalModsec.Args = test.modsecAgentArgs
+		if test.modsecAgentArgs != nil {
```
jcmoraisjr, line 4942: Maybe `if test.useCoraza`?

pkg/haproxy/types/types.go
```diff
@@ -89,6 +89,7 @@ type Global struct {
 	CustomHTTPHAResponses   []HTTPResponse
 	CustomSections          []string
 	CustomTCP               []string
+	UseCoraza               bool
```
jcmoraisjr, line 92: Better moving to ModSecurity struct.

https://github.com/jcmoraisjr/haproxy-ingress/blob/635669c5debb1b479117384c62f6171fff4dc3e6/pkg/haproxy/types/types.go#L182-L186

jcmoraisjr (commented):
Thanks again for the contribution! See some comments below. Mostly good, just bothering a bit about where to place the use-coraza config key.

I'm also proposing not to move the defaults.go file, which would create another large commit, so try to squash everything into a single commit - or a few more commits with incremental steps if you prefer.

pkg/converters/ingress/types/defaults.go
mac-chaffee: Even if I remove `DefaultCorazaArgs` from instance_test.go, the string still needs to be referenced from `/pkg/converters/ingress/ingress.go`. That was the only place I could find where both the output of `CreateDefaults` and the user-supplied overrides (use-coraza) are both present, which are needed so I can conditionally set the default args.

Since this idea of a "default value that depends on another value" is new, what if we just tell users to set their own custom args in addition to setting `modsecurity-use-coraza: true`? 

I predict many people might set their own `app=hdr(host)` arg anyway to make their Coraza config.yaml easier to write. This would simplify the code too since we wouldn't have these encapsulation issues.

jcmoraisjr: > the string still needs to be referenced from /pkg/converters/ingress/ingress.go

Which will be in the same package after moving back to pkg/converters/ingress

> what if we just tell users to set their own custom args in addition to setting modsecurity-use-coraza: true

Sure let's try that approach. Can you describe a bit more what do you have in mind? Just an overview on how would be the user intf - keys and their meaning.

mac-chaffee: I'm thinking the docs we give to users would be essentially:

> In order to use Coraza, you must:
> 1. Set `modsecurity-use-coraza: true` in your values.yaml
> 2. Set `modsecurity-args: "app=hdr(host) id=unique-id src-ip=src src-port=src_port dst-ip=dst dst-port=dst_port method=method path=path query=query version=req.ver headers=req.hdrs body=req.body"`
> 3. Set `extraContainers` to run coraza-spoa ...
> 4. Create a ConfigMap to store the Coraza config.yaml, making sure that you either specify `default_app` or change the `app` arg in `modsecurity-args` to match your application name(s) in config.yaml

 Then all `modsecurity-use-coraza` would do is change the `deny if` conditional and the `message` name.

mac-chaffee: An added bonus of us not hard-coding the default coraza `args` is that we aren't tying ourselves to that `experimental` branch of coraza-spoa, which still seems to be experimenting with the exact args that they expect us to set.

jcmoraisjr: I see, there are no magic args, but instead a user copy/paste from doc to a configmap entry. It sounds great, let's make that way.

mac-chaffee:
@sealneaward The missing `hostname` field in the logs is indeed a Coraza bug, filed this issue about it: https://github.com/corazawaf/coraza/issues/515

mac-chaffee:
@jcmoraisjr sorry for the force-push spam. I removed the custom `args` like we discussed (user must supply them now) and expanded the modsecurity.md example docs to include a section on Coraza.

docs/content/en/docs/examples/modsecurity.md
```diff
@@ -186,3 +186,24 @@ modsecurity-spoa-6596c6b444-cht27   2/2     Running   0          14m
...
+In order to use Coraza, the process is essentially the same as described in the above sections, with a few exceptions. First, in the haproxy-ingress ConfigMap, add the following two additional keys:
+
+```yaml
+  modsecurity-use-coraza: true
+  modsecurity-args: "app=hdr(host) id=unique-id src-ip=src src-port=src_port dst-ip=dst dst-port=dst_port method=method path=path query=query version=req.ver headers=req.hdrs body=req.body"
+```
+
+You may require different `modsecurity-args` depending on your Coraza config and your version of coraza-spoa. Check the [coraza-spoa README](https://github.com/corazawaf/coraza-spoa) for full details.
+
+Second, you'll need to change the container image to a coraza-spoa image (you must supply your own) and create a configmap to hold the Coraza config.yaml. An example can be found [here](/resources/coraza-deployment.yaml).
```
jcmoraisjr, line 203: missing this file?

pkg/converters/ingress/defaults.go
```diff
@@ -96,6 +96,7 @@ func createDefaults() map[string]string {
 		types.GlobalModsecurityTimeoutIdle:       "30s",
 		types.GlobalModsecurityTimeoutProcessing: "1s",
 		types.GlobalModsecurityTimeoutServer:     "5s",
+		types.GlobalModsecurityUseCoraza:         "false",
```
jcmoraisjr, line 99: No need to initialize with zero values - empty string translates to false.

pkg/haproxy/instance_test.go
```diff
@@ -4922,18 +4951,35 @@ func TestModSecurity(t *testing.T) {
...
+<<backends-default>>
+<<frontends-default>>
+<<support>>` + modsec)
+		}
 		if test.modsecAgentExp != "" {
 			c.containsText("spoe-modsecurity.conf", c.readConfig(c.tempdir+"/spoe-modsecurity.conf"), test.modsecAgentExp)
 		}
+		if test.modsecOtherExp != "" {
+			c.containsText("spoe-modsecurity.conf", c.readConfig(c.tempdir+"/spoe-modsecurity.conf"), test.modsecOtherExp)
+		}
```
jcmoraisjr, line 4982: Do we need both `modsecAgentExp` and `modsecOtherExp`?

jcmoraisjr (commented):
Awesome thanks! A few other comments below.

> sorry for the force-push spam

No problem at all, keep pushing =)

docs/content/en/docs/examples/modsecurity.md
mac-chaffee: Turns out the folder was being .gitignored since `resources/` [matches any subfolder](https://git-scm.com/docs/gitignore) named `resources`, including `static/resources`. Fixed in the latested push

pkg/haproxy/instance_test.go
mac-chaffee: Yes `modsecOtherExp` is just another contiguous substring to search for inside the templated file. Without it, the tests that check `var-prefix` would need to basically include a full copy of `spoe-modsecurity.conf` since the `var-prefix` and the `args` are so far apart in the file.

jcmoraisjr: Ah sure it's a substring. Nevermind. Just missing a local e2e and we're done.

docs/static/resources/coraza-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+    matchLabels:
+      run: coraza-spoa
+  template:
+    metadata:
+      labels:
+        run: coraza-spoa
+    spec:
+      containers:
+      - name: coraza-spoa
+        image: YOU-MUST-PROVIDE-YOUR-OWN-IMAGE
```
jcmoraisjr, line 21: Just pushed `docker.io/jcmoraisjr/coraza-spoa:experimental` and also `quay.io/jcmoraisjr/coraza-spoa:experimental`, adding PR 36 (which I'll add some comments shortly) and the diff below:

```diff
diff --git a/internal/request.go b/internal/request.go
index 841e924..c1fb139 100644
--- a/internal/request.go
+++ b/internal/request.go
@@ -51,13 +51,13 @@ func (s *SPOA) processRequest(msg spoe.Message) ([]spoe.Action, error) {
 			app, ok = s.applications[arg.Value.(string)]
 			if !ok {
 				if len(s.defaultApplication) > 0 {
-					fmt.Printf("application %q not found, using default application %q\n", arg.Value.(string), s.defaultApplication)
 					app, ok = s.applications[s.defaultApplication]
 					if !ok {
-						return nil, fmt.Errorf("default application %q not found", s.defaultApplication)
+						return nil, fmt.Errorf("default application not found: %s", s.defaultApplication)
 					}
+					app.logger.Debug("application not found, using default", zap.Any("application", arg.Value), zap.String("default", s.defaultApplication))
 				} else {
-					return nil, fmt.Errorf("application %q not found", arg.Value.(string))
+					return nil, fmt.Errorf("application not found: %v", arg.Value)
 				}
 			}
 		case "id":
```

This should be temporary, lets make the official image happen.

docs/static/resources/coraza-deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coraza-config
+  namespace: ingress-controller
+data:
+  # Check the official documentation: https://github.com/corazawaf/coraza-spoa
+  config.yaml: |
+    bind: 127.0.0.1:12345
```
jcmoraisjr, line 59: This container couldn't be reached (except as a sidecar) if binding to localhost. Besides that it's also failing the liveness probe, which uses the public IP. Need to change to `:12345` instead so all the network intf will listen this port.

pkg/converters/ingress/defaults.go
```diff
@@ -96,6 +96,7 @@ func createDefaults() map[string]string {
 		types.GlobalModsecurityTimeoutIdle:       "30s",
 		types.GlobalModsecurityTimeoutProcessing: "1s",
 		types.GlobalModsecurityTimeoutServer:     "5s",
+		types.GlobalModsecurityUseCoraza:         "",
```
jcmoraisjr, line 99: Empty string is the zero value for the string type, so this whole line can be removed.

jcmoraisjr (commented):
A few other findings after run an e2e locally. Other than that LGTM and we're ready to go.

docs/static/resources/coraza-deployment.yaml
jcmoraisjr: Forgot to comment in the respective lines - if changing the default image, need to remove (or change) the last paragraph from the doc as well.

jcmoraisjr:
Lgtm, thanks! Merging now.

PR #968, opened 2022-11-23
Update to hugo v0.110.0 and update node deps

mac-chaffee:
I noticed while working on the Coraza docs that my version of Hugo was too new to render the docs locally, specifically the change from `blackfriday` to `goldmark` as the markdown renderer on v0.60. I also updated the node deps too.

**NOTE:** @jcmoraisjr It looks like you build and publish the site locally. If so, you'll need to update your local hugo version and run `npm update` before publishing if this PR is merged.

Other changes:

* Gitignore'd the .hugo_build.lock file (which was added [in a later version of Hugo](https://gohugo.io/news/0.89.0-relnotes/))
* Changed the baseURL, it's apparently supposed to be the full URL where the site is hosted: https://gohugo.io/getting-started/configuration/#baseurl Without this change, I just get 404s locally.
* Allowed rendering raw HTML which is needed on the home page
* Removed some tab characters on the home page which were causing the HTML to render as code blocks for some reason
* Added some extra tips to the contribution guide
    - Indented the code block with spaces, which is needed to prevent the numbered list from re-starting at "1."
 * Updated to docsy 0.6.0

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -2,3 +2,4 @@ package-lock.json
 public/
 resources/
 node_modules/
+.hugo_build.lock
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -1,4 +1,4 @@
-baseURL = "/v0.14"
+baseURL = "https://haproxy-ingress.github.io"
 title = "HAProxy Ingress"
 
 enableRobotsTXT = true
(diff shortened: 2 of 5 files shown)


jcmoraisjr:
Wow, thank you so much for investing some time on that! Sorry about taking so long to have a proper look. Just a few suggestions below.

- It seems that the baseURL problem is the path, maybe you can change to just `/` instead. This will not work for deploying older versions, which needs the path prefix, but that change I can take care in the deployment scripts - I have some scripts and a pinned hugo version to deploy the doc as you suggested. Now I can remove the pin o/
- I can (apparently) successfully run with Docsy v0.6.0, maybe you can update its version as well?

mac-chaffee:
I updated docsy to 0.6.0. They did have one notable change: Docsy no longer includes submodules of its own. Instead you have to run `npm install` inside the docsy folder after you clone it to fetch those same deps, so I added that step to our docs.

Not sure what you mean about the baseURL. The docs say it should be a full URL, not just `/`. Is there a change you'd like me to make?

jcmoraisjr:
Perfect thanks!

Regarding the baseURL it works with relative path as well, and being relative helps me on some specific use cases that I need to test/run the final html version before syncing to the prod site.

mac-chaffee:
Should be good now!

jcmoraisjr:
Sure, thank you for the great job! Merging now.

Issue #950, opened 2022-09-22
Error signing new certificate for redirected domain


tumd:
This issue was briefly discussed on Slack, but I'm adding it here as well, for completeness.

**Description of the problem**

I would like to add a second domain to an existing ingress and redirect all requests hitting secondary domain to the primary domain using the annotation `redirect-from: secondary.domain.tld`. Both domains should receive a valid certificate from the acme environment that's bundled with haproxy-ingress.
To make it more complex, I would like to only allow requests from a list of IP’s using the `haproxy-ingress.github.io/config-backend`-annotation - however I don't think this is relevant to the bug.

The bellow attached ingress-manifest fails when obtaining a certificate for both domains.

When looking at the generated haproxy.cfg inside the pod, I can see that my `config-backend` statement should not matter as the path `/.well-known/acme-challenge/` is handled in `_front_http` section, before any rules in the backend is applied.

However, it seems that the `http-request redirect-rule` may be hit regardless of the `acme-challenge`-acl is true of not - and this is what I think is the problem. It should only perform the redirect if `acme-challenge`-acl is false.

**Expected behavior**

A certificate being issued for both primary and secondary domain, regardless of any redirects.

Temoporary removing `haproxy-ingress.github.io/redirect-from` from the manifest can be used as a work-around while issuing the certificate.

**Steps to reproduce the problem**

1. Apply a manifest similar to the one bellow, with your own domains, using the `redirect-from`-annotation to redirect from one of the domains.

**Environment information**

HAProxy Ingress version: `v0.13.9`

Command-line options:

```
--acme-server
--acme-track-tls-annotation
```

Global options:

```yaml
config-frontend: |
  unique-id-format %{+X}o\ ID-%[uuid(4)]
  unique-id-header X-Request-ID
  http-request capture unique-id len 40
  capture cookie elsess len 30
```

Ingress objects:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    haproxy-ingress.github.io/cert-signer: acme
    haproxy-ingress.github.io/config-backend: http-request deny unless { src -f /etc/haproxy/custom-maps/office.map }
    haproxy-ingress.github.io/redirect-from: secondary.some.tld
    kubernetes.io/ingress.class: haproxy
  labels:
    app: devportal
  name: devportal
  namespace: devportal
spec:
  rules:
  - host: primary.some.tld
    http:
      paths:
      - backend:
          service:
[...]

PR #971, opened 2022-12-03
Fix host redirects when acme is enabled

jcmoraisjr:
Acme validation happens on plain HTTP and should skip redirects, so that the acme service (e.g. Let's Encrypt) can properly validate the domain ownership. This wasn't properly happening with `redirect-from` and `redirect-to` keys, which makes the signer to fail. This update skips those redirects when acme is enabled and path matches the acme challenge.


jcmoraisjr:
#950

Issue #970, opened 2022-12-01
Ignore Ingress objects that contain invalid configuration


mrueg:
**Description of the problem**

Setting `ingress.kubernetes.io/auth-headers-succeed: ""` (empty string) in an Ingress Object will create an invalid haproxy.cfg and haproxy will log:

 ```
[ALERT] xx/yy (zz) : parsing [/etc/haproxy/haproxy.cfg:800] : error detected in backend 'mybackend' while parsing 'http-request lua.auth-intercept' rule : expect 6 arguments.
```

the configuration will stay at the old config, but other new ingress objects won't be recognized.

**Expected behavior**
Ingress updates will continue and the invalid configuration will be ignored.

**Steps to reproduce the problem**

1. Set `ingress.kubernetes.io/auth-headers-succeed: ""` in an Ingress object

**Environment information**

HAProxy Ingress version: `v0.13.9`

**Further information**
Likely globToLuaPattern() in 
https://github.com/jcmoraisjr/haproxy-ingress/commit/0b315c4ffe2efd033c4197cbcfed00cb41725334 needs to be adjusted for empty strings.

jcmoraisjr:
Hi thanks for reporting this. In fact there should be quite some config keys that break the configuration, so feel free to report any of them you can find. This however will never be complete without a webhook that parses a full configuration on every ingress change, because some of them simply cannot be validated like the `config-*` family.

I've changed the suggested behavior in the case of an empty string, my idea is that one will want to remove all the headers if configuring these keys like that. Please share your thoughts if this doesn't make much sense. This option (empty or dash) is now documented as well.

mrueg:
@jcmoraisjr Yes you're right an admissionwebhook is probably the best solution and then dry-run the generated config. 

I think your solution in https://github.com/jcmoraisjr/haproxy-ingress/pull/972 sounds fine, as it's documented and likely what someone has in mind when adding an empty string.

jcmoraisjr:
Fixed in v0.14.0-beta.3, closing.

PR #972, opened 2022-12-03
Makes auth-headers not copying on empty string

jcmoraisjr:
By default `auth-headers-*` copy all provided headers to/from auth service, client or backend. A non documented `-` makes no header to be copied and an empty string breaks the configuration. Now an empty string is internally converted to a dash and this option is also documented.


jcmoraisjr:
#970

PR #974, opened 2022-12-13
go.mod: Update dependencies / haproxy 2.4.20

mrueg:
Update dependencies before release.

* go dependencies
* golang itself (1.17 -> 1.18)
* haproxy (2.4.19 -> 2.4.20)


mrueg:
Looks like this change removes ConfigMapResourceLock and breaks the upgrade to k8s 1.24+ https://github.com/kubernetes/client-go/commit/276ea3ed979947d7cdd4b3d708862245ddcd8883

jcmoraisjr:
Lgtm, merging now, thanks!

k8s v1.24+ update was a bit noisy to be merge to v0.14 so I preferred to leave client-go on v0.23. I wasn't considering however take too long to release this branch. New apiserver versions should maintain backward compatibility regarding leader election, so we might be good here.

Issue #981, opened 2023-02-03
auth-url does not handle well service with port names instead of numbers


didiercolens:
**Description of the problem**

Assume a service with the following manifest (snippet):
```yaml
  name: my-authentication
spec:
  ports:
  - name: authentication
    port: 1234
    protocol: TCP
    targetPort: authentication
```
the corresponding deployment has the following port definition:
```yaml
spec:
  containers:
  - ports:
      - containerPort: 1234
        name: authentication
```

if an ingress is configured as follows (the documentation says to use port number):
```
haproxy-ingress.github.io/auth-url: svc://bdb-authentication:1234/api/v2/auth
```

it will silently fail (no error logged anywhere) and always deny all requests 

**Expected behaviour**

Either an error is logged to warn that the backend could not be found or the code always converts ports to port number so it works in both cases.

What works is haproxy-ingress.github.io/auth-url: svc://bdb-authentication:authentication/api/v2/auth

So an alternate (quick) solution is to update the doc and specify that the port should match the targetPort value in the service.

**Steps to reproduce the problem**

1. create a deployment that has a port name
2. expose that deployment and make sure targetPort uses the port name
3. set haproxy-ingress.github.io/auth-url to svc url with port number used

**Environment information**

HAProxy Ingress version: `v0.14.0`

mischmi2:
As an additional piece of this, it took a long time to debug what was happening due to the silent error catch in https://github.com/jcmoraisjr/haproxy-ingress/blob/4862b877a9fb34069478a31ec314fc42955752ad/pkg/converters/ingress/annotations/backend.go#L184, that doesn't log when it gets hit. In this case, the port name mismatch caused the backend to not be found. 

https://github.com/jcmoraisjr/haproxy-ingress/blob/4862b877a9fb34069478a31ec314fc42955752ad/pkg/haproxy/types/backends.go#L331 constructed

```
namespace + "_" + name + "_" + port
```

As the map key for the backend, where `port` appears to actually be coming from the service object `targetPort`. When the backend gets stored it's as `my-namespace_my-authentication_authentication` instead of `my-namespace-my-authentication_1234`, and when the ingress object tries to lookup the backend it's looking for `my-namespace-my-authentication_1234`.

jcmoraisjr:
Hi, thanks for the detailed description and digging into the code. You folks are 💯!

[...]

PR #982, opened 2023-02-05
Add warning if auth external svc isnt found

jcmoraisjr:
Auth External works by parsing the authentication service url and looking for a pre-configured backend. The pre-configuration part works by making a normalization in the k8s service port number, so different inputs for the same service will output the same backend. This might change the port name of the backend on named port configurations.

A proper fix is to make the backend creation accessible by the updater code, so we don't need to pre-initialize the backend in the ingress step, and don't need to guess the backend port name in the updater step. So, for now, we'll improve logging and documentation.

This should be merged as far as v0.13, which implements the first version of auth external.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -809,7 +809,7 @@ Configures External Authentication options.
 
 * `<proto>`: can be `http`, `https`, `service` or `svc`.
 * `<name>`: the IP or hostname if `http` or `https`, or the name of a service if `service`. `svc` is an alias to `service`. Note that the hostname is resolved to a list of IP when the ingress is parsed and will not be dynamically updated later if the DNS record changes.
-* `<port>`: the port number, must be provided if a service is used and can be omitted if using `http` or `https`.
+* `<port>`: the port number, must be provided if a service is used and can be omitted if using `http` or `https`. If the service uses named ports, use the service's `port.targetPort` field value instead.
 * `<path>`: optional, the fully qualified path to the authentication service.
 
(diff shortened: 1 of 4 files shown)


jcmoraisjr:
#981

docs/content/en/docs/configuration/keys.md
```diff
@@ -809,7 +809,7 @@ Configures External Authentication options.
 
 * `<proto>`: can be `http`, `https`, `service` or `svc`.
 * `<name>`: the IP or hostname if `http` or `https`, or the name of a service if `service`. `svc` is an alias to `service`. Note that the hostname is resolved to a list of IP when the ingress is parsed and will not be dynamically updated later if the DNS record changes.
-* `<port>`: the port number, must be provided if a service is used and can be omitted if using `http` or `https`.
+* `<port>`: the port number, must be provided if a service is used and can be omitted if using `http` or `https`. If the service uses named ports, service's target port name or number should be used instead.
```
didiercolens, line 812: remove "or number"

jcmoraisjr: According to the api description, a port name or number can be used here:

https://github.com/kubernetes/api/blob/0478a3e95231398d8b380dc2a1905972be8ae1d5/core/v1/types.go#L4776-L4785

Maybe we can give some highlight on the correct value to use by rephrasing this to "If the service uses named ports, use the service's `port.targetPort` field value instead."?

didiercolens: yes, that would clarify :-)

Issue #979, opened 2023-01-26
Annotation allowlist-source-range or denylist-source-range in a Service used as backend by multiple paths are applied only to the first one


kcnpnew:
**Description of the problem**

In version 0.13.4 I noticed what I think is a strange and undocumented behavior when using allowlist-source-range or denylist-source-range annotation in a Service used as backend by multiple paths

Below is a simple test case with two Ingress and one Service

```
apiVersion: v1
kind: Namespace
metadata:
  name: test
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  namespace: test
  name: test1
  annotations:
    kubernetes.io/ingress.class: "haproxy"
spec:
  rules:
  - host: test1.app1
    http:
      paths:
      - path: /test
        pathType: Prefix
        backend:
          service:
            name: test-service
            port:
              name: http
      - path: /test2
        pathType: Prefix
        backend:
          service:
            name: test-service
            port:
              name: http
  - host: test1.app2
    http:
      paths:
      - path: /test
        pathType: Prefix
        backend:
          service:
            name: test-service
            port:
              name: http
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  namespace: test
  name: test2
  annotations:
    kubernetes.io/ingress.class: "haproxy"
spec:
  rules:
  - host: test2.app
    http:
      paths:
      - path: /test
        pathType: Prefix
        backend:
          service:
            name: test-service
            port:
              name: http
---
apiVersion: v1
kind: Service
metadata:
  name: test-service
  labels:
    app: test
  namespace: test
  annotations:
    ingress.kubernetes.io/allowlist-source-range: "10.1.0.0/16"
spec:
  selector:
    app: test
  ports:
  - name: http
    port: 80
    targetPort: 8080

```

**Expected behavior**

As the annotation was applied in the Service I was hoping all paths using that Service as backend would get the source restriction rules.

**Actual behavior**

The source restriction rules are limited to the first path using the Service as a backend.

The generated backend section at the haproxy config is.

```
backend test_test-service_8080
    mode http
    balance roundrobin
    acl https-request ssl_fc
    # path01 = test1.app1/test
    # path02 = test1.app1/test2
    # path03 = test1.app2/test
    # path04 = test2.app/test
    http-request set-var(txn.pathID) var(req.base),map_dir(/etc/haproxy/maps/_back_test_test-service_8080_idpath__prefix.map)
[...]

PR #984, opened 2023-02-08
Fix path scoped annotation on service resources

jcmoraisjr:
Backend scoped annotations apply just once per service resource, and have the most precedence. With this behavior in mind, service annotations were being applied along with the annotation mapper creation. This works nice with backend scoped annotations.

Path scoped, however, has the ability to apply distinct values to the same backend. When path scoped annotations are applied via a service resource, only the first path is being applied, along with annotation mapper creation. This update ensures that the service annotations are applied on every pathlink that points to it. This change will spread service annotations over all referenced pathlinks, however this doesn't end up with duplicated configurations because the config builder already deduplicates configurations on the same key, with the same value, to the same backend. Warning on annotation conflict isn't a problem as well, a conflict is only warned on distinct values.


jcmoraisjr:
#979

PR #986, opened 2023-02-11
Prefer ingressClassName over annotations in docs

mac-chaffee:
Apparently using ingressClassName has been preferred over annotations since k8s 1.18: https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

This PR just updates the two examples we have in the docs to use ingressClassName instead of the annotations. Figure we ought to send new users down the right path!


jcmoraisjr:
Makes sense, thanks for the suggestion. This however needs to be followed by another snippet that creates the ingress class.

jcmoraisjr:
Hi, just realized I wrongly tagged this as v0.15 only, but it makes sense to add it to v0.14 as well so we update doc of the current version earlier. That said, sorry about the long delay. lgtm, merging it now.

Issue #985, opened 2023-02-10
Use oauth X-Auth-* with http-header-match annotation 


knfoo:
**Description of the problem**
I have an application where users login via the oauth2_proxy. I want to route the users to different backends based on the value in X-Auth-Groups.
I have asked about this in [slack](https://kubernetes.slack.com/archives/C01D8GKKT5H/p1674744569214379) and was pointed to a feature that was merged but not yet released Add http-header-match annotation #944. Which seems to fit the usecase quite nicely.
I have build haproxy-ingress from source to get it deployed. When I create a ingress object now the header match is added to the haproxy configuration. But I am no longer able to login as the header match denies access to the oauth2 flow.

Frontend config
```
frontend _front_https
    mode http
    bind :443 ssl alpn h2,http/1.1 crt-list /etc/haproxy/maps/_front_bind_crt.list ca-ignore-err all crt-ignore-err all
    option httplog
    http-request set-var(req.path) path
    http-request set-var(req.host) hdr(host),field(1,:),lower
    http-request set-var(req.base) var(req.host),concat(\#,req.path)
    http-request set-var(req.hostbackend) var(req.base),map_dir(/etc/haproxy/maps/_front_https_host__prefix_01.map) if { hdr(X-Auth-Request-Groups) -m reg -- '(group-id)' }
    http-request set-var(req.hostbackend) var(req.base),map_dir(/etc/haproxy/maps/_front_https_host__prefix.map) if !{ var(req.hostbackend) -m found }
    http-request set-header X-Forwarded-Proto https
    http-request del-header X-SSL-Client-CN
    http-request del-header X-SSL-Client-DN
    http-request del-header X-SSL-Client-SHA1
    http-request del-header X-SSL-Client-SHA2
    http-request del-header X-SSL-Client-Cert
    use_backend %[var(req.hostbackend)] if { var(req.hostbackend) -m found }
    default_backend ingress-controllers_dh2-2-haproxy-ingress-default-backend_http
```

The /etc/haproxy/maps/_front_https_host__prefix_01.map
```
domain.com#/oauth2 observability_oauth2-proxy-helm_http
domain.com#/ observability_grafana_3000

```

Backend config
```
backend observability_grafana_3000
    mode http
    balance roundrobin
    acl https-request ssl_fc
    http-request redirect scheme https if !https-request
    option forwardfor if-none
    http-request set-header X-Real-IP %[src]
[...]

PR #988, opened 2023-02-13
Add optional frontend based external authentication call

jcmoraisjr:
External authentication services has the ability to inject headers on the fly in the current request. Such header will be visible to the backend server in the case the authentication succeed, and to the caller/client in the case the authentication fails. This works that way since v0.13.

Starting on v0.15 HAProxy Ingress has the ability to route requests based on HTTP headers, along with the well known hostname+path supported since the very first version. However headers added or updated by the authentication service isn't visible to HAProxy Ingress routing rules, this happens because the authentication service is called from the backend declaration for performance reasons, but the routing rules must be configured in the frontend due to its logic: it needs to choose a backend based on hostname, path, and now HTTP headers.

This is a very first step on behalf of optionally bringing the authentication service call to the frontend. This should start in a very rudimentar way and with some performance impact if used on several distinct pathlinks, but it's good enough on use cases that such functionality is far more important than a small performance penalty. Other than that: 1) this is optional, enabled by the user, and doesn't make any difference on the former configuration; 2) future versions will make improvements on such frontend configuration; 3) the performance penalty grows based on the number of pathlinks, so adding a dozen or so on moderately used proxy won't hurt, as much as adding one or two on a highly used one.

TODO

* [x] haproxy types and logic
* [x] haproxy template
* [x] converter implementation
* [x] docs
* [x] unit tests
* [x] exploratory tests

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -305,6 +305,7 @@ The table below describes all supported configuration keys.
 | [`allowlist-source-header`](#allowlist)              | Header name that will be used as a src  | Path    |                    |
 | [`app-root`](#app-root)                              | /url                                    | Host    |                    |
 | [`assign-backend-server-id`](#backend-server-id)     | [true\|false]                           | Backend | `false`            |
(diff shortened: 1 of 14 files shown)


jcmoraisjr:
#985

PR #994, opened 2023-02-22
Add gateway version v1beta1

jcmoraisjr:
Add converter, method facade, watcher, schema and converter that just duplicates current v1alpha2 version.


withnale:
Is this going to be merged soon? I want to move to the latest release for the gateway CRDs 0.8.x but this means that v1alpha versions have disappeared.

Issue #1000, opened 2023-05-17
TCP service configuration lost after restart


arun11299:
HAProxy Ingress version: 0.13.7

I am using "--tcp-services-configmap" option to provide configuration required to forward TCP traffic to backend services. The config map looks like:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: haproxy-tcp-configmap
  namespace: {{ .Values.global.namespace.name }}
data:
  "8006": {{ .Values.global.namespace.name }}/{{ include "mongodb.fullname" . }}:8006
  "8014": {{ .Values.global.namespace.name }}/{{ include "redis-tunnel.fullname" . }}:8014
```

On first install (using helm, kubernetes v1.24.0) the generated config inside the ingress pod has all the configuration related to TCP services:
```
  # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
# #
# #   TCP SERVICES (legacy, via configmap)
# #
#
listen _tcp_amuralid-test_amuralid-test-mongodb_8006
    bind :8006
    mode tcp
    server srv001 192.168.160.205:8006 check port 8006 inter 2s
listen _tcp_amuralid-test_amuralid-test-redis-tunnel_8014
    bind :8014
    mode tcp
    server srv001 192.168.160.203:8014 check port 8014 inter 2s
``` 

Now I delete the haproxy-ingress pod and when it comes up again, the generated config does not have any of the TCP services.

What is the WA ? I tried to dabble with "config-tcp-service" option, but I could not figure out from the documentation on how to convert the old configmap to new one.

Thanks.

arun11299:
Seeing the same issue with 0.14.2 as well.

arun11299:
If I make any dummy change to the tcp-services configmap, then it takes up the config.

jcmoraisjr:
Hi, thanks for reporting and sorry about the delay. I can confirm this behavior, sorry for that. We're planning a new tag soon in order to update dependencies and should patch that for both branches.

arun11299:
Thanks a lot. Happy to help.

jcmoraisjr:
Fixed in v0.14.3 and v0.13.12, closing.

PR #1001, opened 2023-05-27
Fixes configmap based tcp sync

jcmoraisjr:
This update ensures that configmap based TCP services are applied in the very first controller reconciliation. Cur config is only defined from the second reconciliation, so the TCP service would be empty if the controller bootstrap doesn't call the reconciliation at least twice.


jcmoraisjr:
#1000

Issue #999, opened 2023-04-04
HAProxy Ingress continuously reloading: tcp-services (configmap)


bootc:
**Description of the problem**

Pods come and go in my cluster quite a lot, as one namespace in particular is used to run CI jobs for some of our projects. HAProxy Ingress notices these Pods and their associated Secrets coming and going, but notices non-existent changes to the tcp-services ConfigMap, triggering an unnecessary reload of HAProxy.

**Expected behavior**

The Ingress controller should see that no changes are actually needed and not reload HAProxy.

**Steps to reproduce the problem**

1. Configure a TCP services ConfigMap using `controller.tcp` in the Helm chart, for example.
2. Delete or create some Pods in the cluster.
3. `dynupdate.go:104] need to reload due to config changes: [tcp-services (configmap)]`

**Environment information**

HAProxy Ingress version: `v0.14.2`

Command-line options:

```
--configmap=tcl-ingress/haproxy-ingress
--ingress-class=haproxy
--master-socket=/var/run/haproxy/master.sock
--sort-backends
--tcp-services-configmap=tcl-ingress/haproxy-ingress-tcp
--ignore-ingress-without-class=true
--publish-service=tcl-ingress/haproxy-ingress-nlb
--wait-before-shutdown=30
```

Global options:

```
healthz-port: "10253"
https-log-format: default
prometheus-port: "9101"
stats-port: "1936"
syslog-endpoint: stdout
syslog-format: raw
tcp-log-format: default
use-haproxy-user: "true"
use-proxy-protocol: "true"
```

Ingress objects:

_Probably not relevant._

TCP Services ConfigMap keys:

```
"22": gitlab/gitlab-gitlab-shell:22:PROXY
"8140": puppet/puppetserver:443:PROXY
```

Log from one of the updates:

```
I0404 12:07:37.002014       7 controller.go:338] starting haproxy update id=366
I0404 12:07:37.002096       7 converters.go:67] applying 2 change notifications: [update/Secret:gitlab-runner/runner-2gxzz8rg-project-63-concurrent-0q2tqt add/Secret:gitlab-runner/runner-2gxzz8rg-project-63-concurrent-0q2tqt]
I0404 12:07:37.002307       7 ingress.go:232] syncing 0 host(s) and 0 backend(s)
I0404 12:07:37.004877       7 instance.go:441] updating 0 host(s): []
I0404 12:07:37.004888       7 instance.go:458] updating 0 backend(s): []
[...]

PR #1003, opened 2023-05-28
Ensure predictable tcp by sorting endpoints

jcmoraisjr:
Dynamic update code relies on comparing tcp services structs in order to verify whether any tcp service has changed or not, and if so, haproxy need to be restarted. Endpoint list is the only tcp data that might lead to differences in the case the same values are present, but in distinct order. Because of that we're applying a sort in order to generate predictable, hence comparable results.


jcmoraisjr:
#999

Issue #978, opened 2022-12-31
HAProxy fails to start


bootc:
**Description of the problem**

With HAProxy Ingress 0.14.0 and the external HAProxy configuration, the HAProxy container fails to start in my configuration.

```
[NOTICE]   (1) : haproxy version is 2.4.20-d59cd78
[ALERT]    (1) : parsing [/etc/haproxy/haproxy.cfg:137] : 'bind unix@/var/run/haproxy/_https_socket.sock' : 'crt-list' : cannot open the file '/var/lib/haproxy/crt/default-fake-certificate.pem'.
[ALERT]    (1) : Error(s) found in configuration file : /etc/haproxy/haproxy.cfg
[ALERT]    (1) : Fatal errors found in configuration.
```

It works fine if I use the combined HAProxy + HAProxy Ingress container.

In case it's relevant, I only have two Ingress resources picked up by HAProxy Ingress and both are configured with `ingress.kubernetes.io/ssl-passthrough: "true"`.

**Expected behavior**

HAProxy should start when run separately from the HAProxy Ingress container.

**Steps to reproduce the problem**

1. Install the chart with `controller.haproxy.enabled=true`.

**Environment information**

HAProxy Ingress version: `v0.14.0`

Helm values.yaml:

```
controller:
  extraArgs:
    ignore-ingress-without-class: "true"
    wait-before-shutdown: "30"

  config:
    bind-ip-addr-http: "[::]"
    bind-ip-addr-tcp: "[::]"
    https-log-format: default
    tcp-log-format: default

  replicaCount: 2

  resources:
    requests:
      cpu: 10m
      memory: 64Mi
    limits:
      memory: 192Mi

  service:
    type: LoadBalancer
    externalTrafficPolicy: Local
    ipFamilyPolicy: RequireDualStack

  haproxy:
    enabled: true

    resources:
      requests:
        cpu: 10m
        memory: 64Mi
      limits:
        memory: 96Mi

  stats:
    enabled: true

  metrics:
    enabled: true

  serviceMonitor:
    enabled: true
```

Ingress objects:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/ssl-passthrough: "true"
  name: test
spec:
  ingressClassName: haproxy
  rules:
  - host: test.example.com
    http:
      paths:
      - backend:
          service:
            name: test
            port:
              name: https
        path: /
        pathType: Prefix
```

[...]

PR #1004, opened 2023-05-28
Change owner of crt/key files to haproxy pid

jcmoraisjr:
HAProxy Ingress v0.14 and older starts as root by default and create files as 0600 that should be read by HAProxy. When HAProxy runs as a sidecar and stars as non root (default since its 2.4) it won't have permission to read those files. This update makes files generated as 0600 (crt/key) having its owner to PID 99, which's the PID used by HAProxy containers.


jcmoraisjr:
#978

Issue #1005, opened 2023-06-13
Redirect loop using redirect-from-regex annotation


Nikarous:
**Description of the problem**

I am setting up redirects for 3 domains with redirect-from-regex annotations on k8s ingress object:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ...
    cert-manager.io/cluster-issuer: letsencrypt
    acme.cert-manager.io/http01-edit-in-place: "true"
    ingress.kubernetes.io/redirect-from-code: '301'
    ingress.kubernetes.io/redirect-from-regex: >-
      ^www\.service\.dev\.example\.com$|^service-test\.example\.com$|^www\.service-test\.example\.com$
    kubernetes.io/ingress.class: haproxy
spec:
  rules:
    - host: service.dev.example.com
      http:
        paths:
          - backend:
              service:
                name: example-service
                port:
                  number: 80
            path: /
            pathType: Prefix
  tls:
    - hosts:
        - service.dev.example.com
        - service-test.example.com
        - www.service-test.example.com
        - www.service.dev.example.com
      secretName: lets-cert
```
And reloading configuration (in case of scaling for example) setups redirects randomly to one of those 4 domains with message:
```
W0612 09:12:56.334065       7 host.go:107] ignoring regex redirect from '^www\.service\.dev\.example\.com$|^service-test\.example\.com$|^www\.service-test\.example\.com$' on Ingress 'example-ns/ingress-controller-for-example-service', it's already targeting to 'www.service-test.example.com'
I0612 09:13:47.721941       7 instance.go:441] updating 4 host(s): [service.dev.example.com service-test.example.com www.service-test.example.com www.service.dev.example.com]
```
And its not refreshing setup to point redirects to [service.dev.example.com](http://service.dev.example.com/). This message repeats when new pod is added and after some time it manages to lands correctly on hostname setup in ingress.

**Important notes:
I did some testing and at first I thought that | (or) symbol in regex were causing issues but it still persist in second example of ingress below. 
Now I am pretty sure that for some reason HAproxy-ingress takes host values for redirect map from hosts array under tls. This is because it puts values in the _front_redir_from__regex.map that it could not know other than from this field in ingress resource, example in second ingress below.**

example I got directly from my haproxy-ingress pod:
```
/etc/haproxy # cat maps/_front_redir_from__regex.map
[...]

PR #1010, opened 2023-07-02
Redirect hosts only to domains with associated backends

jcmoraisjr:
Redirect-from configuration key defines the source hostnames that should redirect to the one configured in the ingress manifest. This is not properly working with ingress that adds more hostnames in the TLS section because haproxy ingress adds all those hostnames as regular ones. Adding as a regular hostname is needed in order to make haproxy choose the right certificate in the TLS handshake.

The following change is making redirect-from apply redirects only on hosts that have a backend association, which would be otherwise ending up in a 404 error anyway.


jcmoraisjr:
#1005

PR #1011, opened 2023-07-02
Create endpoints on a predictable order

jcmoraisjr:
Endpoints (internal haproxy ingress struct, not the k8s one) are part of the building blocks to configure a backend or a tcp listener. HAProxy Ingress uses deep equals on some structs in order to identify equality, hence the ability to skip a reload event. When two endpoint arrays have the same content but on distinct order, they are not considered equals, although it should from the haproxy and haproxy ingress perspective. Because of that we're sorting them.

The former update (which we're reverting here) was sorting the endpoints after they get named, so the order is right but some names might be changed. Now we're sorting the endpoints on its source.

### DIFF
--- a/pkg/converters/utils/services.go
+++ b/pkg/converters/utils/services.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"fmt"
+	"sort"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


jcmoraisjr:
#999

Issue #1014, opened 2023-07-04
ignoring invalid cors headers on Ingress "*" 


a180285:
**Description of the problem**

I want to set CORS access-control-allow-headers to "*". but I got this errror.

`ignoring invalid cors headers on Ingress "xx/xx" * `

base on description [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers). "*" should be valid headers.

```
* (wildcard)
The value "*" only counts as a special wildcard value for requests without credentials (requests without [HTTP cookies]

(https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies) or HTTP authentication information). In requests with credentials, it is

 treated as the literal header name "*" without special semantics. Note that the [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) header can't be wildcarded and always needs to be listed explicitly.


```

**Expected behavior**

**Steps to reproduce the problem**

1.  add annotation of `haproxy-ingress.github.io/cors-allow-headers: '*'` to ingress

**Environment information**

HAProxy Ingress version: `v0.xx`
image version: quay.io/jcmoraisjr/haproxy-ingress:v0.14.2

# Root cause

Code [corsHeadersRegex](https://github.com/jcmoraisjr/haproxy-ingress/blob/master/pkg/converters/ingress/annotations/validator.go#L38) are hard coded. and without "*".

a180285:
@jcmoraisjr Thanks for quick fix.

jcmoraisjr:
You're welcome. Closing.

PR #1016, opened 2023-07-06
Add wildcard as a valid cors allowed header

jcmoraisjr:
Wildcard is a valid cors allowed header option, but validation wasn't allowing to configure it.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers


jcmoraisjr:
#1014

PR #1029, opened 2023-10-17
allow multiple certificates

zanettea:
I've implemented an heuristic to provide basic support for multiple certificates. It just works in my use case.
Please consider merging this change.
Thanks

### DIFF
--- a/pkg/converters/gateway/gateway.go
+++ b/pkg/converters/gateway/gateway.go
@@ -19,10 +19,8 @@ package gateway
 import (
 	"fmt"
 	"reflect"
-	"regexp"
 	"sort"
 	"strconv"
-	"strings"
 
 	api "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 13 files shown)


pkg/converters/gateway/gateway.go
```diff
@@ -451,12 +453,32 @@ func (c *converter) applyCertRef(source *Source, listener *gatewayv1alpha2.Liste
...
+	// iterate over all certRefs
+	for _, certRef := range certRefs {
+		crtFile, err := c.readCertRef(source.namespace, certRef)
+		if err != nil {
+			c.logger.Warn("skipping certificate reference on %s listener '%s': %s",
+				source, listener.Name, err)
+			return
+		}
+		// in crtFile.CommonName replace wildcards * with common regex wildcard .*
+		crtFilePattern := strings.ReplaceAll(crtFile.CommonName, "*", ".*")
```
jcmoraisjr, line 465: I'd use `HasPrefix("*.")` and, if true, use the resulting of a `TrimPrefix("*.")` into a `HasSuffix()` instead of parse and match a regex on every iteration. This is the hot path of a cluster with dozens of thousands of httproutes, so we want all the processing to be as fast as possible.

jcmoraisjr (commented):
I liked the approach, thanks for the contribution. See below a changing suggestion, and please create the pr against master, changes to branches are made later via cherry-pick.

PR #1083, opened 2024-02-29
Keep all vary header values when adding Origin

Jul13nT:
When a backend response contains multiple values for the Vary header, like "Vary: header1, header2", the resulting Vary header response from haproxy is : "Vary: header2, Origin".
This is linked to this PR : #861

The official documentation states that `req.hdr` (or `res.hdr`) returns "only the last comma-separated value of the header".
So it should be `res.fhdr` instead.


jcmoraisjr:
Lgtm thanks! Merging now.

jcmoraisjr:
We've a few bugs in the queue to be fixed, and a new round of releases should be issued just after that. It should be within a few weeks from now.

Issue #1085, opened 2024-03-01
Coraza WAF modsecurity addon injects wrong backend config


tomklapka:
**Description of the problem**
What is not working - injected backend config when using Ingress annotations:
    haproxy-ingress.github.io/waf: "modsecurity"
    haproxy-ingress.github.io/waf-mode: "deny"

```
    filter spoe engine modsecurity config /etc/haproxy/spoe-modsecurity.conf
    http-request deny deny_status 504 if { var(txn.coraza.error) -m int gt 0 }
    http-request deny if !{ var(txn.coraza.fail) -m int eq 0 }
```
Resulting in:
```
403 Forbidden
Request forbidden by administrative rules.
```
On all requests.

What is working (my current workaround):

```
    filter spoe engine modsecurity config /etc/haproxy/spoe-modsecurity.conf
    # Currently haproxy cannot use variables to set the code or deny_status, so this needs to be manually configured here
    http-request redirect code 302 location %[var(txn.coraza.data)] if { var(txn.coraza.action) -m str redirect }
    http-response redirect code 302 location %[var(txn.coraza.data)] if { var(txn.coraza.action) -m str redirect }

    http-request deny deny_status 403 hdr waf-block "request"  if { var(txn.coraza.action) -m str deny }
    http-response deny deny_status 403 hdr waf-block "response" if { var(txn.coraza.action) -m str deny }

    http-request silent-drop if { var(txn.coraza.action) -m str drop }
    http-response silent-drop if { var(txn.coraza.action) -m str drop }

    # Deny in case of an error, when processing with the Coraza SPOA
    http-request deny deny_status 504 if { var(txn.coraza.error) -m int gt 0 }
    http-response deny deny_status 504 if { var(txn.coraza.error) -m int gt 0 }
```

**Expected behavior**

Working backend config OOTB

**Steps to reproduce the problem**

0. Prepage HaProxy deployment (via Helm chart)
1. Configure Coraza modsecurity according to instructions: https://haproxy-ingress.github.io/docs/examples/modsecurity/
    - It also requires to use an up to date config reference https://github.com/corazawaf/coraza-spoa
3. Configure Ingress with appropriate annotations 

**Environment information**

HAProxy Ingress version: `v0.14`

Global options:

```
controller:
  config:
    modsecurity-use-coraza: "true"
[...]

PR #1094, opened 2024-03-14
fix coraza configuration to use the action variable

JanHolger:
This is a fix for #1085 

The snippet used in the fix was provided by the issue's op (@tomklapka)

It uses txn.coraza.action instead of the txn.coraza.fail and allow for a more fine grained control about how requests should be rejected and fixes the issue that with the current coraza version all requests are rejected (txn.coraza.fail is most likely not set anymore).

I've based this fix of 0.14 because I wasn't able to get master running in our testing environment. Please let me know if I need to change anything here.

### DIFF
--- a/pkg/haproxy/instance_test.go
+++ b/pkg/haproxy/instance_test.go
@@ -5090,8 +5090,14 @@ spoe-message check-request
 			endpoints: []string{"10.0.0.101:12345"},
 			backendExp: `
     filter spoe engine modsecurity config /etc/haproxy/spoe-modsecurity.conf
+    http-request redirect code 302 location %[var(txn.coraza.data)] if { var(txn.coraza.action) -m str redirect }
+    http-response redirect code 302 location %[var(txn.coraza.data)] if { var(txn.coraza.action) -m str redirect }
+    http-request deny deny_status 403 hdr waf-block "request"  if { var(txn.coraza.action) -m str deny }
+    http-response deny deny_status 403 hdr waf-block "response" if { var(txn.coraza.action) -m str deny }
+    http-request silent-drop if { var(txn.coraza.action) -m str drop }
+    http-response silent-drop if { var(txn.coraza.action) -m str drop }
     http-request deny deny_status 504 if { var(txn.coraza.error) -m int gt 0 }
-    http-request deny if !{ var(txn.coraza.fail) -m int eq 0 }`,
+    http-response deny deny_status 504 if { var(txn.coraza.error) -m int gt 0 }`,
 			modsecExp: `
     timeout connect 1s
     timeout server  2s
--- a/rootfs/etc/templates/haproxy/haproxy.tmpl
+++ b/rootfs/etc/templates/haproxy/haproxy.tmpl
@@ -595,8 +595,14 @@ backend {{ $backend.ID }}
 {{- if eq $waf.Mode "deny" }}
 {{- range $pathIDs := $wafCfg.PathIDs $i }}
 {{- if $global.ModSecurity.UseCoraza }}
+    http-request redirect code 302 location %[var(txn.coraza.data)] if { var(txn.coraza.action) -m str redirect }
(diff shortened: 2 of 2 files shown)


jcmoraisjr:
Thanks for the contribution! A few questions:
- Why did you add both the http-request and http-response? The former should be used if you want the proxy to block requests to you server, or modify how the request is changed before reach your server, and the later is used to change the response your server made before being delivered to the client/user. In the case of denies, redirects and drops, we're always talking about blocking when using in requests, so without the request reaching the underlying server, which apparently makes  the http-response useless.
- What sort of problems did you have with master? Maybe you found another incompatibility what worth to be improved in code, or better documented as a known incompatibility. Despite of that, that part of the code didn't change (that much) from v0.14 so I'd say it's safe to be sent to master. We'll probably cherry-pick it to v0.14 as well, just need to better understand the impact.

h2o-commits:
> Thanks for the contribution! A few questions:
> 
>     * Why did you add both the http-request and http-response? The former should be used if you want the proxy to block requests to you server, or modify how the request is changed before reach your server, and the later is used to change the response your server made before being delivered to the client/user. In the case of denies, redirects and drops, we're always talking about blocking when using in requests, so without the request reaching the underlying server, which apparently makes  the http-response useless.
> 

Not tested, but I thing you get the point. http-response is useless then.

>     * What sort of problems did you have with master? Maybe you found another incompatibility what worth to be improved in code, or better documented as a known incompatibility. Despite of that, that part of the code didn't change (that much) from v0.14 so I'd say it's safe to be sent to master. We'll probably cherry-pick it to v0.14 as well, just need to better understand the impact.

The problem with master has nothing to do with this fix. We were not able to test this fix on master (v0.15) as haproxy wasn't able to be bind to network socket there.

JanHolger:
> * Why did you add both the http-request and http-response? The former should be used if you want the proxy to block requests to you server, or modify how the request is changed before reach your server, and the later is used to change the response your server made before being delivered to the client/user. In the case of denies, redirects and drops, we're always talking about blocking when using in requests, so without the request reaching the underlying server, which apparently makes  the http-response useless.

Like I mentioned, we used the snippet from the issue's op and just verified that it doesn't do anything harmful and is actually working but we didn't really check what he did in the snippet. If I check the official coraza docs (https://coraza.io/connectors/coraza-spoa/#haproxy-configuration) they also do it for both request and response and according to the flow diagram on their website (https://owasp.org/www-project-coraza-web-application-firewall/) coraza will not only check the request but also the response.

> * What sort of problems did you have with master? Maybe you found another incompatibility what worth to be improved in code, or better documented as a known incompatibility. Despite of that, that part of the code didn't change (that much) from v0.14 so I'd say it's safe to be sent to master. We'll probably cherry-pick it to v0.14 as well, just need to better understand the impact.

The first issue was with our configuration and the newer haproxy version used in master (the http-use-htx option was removed). Since this option is the default value in newer versions, I was able to fix that by simply removing the option from our config.

The second issue (that I wasn't able to resolve) is a permission denied error while trying to bind 80/443, even after manually adding the NET_BIND_SERVICE capability. Do we need to change anything besides replacing the image version to make the new version work?

I've switched the base to master

JanHolger:
> > permission denied error while trying to bind 80/443, even after manually adding the NET_BIND_SERVICE capability
> 
> This is somewhat expected since we started to run haproxy ingress as non root on v0.15. How did you add the capability? Are you running the embedded haproxy or configuring an external one via a sidecar?

I guess you are already providing a fix for this in #1096?

jcmoraisjr:
> I guess you are already providing a fix for this in https://github.com/jcmoraisjr/haproxy-ingress/pull/1096?

This is more like a "best effort" than a fix, because capabilities rely on extended attributes, and it doesn't work properly on all file systems and controller runtime environments. I'll provide a proper doc and link in the v0.15 release notes, with all the options and their issues.

jcmoraisjr:
Lgtm, merging now.

PR #1116, opened 2024-05-01
Fix label generation for node discovery

jcmoraisjr:
A set of labels are used to find all the nodes of a cluster of ingress controllers. The controller instance running the code takes its own set of labels and use it to find the others via apiserver. The set of labels has, however, some labels that uniquely identify a pod, like pod-template-hash created by the ReplicaSet, making the node discovery to discover only one node, the one being running. On shutdowns, this makes the controller to incorrectly think it is running alone, leading to remove the status of all the ingress resources. Removing all the well-known label keys that uniquely identifies a pod solves this issue.


avmnu-sng (approved):
👍

Issue #1067, opened 2024-01-22
Redirect app-root is happening before redirect https


kcnpnew:
**------------------------------**
**Description of the problem**
**------------------------------**

If in an ingress the annotations app-root and ssl-redirect (with value true) are both used at the same time the app-root redirection happens first, making an unnecessary intermediary http redirection in case the user request is http.

Some risk manager tools expect the first redirection be to https already, flagging the current behavior as a security risk

**----------------------**
**Expected behavior**
**----------------------**

When we use both annotations the first redirection to app-root should already use https when the user requests http

**-------------------**
**Actual behavior**
**-------------------**

An intermediary app-root redirection happens using http

**------------------------------------**
**Steps to reproduce the problem**
**------------------------------------**

Add both annotations to an ingress

`
ingress.kubernetes.io/ssl-redirect: 'true'
ingress.kubernetes.io/app-root: '/test'
`

and than see the redirection order with a curl

`
curl -IL http://<hostname>
`

**----------------------------**
**Environment information**
**----------------------------**

Using HAProxy Ingress version: v0.13.11

**------------**
**Remarks**
**------------**

Analyzing the problem I noticed that it is rooted in the fact that while the https redirection happens inside the backend configuration the app-root one happens in the global frontend using a mapper file to find out the redirection path

`
http-request set-var(req.rootredir) var(req.host),map_str(/etc/haproxy/maps/_front_redir_fromroot__exact.map) http-request redirect location %[var(req.rootredir)] if { path / } { var(req.rootredir) -m found }
`

Looks like to solve the problem the mapper file should be augmented with the protocol information (for exemple saving the whole url instead of only the path in this case) if an https redirection annotation with value true is used. That way both redirections could happen at once, without an intermediary http one.

jcmoraisjr:
Merged a fix that will be available in the next `v0.12`, `v0.13`, `v0.14` and `v0.15` versions.

PR #1117, opened 2024-05-04
Ensure https redirect happens before root redirect

jcmoraisjr:
`app-root` config key configures the root path redirect in haproxy frontend. https redirect however is configured in the backend. Because of that haproxy is redirecting from the root path to the application path in plain http, before redirecting to https. This is not a good approach because it makes security scanners infer that the application does not have a secure proxy.

This update adds a https redirect before the application redirect, in the case the root path of the host renders its ssl-redirect to true.


jcmoraisjr:
#1067

Issue #1064, opened 2024-01-09
`ssl-always-add-https` causes `ssl-redirect` to redirect even without a TLS certificate installed


crabique:
This is a follow-up to [the Slack thread](https://kubernetes.slack.com/archives/C01D8GKKT5H/p1702555406200499) regarding this issue.

Prior to v0.13, it was the default behaviour to include every ingress host to both HTTP and HTTPS maps. 

The`ssl-redirect` to HTTPS only happened if there was an actual `tls` block in the ingress object spec, it never redirected to the implicit HTTPS map entries created for hosts without a  valid `tls` block.

In v0.14, this behaviour is no longer default and is an opt-in configuration option `ssl-always-add-https`. With it enabled, all hosts are added to both HTTP and HTTPS maps, but now enabled `ssl-redirect` causes the HTTP access to _always_ redirect to HTTPS, regardless whether a host has `tls` configuration or not.

**Expected behavior**

Enabling `ssl-always-add-https` causes the ingress controller HTTP/HTTPS redirects/frontends to work the same way as in ≤v0.12.

**Environment information**

HAProxy Ingress version: `v0.14`

Global options:

```
ssl-always-add-https: "true"
```

jcmoraisjr:
Merged and will be available in the next `v0.14` and `v0.15` versions a new config key adding the ability to change to the old behavior.

PR #1118, opened 2024-05-04
Add ssl-always-follow-redirect option

jcmoraisjr:
Controller v0.11 and older had an incorrect behavior compared with ingress spec, regarding HTTPS configuration and SSL redirect.

v0.12 added `ssl-always-add-https` option that adds the host in the HTTPS map when enabled, regardless of the `tls` attribute configuration. This was a v0.11 and older behavior.

This update adds the ability to avoid SSL redirect when TLS was added automatically. This was also a v0.11 and older behavior.


jcmoraisjr:
#1064

Issue #1073, opened 2024-01-27
Secure backends parameters cause a SIGSEGV


hkmo99:
**Description of the problem**

Adding secure backends parameters causes a SIGSEGV: segmentation violation. Is there a fix or workaround for this issue?

Logs:

```
I0127 03:39:22.152006       7 launch.go:343]
Name:       HAProxy
Release:    v0.14.5
Build:      git-7e2c578
Repository: https://github.com/jcmoraisjr/haproxy-ingress
I0127 03:39:22.152195       7 launch.go:347] watching for ingress resources with 'kubernetes.io/ingress.class' annotation: haproxy
I0127 03:39:22.152213       7 launch.go:354] watching for ingress resources with IngressClass' controller name: haproxy-ingress.github.io/controller
I0127 03:39:22.152229       7 launch.go:359] ignoring ingress resources without any class reference - --watch-ingress-without-class is false
I0127 03:39:22.152244       7 launch.go:371] watching for Gateway API resources - --watch-gateway is true
I0127 03:39:22.152262       7 launch.go:385] running embedded haproxy, mode is daemon
I0127 03:39:22.152281       7 launch.go:389] watching endpointslices - --enable-endpointslices-api is true
I0127 03:39:22.152923       7 launch.go:780] Creating API client for https://10.96.0.1:443
I0127 03:39:22.166259       7 launch.go:800] Running in Kubernetes Cluster version v1.27 (v1.27.4) - git (clean) commit fa3d7990104d7c1f16943a67f11b154b71f6a132 - platform linux/amd64
I0127 03:39:22.179037       7 launch.go:416] watching for global config options from configmap 'ingress-controller/haproxy-ingress' - --configmap was defined
I0127 03:39:22.184317       7 launch.go:521] using 2 distinct annotations prefix, with the following precedence: haproxy-ingress.github.io, ingress.kubernetes.io
W0127 03:39:22.195386       7 listers.go:170] gateway API 'networking.x-k8s.io/v1alpha1' was not found, skipping
W0127 03:39:22.198422       7 listers.go:183] watching 'gateway.networking.k8s.io/v1alpha2'
I0127 03:39:22.842507       7 listers.go:214] loading object cache...
W0127 03:39:22.857796       7 cache.go:805] error reading GatewayClass v1alpha2 'haproxy': gatewayclass.gateway.networking.k8s.io "haproxy" not found
I0127 03:39:23.150580       7 listers.go:284] cache successfully synced
[...]

PR #1119, opened 2024-05-04
Allows secure backend configuration from global

jcmoraisjr:
Some configuration keys were not designed to be made from global due to the lack of a namespace name. Secure backend needs a namespace to read secrets for a CA bundle and client certificate. This update allows to provide a secret in the global ConfigMap by adding the namespace along with the name of the secret.


jcmoraisjr:
#1073

Issue #1105, opened 2024-04-01
Global configuration for an external auth service has no effekt


dadrus:
According to the official documentation, the Ingress Controller can be configured globally using a [ConfigMap](https://haproxy-ingress.github.io/docs/configuration/keys/#configmap) to forward all requests to an [external auth](https://haproxy-ingress.github.io/docs/configuration/keys/#auth-external) service. Unfortunately, the following configuration has no effect:

```yaml
apiVersion: v1
kind: ConfigMap
data:
  auth-url: "https://<auth service name>.<namespace>.svc.cluster.local:<port>" 
  auth-headers-succeed: "authorization" 
  headers: | 
    X-Forwarded-Uri: %[pathq]
    X-Forwarded-Method: %[method]
    X-Forwarded-Host: %[req.hdr(host)]
```

Same keys used as annotations on the ingress resource work however perfectly fine:

```yaml
annotations:
  haproxy-ingress.github.io/auth-url: "https://<auth service name>.<namespace>.svc.cluster.local:<port>"
  haproxy-ingress.github.io/auth-headers-succeed: "authorization"
  haproxy-ingress.github.io/headers: |
    X-Forwarded-Uri: %[pathq]
    X-Forwarded-Method: %[method]
    X-Forwarded-Host: %[req.hdr(host)]
```

Am I doing something wrong, or is it indeed a bug?

jcmoraisjr:
Hi, this is a config limitation, a few options wasn't designed to work properly as a global config, mostly due to the missing of a namespace. Regarding external authentication it was somewhat simple to fix and I've just addressed. Should be available in the next v0.15 and v0.14 releases.

jcmoraisjr:
Fixed, merged, and the new behavior will be available in the next `v0.14` and `v0.15` versions.

PR #1120, opened 2024-05-05
Allows to configure auth-url globally

jcmoraisjr:
Adds the option to configure auth-url globally. Global configurations don't rely on a namespace name, so the only change in the configuration is that a service url needs to add the namespace of the service when configured globally.


jcmoraisjr:
#1105

PR #1127, opened 2024-05-14
Update RBAC configuration and docs to include leases resource for leader election

jzinkweg:
Same change as done in the helm chart:
https://github.com/haproxy-ingress/charts/pull/58


jcmoraisjr:
Hey this helped me to remember that we're missing to remove some really old files. Thanks for the update! Lgtm, merging now.

Issue #1109, opened 2024-04-10
Change order of use_backend haproxy.tmpl


jessequinn:
**What are you trying to do**

We require that the SNIBACKEND come before the HOSTBACKEND on the frontend configuration.

**What HAProxy Ingress should do or how it should behave differently**

```
frontend _front_https
    mode http
    bind :443 accept-proxy ssl alpn h2,http/1.1 crt-list /etc/haproxy/maps/_front_bind_crt.list ca-ignore-err all crt-ignore-err all
    log-format %ci:%cp\ method=%HM\ uri=%HU\ rcvms=%TR\ serverms=%Tr\ activems=%Ta\ bytes=%B\ status=%ST
    http-request set-var(req.path) path
    http-request set-var(req.host) hdr(host),field(1,:),lower
    http-request set-var(req.base) var(req.host),concat(\#,req.path)
    http-request set-var(req.hostbackend) var(req.base),map_reg(/etc/haproxy/maps/_front_https_host__regex.map)
    http-request set-header X-Forwarded-Proto https
    http-request del-header X-SSL-Client-CN
    http-request del-header X-SSL-Client-DN
    http-request del-header X-SSL-Client-SHA1
    http-request del-header X-SSL-Client-SHA2
    http-request del-header X-SSL-Client-Cert
    acl tls-has-crt ssl_c_used
    acl tls-need-crt ssl_fc_sni -i -m str -f /etc/haproxy/maps/_front_tls_needcrt__exact.list
    acl tls-host-need-crt var(req.host) -i -m str -f /etc/haproxy/maps/_front_tls_needcrt__exact.list
    acl tls-has-invalid-crt ssl_c_verify gt 0
    acl tls-check-crt ssl_fc_sni -i -m str -f /etc/haproxy/maps/_front_tls_auth__exact.list
    http-request set-var(req.snibase) ssl_fc_sni,lower,concat(\#,req.path)
    http-request set-var(req.snibackend) var(req.snibase),map_dir(/etc/haproxy/maps/_front_https_sni__prefix.map)
    http-request set-var(req.snibackend) var(req.base),map_dir(/etc/haproxy/maps/_front_https_sni__prefix.map) if !{ var(req.snibackend) -m found } !tls-has-crt !tls-host-need-crt
    http-request set-var(req.tls_nocrt_redir) str(_internal) if !tls-has-crt tls-need-crt
[...]

PR #1133, opened 2024-06-06
Remove dedicated maps for SNI match

jcmoraisjr:
SNI maps were incorrectly used to match requests on ancient versions of HAProxy Ingress - v0.4 or so. A separated group of match files were being used since then on TLS based authentication configurations. We don't need it anymore, since all the mTLS configurations don't depend on the maps, so we're now dropping its support. Moreover, having a distinct group of match files leads to misbehavior depending on the configurations: a host and path with lower priority should be chosen if the one with more priority is added in the sni maps.

There is one behavior change with this update: a missing or misconfigured host header, for an ingress with mTLS, with optional certificate, without sending a certificate, would fallback to SNI in order to try a match. Now, since only the host header is the source of truth, a non matching host header with a distinct SNI will 404 despite of its mTLS configuration.


jcmoraisjr:
#1109

Issue #1028, opened 2023-10-16
Premature termination of haproxy process


tumd:
**Description of the problem**

We are running haproxy-ingress in GKE, fronted by their default/legacy [target pool-based lb](https://cloud.google.com/load-balancing/docs/network/networklb-target-pools). However I do think this is likely to be an issue in other environments too.

For quite some time we have noticed short outages on our services behind haproxy-ingress when it is being redeployed.

A note about GCP LB;
GCP LB sets up health-checks performing checks with an interval of 8 seconds, and sets the backend as unhealthy after 3 failed checks. I'm not that versed in k8s, but these health-checks probes a public port on each node reporting the (termination) status of the local running haproxy-ingress.

In this caase this means that we would need haproxy-ingress to remain for at least 8*3 seconds before it terminates, to allow for the GCP LB to register the terminating instance as unhealthy.

The haproxy-ingress cmd-line option `--wait-before-shutdown` set to `30` (or something higher) would supposedly help with this issue. You would also need to set the `spec.template.spec.terminationGracePeriodSeconds` to something like `60` to allow the pod more time to gracefully terminate.

But here in lies the problem. Even with `--wait-before-shutdown=30` the haproxy _process_ inside the pod exits _immediately_ upon pod termination - even though the `haproxy-ingress-controller` process remains for 30s.

Bellow is a log from a haproxy-ingress pod that just got a termination signal.
<details>
<summary>Log</summary>

```
...
10.40.0.23:39086 [13/Oct/2023:12:49:29.249] _front_https~ app-dev_file-service_8080/srv001 0/0/0/2/2 200 237 - - ---- 6/4/0/0/0 0/0 {} {xxx} "GET https://app.xy/monitor/status HTTP/2.0"
[WARNING]  (21) : Exiting Master process...
[NOTICE]   (21) : haproxy version is 2.4.24-d175670
[ALERT]    (21) : Current worker #1 (23) exited with code 143 (Terminated)
[WARNING]  (21) : All workers exited. Exiting... (0)
I1013 12:49:29.351729       7 main.go:47] Shutting down with signal terminated
I1013 12:49:29.351798       7 controller.go:269] Waiting 1m30s before stopping components
I1013 12:49:29.444615       7 controller.go:339] starting haproxy update id=37
[...]

PR #1136, opened 2024-06-07
Move embedded haproxy process to a distinct pid group

jcmoraisjr:
The embedded haproxy process is being created in the same pid group of the controller, which makes it receive the same signals received by the controller when running under a process supervisor. This makes haproxy be prematurely terminated when the controller is still shutting down.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -44,4 +44,7 @@ RUN mkdir -p /var/empty /etc/haproxy /var/lib/haproxy /var/run/haproxy\
 
 STOPSIGNAL SIGTERM
 USER haproxy
+
+# dumb-init reaps the old haproxy process in the embedded, non master-worker mode,
+# after receiving SIGUSR1, avoiding it to become a zombie.
 ENTRYPOINT ["/usr/bin/dumb-init", "--", "/start.sh"]
--- a/pkg/haproxy/instance.go
+++ b/pkg/haproxy/instance.go
@@ -637,6 +637,7 @@ func (i *instance) startHAProxySync() {
 		"-f", i.options.HAProxyCfgDir)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 	if err := cmd.Start(); err != nil {
 		i.logger.Error("error starting haproxy: %v", err)
 		return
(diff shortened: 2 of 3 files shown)


jcmoraisjr:
#1028

avmnu-sng:
@jcmoraisjr Don't we need to update `haproxy-shutdown.sh` to send `USR1`, it still sends `TERM` which is hard stop. https://github.com/jcmoraisjr/haproxy-ingress/blob/914b58192a0a76fefc52f46d3a65a608f21ced90/rootfs/haproxy-shutdown.sh#L23
`kill -USR1 $pid`

jcmoraisjr:
Hi, in fact this shouldn't make any difference because, when shutdown.sh is called, the controller process is being shutting down as well, dropping all the other process in that same container. So moving from term to usr1 would be a noop to the observer. I understand your concern, and I also think we've already discussed that on another issue, and the point here is that the correct place to configure a long living haproxy connection during rolling updates is not here, but instead making the controller to postpone its own shutdown. Moreover, this script belongs to a way to run haproxy which is being deprecated. master-worker is the new default from v0.15, and it calls haproxy directly without any help of external scripts.

avmnu-sng:
@jcmoraisjr Yeah, we discussed it earlier. I guess I understand now why `USR1` is still a no-op. For example, suppose we have `wait-before-shutdown=10s`. Then, after 10s, irrespective of `USR1` or `TERM` on HAProxy, the ingress controller exits immediately. It would only work if we wait after triggering the HAProxy shutdown. Even if we move to master-worker, we still need a soft stop on HAProxy, and shouldn't ideally, the controller should exit after HAProxy shutdown completes.

jcmoraisjr:
I see your point. When running haproxy as a daemon we don't have a good control of the underlying process. In master worker mode we run it in the foreground, this gives us a few more options.

Here is where the process is started, stopped, and asked to be stopped:

https://github.com/jcmoraisjr/haproxy-ingress/blob/a7e4b1f02d91d919feb11d83fe3d48a09dad185d/pkg/haproxy/instance.go#L632-L667

and here is the channel that makes the controller wait until haproxy is gone:

https://github.com/jcmoraisjr/haproxy-ingress/blob/a7e4b1f02d91d919feb11d83fe3d48a09dad185d/pkg/haproxy/instance.go#L402-L423

In this scenario it makes sense to send usr1 because the controller will wait for it. I'll give this a chance, maybe adding as a config option. Tks btw for bringing this scenario again and again =)

avmnu-sng:
@jcmoraisjr No worries; finally, we are on the same page. For now, with `Setpgid: true`, running in master-worker mode with a sensible `wait-before-shutdown` value would take us closer to a soft-stop experience.

Please give the config option approach a shot; it would be helpful. Thanks!

Issue #342, opened 2019-05-21
HAProxy instance stdout/stderr go to /dev/null


smartyjohn:
(The test case is only relevant to haproxy v1.9+ [future bug/enhancement], but the underlying cause may be impacting current versions as well. )

I've been trying out haproxy v1.9.8 in a custom docker build, and it's support of new logging methods to stdout and stderr.  These do not work.  When haproxy is spawned (reloaded?) by the controller its file descriptors for stdout and stderr are set to `/dev/null`.  This can be seen from an `lsof` in the running container of the first three file descriptors for the process (0=stdin, 1=stdout, 2=stderr):
```
[...]
47817	/usr/local/sbin/haproxy	/dev/null
47817	/usr/local/sbin/haproxy	/dev/null
47817	/usr/local/sbin/haproxy	/dev/null
47817	/usr/local/sbin/haproxy	anon_inode:[eventpoll]
[...]
```

This may affect current versions as it would seem as soon as a reload occurs then all output via stdout/err from the haproxy instance would be lost.

It seems to me the correct value of stdout/stderr when the process is first spawned should either be the same as the controller's stdout/stderr or else the file descriptors from opening `/proc/1/fd/1` and `/proc/1/fd/2` respectively as those are collected by Docker (they should all point to the same endpoint anyway).  I did see the reload spawning [controller.go/OnUpdate()] and it may be as simple as setting reloadCmd.StdOut/StdErr; but that's outside my realm.

As an aside for others who may be looking for important debugging messages, you can use the existing syslog facilities or you can use socat (already present in the container) to create a datagram socket to forward to `/proc/1/fd/1` as a hackish workaround (e.g. via a manual exec shell or a modified init script--definitely not for production usage):

```
socat -u UNIX-RECVFROM:/var/run/forward-stdout.sock,fork OPEN:/proc/1/fd/1,append
```

And the ConfigMap change:
```
syslog-endpoint: /var/run/forward-stdout.sock
```

jcmoraisjr:
Hi, thanks for pointing this as this will be an issue on v0.9 - with HAProxy 1.9 and its stdout logging.

The only reason to capture stdout/err in the current code is to log any warnings after a reload which would be redirected to fd1 anyway.

aperullo:
Has this been fixed? I don't see the logs updating when a connection tries to come in

jcmoraisjr:
Hi, this is a v0.9 feature when haproxy will be updated to 1.9.

[...]

PR #1138, opened 2024-06-07
doc: add haproxy logging to stdout

jcmoraisjr:
Add info on how to configure HAProxy to log to stdout.

Remove the line with v0.14 related info when merging to older versions.


jcmoraisjr:
#342

PR #1140, opened 2024-06-08
Cleanup outdated properties of golangci-lint gh actions plugin

Spikhalskiy:
From golangci-lint-action [release notes](https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#compatibility):

> v5.0.0+ removes skip-pkg-cache and skip-build-cache because the cache related to Go itself is already handled by actions/setup-go.


jcmoraisjr:
Lgtm thanks! Merging now.

Issue #1148, opened 2024-06-20
"TCP SERVICES frontends" uses only 1 certificate


GuilhermeAbraham:
We want to have 2 tcp frontends for the same backend.

**Steps to reproduce the problem**
1. Configure a ingress for TCP services with 2 hosts and different certificates for each host
2. Execute `openssl s_client -showcerts -servername HOST -connect HOST:443 </dev/nul` for each host
3. Check only 1 certificate gets returned

**Environment information**
HAProxy Ingress version: `v0.14.7`

This is our ingress:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: "haproxy-mtls"
    haproxy-ingress.github.io/tcp-service-port: "443"
    haproxy-ingress.github.io/secure-backends: "false"
    haproxy-ingress.github.io/secure-verify-ca-secret: "namespace1/mtls-ca"
    haproxy-ingress.github.io/auth-tls-secret: "namespace1/mtls-ca"
    haproxy-ingress.github.io/auth-tls-verify-client: "on"
    haproxy-ingress.github.io/ssl-options-backend: ""
    haproxy-ingress.github.io/tcp-service-proxy-protocol: "v2"
    haproxy-ingress.github.io/backend-check-interval: "2s"
    haproxy-ingress.github.io/proxy-protocol: "v2"
    haproxy-ingress.github.io/health-check-port: "1885"
  name: mtls
  namespace: namespace1
spec:
  rules:
    - host: mtls.domain.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: do-stuff
                port:
                  name: mqtt-do-stuff
    - host: mtls.old.domain.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: do-stuff
                port:
                  name: mqtt-do-stuff
  tls:
    - hosts:
        - mtls.domain.com
      secretName: root
    - hosts:
        - mtls.old.domain.com
      secretName: mqtt
```

This is how the configuration looks like:
```  # # # # # # # # # # # # # # # # # # #
# #
#     TCP SERVICES frontends
#
frontend _front_tcp_443
    bind :443 ssl crt /var/lib/haproxy/crt/root.pem ca-file /var/lib/haproxy/cacerts/mtls-ca.pem verify required
    mode tcp
    option tcplog
[...]

PR #1171, opened 2024-09-22
Support list of server crt on tls tcp service

jcmoraisjr:
Add the ability to configure more than one server certificate on TCP services.


jcmoraisjr:
#1148

Issue #1175, opened 2024-10-15
Haproxy ingress doesn't report correct error messages or have a retry when an reload issue happens


danjiawork:
**Description of the problem**

We noticed that when haproxy reload faces an issue, especially when multiple reloads happen at the same time
-> haproxy ingress logs show that the communication to the master is refused:
`error reloading server: error sending reload to master socket: error connecting to /var/run/haproxy/master.sock: dial unix /var/run/haproxy/master.sock: connect: connection refused`
but also: 
`10 controller.go:366] finish haproxy reload id=2: reload_haproxy=5000.375786ms total=5000.375786ms`, 
which is very confusing, sounds like the haproxy reload finished successfully in 5 sec. 
But actually, the reload was not successful and there was no retry.
-> In this case, the failed/latest haproxy config update cannot be triggered successfully, they are not able to reach out to the backends, until the next reload succeeds again.

**Expected behavior**

- Once a newer haproxy reload is refused because the last haproxy reload has not finished, the failed reload should be reported and a retry should be triggered between [the lines](https://github.com/jcmoraisjr/haproxy-ingress/blob/release-0.14/pkg/controller/controller.go#L365-L366). There could be some messages like `haproxy reload failed, will retry again`
- Making the [timeout setting for sending the socket](https://github.com/jcmoraisjr/haproxy-ingress/blob/release-0.14/pkg/haproxy/socket/socket.go#L52) configurable 

**Steps to reproduce the problem**

In our cases, the reload issues happen normally when a second reload is triggered when another reload takes longer to get a response and has not finished...
Therefore, we reproduce the issue by:
1. Delaying the startup of each new haproxy process (for example, by adding a sleep statement [here](https://github.com/haproxy/haproxy/blob/v2.9.0/src/haproxy.c#L3470)
2. Tigger the first reload process, and before it finishes, trigger the second haproxy reload
3. haproxy reload logs then show `error reloading server: error sending reload to master socket: error connecting to /var/run/haproxy/master.sock: dial unix /var/run/haproxy/master.sock: connect: connection refused`, and the haproxy connections cannot be routed to the backends, until a following reload succeeds again

**Environment information**

HAProxy Ingress version: `v0.14` and HAProxy version we are using: `v2.9.0`

Command-line options:

```
(paste command-line options here)
```

Global options:

```
(paste global configmap options here)
```

Ingress objects:

```
(paste ingress objects here)
```

jcmoraisjr:
[...]

PR #1214, opened 2025-03-09
retry reload haproxy if failed

jcmoraisjr:
HAProxy Ingress does not use to restart a failed operation because the majority of the failed operations come from misconfiguration, so the configuration fix itself will make the retry.

Former HAProxy Ingress had only embedded and non master-worker mode, so other failures despite misconfigurations would be resource constraint related, like disk space, or bugs in haproxy or controller. Since master-worker mode and external haproxy, other sort of failures might happen, like connection refused to the admin socket. Because of that, a retry operation becomes mandatory, avoiding healthy but outdated load balancer config.


jcmoraisjr:
#1175

Issue #1284, opened 2025-08-11
determine controller pods regardless if using Deployment or StatefulSet


kro-cat:
**What are you trying to do**

I would like for the haproxy-ingress controller to determine controller pods regardless if deploying with a `Deployment` or `StatefulSet`.  However, this feature may require supporting the use of a `StatefulSet`.  Currently, in version 0.15, the project relies on the haproxy-ingress controller pods to be created using a `Deployment` in order to determine the selector labels.

**What HAProxy Ingress should do or how it should behave differently**

The selector labels are determined by obtaining the current pod's labels and deleting unique pod labels: [release-0.15, /pkg/controller/services/svcstatusing.go, line 254](https://github.com/jcmoraisjr/haproxy-ingress/blob/release-0.15/pkg/controller/services/svcstatusing.go#L254).  But the `StatefulSet` pod-name label (`statefulset.kubernetes.io/pod-name`) isn't removed here.

[The func getControllerPodList() in /pkg/controller/services/svcstatusing.go](https://github.com/jcmoraisjr/haproxy-ingress/blob/release-0.15/pkg/controller/services/svcstatusing.go#L244) may either be amended to include this label, or a different method to determine selector labels may be used.

In my local fork of this func, I've opted to determine the selector labels from the metadata.ownerReferences field.  For `Deployments`, however, this would require querying the `ReplicaSet` first, then the `Deployment`.  In my version, I set up a loop to jump through ownerReferences and find the best candidate for which selector labels to use.  It stops at the last ownerReference with the `controller: true` attribute which has a selector spec (currently only supports `StatefulSet`, `ReplicaSet`, or `Deployment`).  It's probably a complicated solution for a simple problem, but I like the flexibility in this way.

kro-cat:
This is maybe going to be a little confusing.  This example code is based on v0.14.9, which is the version I'm using locally.  The GetControllerPodList was backported to my local branch, but the logic is almost exactly the same.  I wanted to include this so that you can have a good idea of what I'm trying to describe in the feature request.

```go
func (c* k8scache) getPodSelectorLabels(pod *api.Pod) (labels map[string]string, err error) {
	if pod == nil {
		err = fmt.Errorf("pointer is nil")
		return nil, err
	}

	var meta metav1.Object = pod.GetObjectMeta()
	var namespace string = meta.GetNamespace()
	var podLabels map[string]string = meta.GetLabels()

[...]

PR #1288, opened 2025-08-13
read controller pod selector from owner

jcmoraisjr:
Controller pod selector is used to list all pods that form the ingress cluster, making it possible to find all the nodes where ingress controller is running (used by default on ingress status) and peers.

Historically ingress controller uses pod labels to find all the other controller pods, but this approach can lead to an inaccurate result in the case some of the labels uniquely identify the pod.

This update changes the approach to use the owner selector instead, being either a DaemonSet or a ReplicaSet. The owner will always have an accurate selector to find all their pods.


jcmoraisjr:
#1284

PR #1289, opened 2025-08-13
fix label selector from daemonset

jcmoraisjr:
Label selector is used to filter pods and find all the replicas of an ingress cluster. DaemonSet could add other labels that uniquely identify a pod, making the returned list inaccurate.

Should be merged to v0.15.


jcmoraisjr:
#1284

jcmoraisjr:
Closing on behalf of #1288

Issue #774, opened 2021-04-30
Add Gateway API support


jcmoraisjr:
**What are you trying to do**

Configure HAProxy using [Gateway API](https://gateway-api.sigs.k8s.io/) resources.

**What HAProxy Ingress should do or how it should behave differently**

Build HAProxy configuration using Gateway API resources, along with Ingress, in order to preserve backward compatibility.

jcmoraisjr:
Hi, v0.13 partially supports gateway api, v0.14 updates its version but with the same support, v0.15 should have a new api version update and fully support it. Unfortunately, due to some other tasks running in parallel, I cannot guarantee when v0.15 starts beta - this should be happening now but I still have some confidence that can happen later this year.

jcmoraisjr:
We are finally Gateway API compliant, closing this issue.

Issue #369, opened 2019-08-17
Path scoped binding to HTTP/s port


dimovnike:
There are cases when multiple HTTP ports are needed. 
Currently this is not possible (except with a custom template and hardcoded secondary ports).

Some ideas of how this could work (borrowed from docker proxy flow):
1. Configure haproxy with bind ports via ConfigMap
2. Configure each ingress with source port via annotations.

But even a simple config http(s)-ports-extra which would duplicate the main port, would be good for a start.

jhohertz:
This would be very useful for cases of very high densities of connections, and as far as I can tell, no other ingress controller so far supports doing so.

jcmoraisjr:
Hi, multi port is already supported, doc [here](http://haproxy-ingress.github.io/docs/configuration/keys/#bind). What do you mean is to bind some ingress to some listening ports - eg listening 8443 and 9443, but only 8443 would find the app while 9443 would find another one or 404? Could you describe a bit more about this use case and why/how the current codebase doesn't work?

jcmoraisjr:
A starting implementation was just merged, to be available in the first v0.17 version.

PR #1295, opened 2025-08-24
configure more than one http/s ports

jcmoraisjr:
Adds the ability to configure the HTTP/s frontends with more than one HTTP port, allowing to multiplex distinct groups of hostnames without colliding each other. This is a pre-requisite to improve Gateway API support.

Several places of HAProxy Ingress code uses the dual and plain HTTP/s model as a premise, so this is a non trivial refactor. Starting the refactor from the top of the internal HAProxy model representation.

TODO list includes:

* [x] Move Hosts and Frontend types under a new Frontends type
* [x] Improve PathLink type to support distinct Frontends
* [ ] Move Bind configurations specific to frontends under the Frontend type
* [ ] Split the single HTTP/s frontend representation into two separated HTTP and HTTPs entries
* [ ] Support a list of HTTP/s frontends
* [ ] Add frontends support into Ingress API
* [ ] Add listener.Port support into Gateway API

### DIFF
--- a/pkg/converters/gateway/gateway.go
+++ b/pkg/converters/gateway/gateway.go
@@ -533,8 +533,8 @@ func (c *converter) createHTTPHosts(routeSource *source, hostnames []gatewayv1.H
 			if hstr == "" || hstr == "*" {
 				hstr = hatypes.DefaultHost
 			}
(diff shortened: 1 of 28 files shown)


jcmoraisjr:
#369
#774

Issue #1294, opened 2025-08-24
The `secure-crt-secret` and `secure-verify-ca-secret` no longer accept `file://` path after v0.14.7


chitoku-k:
**Description of the problem**

The `secure-crt-secret` and `secure-verify-ca-secret` no longer support `file://`-style paths after https://github.com/jcmoraisjr/haproxy-ingress/commit/51b2f317c39b7924b44bedb6d2afe9ded6ccbc26 because newly-added `NamespacedName()`  returns an error for file path and never calls `GetTLSSecretPath()` or `GetCASecretPath()`.

_This issue was separated from https://github.com/jcmoraisjr/haproxy-ingress/issues/1255._

**Expected behavior**

The `file://`-style paths work fine in `secure-crt-secret` and `secure-verify-ca-secret` as documented.

**Steps to reproduce the problem**

1. Configure `secure-verify-ca-secret` to `file:///var/run/secrets/kubernetes.io/serviceaccount/ca.crt`
2. The configuration is ignored with the following log output:
   ```
   skipping CA on Ingress 'monitor/prometheus': unpexpected format for resource name: file:///var/run/secrets/kubernetes.io/serviceaccount/ca.crt
   ```
   There is also a typo in the message above: `unpexpected`.

**Environment information**

HAProxy Ingress version: `v0.14.7`

Command-line options:

```
--configmap=$(POD_NAMESPACE)/haproxy-ingress
--default-ssl-certificate=$(POD_NAMESPACE)/ingress-controller-default-server-certificate
--disable-api-warnings
--master-worker
--sort-backends
```

Global options:

```yaml
bind-http: :80,:::80
bind-https: :443,:::443
bind-ip-addr-tcp: '::'
cross-namespace-secrets-ca: allow
```

Ingress objects:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    ingress.kubernetes.io/backend-protocol: h2-ssl
    ingress.kubernetes.io/health-check-uri: GET /-/healthy
    ingress.kubernetes.io/secure-verify-ca-secret: file:///var/run/secrets/kubernetes.io/serviceaccount/ca.crt
  name: prometheus
  namespace: monitor
spec:
  ingressClassName: haproxy-ingress
  rules:
  - host: prom.example.com
    http:
      paths:
      - backend:
          service:
            name: prometheus
            port:
              number: 9090
        path: /
        pathType: Prefix
  tls:
[...]

PR #1297, opened 2025-08-28
fix reading backend ca certificate from file

jcmoraisjr:
TLS backend can be configured by adding a CA secret (verify server's certificate) and/or a Crt secret (server verify client's certificate). A former controller version was reading the namespace of a secret from the ingress that configures the annotation, leading to nil dereference when the value is configured globally. This was fixed by adding a new step, that verifies if the config is global or an annotation, and if the provided name has already a name or just the name of the secret.

This new step however broke the proto://content approach added on the get secret code because it happens before getting the secret, and the proto://content syntax is incompatible with the ns/name one. This failed deployments that read ca or crt from the file system.

Should be merged as far as v0.13.


jcmoraisjr:
#1294

Issue #1351, opened 2025-12-22
haproxy reload got stuck


asHasnain:
**Description of the problem**

During the HAProxy worker process reload, the HAProxy master process unexpectedly stopped with the exit code -1. The reload was completed by the new master process but the HAProxy ingress controller appears to be stuck in the [infinite reconnection loop](https://github.com/jcmoraisjr/haproxy-ingress/blob/v0.14.10/pkg/haproxy/instance.go#L766) because the current reload count is reset to 0 (from 2307):

```
1766131800271	2025-12-19T08:10:00.271Z	[NOTICE]   (73) : Reloading HAProxy
1766131800405	2025-12-19T08:10:00.405Z	[NOTICE]   (73) : New worker (23091) forked
1766131800405	2025-12-19T08:10:00.405Z	[NOTICE]   (73) : Loading success.
1766131800428	2025-12-19T08:10:00.428Z	I1219 08:10:00.428582      30 instance.go:416] haproxy successfully reloaded (embedded master-worker)
1766131815494	2025-12-19T08:10:15.494Z	I1219 08:10:15.494453      30 instance.go:379] haproxy reload enqueued
1766131817495	2025-12-19T08:10:17.495Z	I1219 08:10:17.495295      30 instance.go:379] haproxy reload enqueued
1766131833061	2025-12-19T08:10:33.061Z	I1219 08:10:33.061584      30 instance.go:379] haproxy reload enqueued
1766131835065	2025-12-19T08:10:35.065Z	I1219 08:10:35.065197      30 instance.go:379] haproxy reload enqueued
1766131849261	2025-12-19T08:10:49.261Z	I1219 08:10:49.261764      30 instance.go:379] haproxy reload enqueued
1766131857673	2025-12-19T08:10:57.673Z	I1219 08:10:57.673085      30 instance.go:379] haproxy reload enqueued
1766131860270	2025-12-19T08:11:00.270Z	[NOTICE]   (73) : Reloading HAProxy
1766131860402	2025-12-19T08:11:00.402Z	[NOTICE]   (73) : New worker (23101) forked
1766131860402	2025-12-19T08:11:00.402Z	[NOTICE]   (73) : Loading success.
1766131860426	2025-12-19T08:11:00.426Z	I1219 08:11:00.426038      30 instance.go:416] haproxy successfully reloaded (embedded master-worker)
[...]

PR #1355, opened 2025-12-23
check if process is the same waiting for reload

jcmoraisjr:
HAProxy up to 2.6 runs the reload command async, so we need to wait until we identify that haproxy finished reloading, succeeding or not. This is done by checking if the number of reloads increased. We however does not check if the process continues to be the same, in that case we'd see the reload count decreasing. This update checks not only the number of reloads, but also the process ID of the previous and the current HAProxy instance.


jcmoraisjr:
#1351

PR #1381, opened 2026-02-18
fix: report all node IPs in status for dual-stack support

mia-mouret:
## Problem

On dual-stack Kubernetes clusters, nodes have multiple addresses of the same type in `node.Status.Addresses` — typically one IPv4 and one IPv6 `InternalIP` (or `ExternalIP`).

The `getNodeIPs` function in `svcaddress.go` only returns the **first** matching address per node, because the inner anonymous function uses an early `return` on the first match:

```go
ipnode := func() string {
    for _, addr := range node.Status.Addresses {
        if s.cfg.UseNodeInternalIP && addr.Type == api.NodeInternalIP {
            return addr.Address // ← stops here, misses IPv6
        }
    }
    return ""
}()
```

This means that when using `--report-node-internal-ip-address`, the Ingress and Gateway status only contains IPv4 (or only IPv6, depending on which is listed first), preventing tools like external-dns from creating both A and AAAA records.

## Fix

Collect **all** addresses of the matching type (`InternalIP` or `ExternalIP`) from each node, using a `seen` map for deduplication. This is a minimal, backward-compatible change:

- Single-stack clusters: behavior is unchanged (one IP per node, as before)
- Dual-stack clusters: both IPv4 and IPv6 are now included in the status

## Testing

- `go vet ./pkg/controller/services/` passes
- No existing unit tests for `getNodeIPs` (the function is unexported and has no test file)

## Context

This enables the common bare-metal pattern of:
1. haproxy-ingress as DaemonSet with `hostNetwork: true`
2. `--report-node-internal-ip-address` to populate Ingress status
3. external-dns `--source=ingress` creating both A and AAAA records

### DIFF
--- a/pkg/controller/services/svcaddress.go
+++ b/pkg/controller/services/svcaddress.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"sort"
 
 	"github.com/go-logr/logr"
@@ -277,7 +278,6 @@ func (s *svcAddress) getNodeIPs(ctx context.Context) []string {
 	}
 	// read node IPs where the controller replicas are running
 	var iplist []string
-	seen := make(map[string]bool)
 	for _, ctr := range podList {
 		node := api.Node{}
 		if err := s.cli.Get(ctx, types.NamespacedName{Name: ctr.Spec.NodeName}, &node); err != nil {
@@ -288,8 +288,7 @@ func (s *svcAddress) getNodeIPs(ctx context.Context) []string {
 			if addr.Address == "" || addr.Type != targetType {
 				continue
 			}
-			if !seen[addr.Address] {
-				seen[addr.Address] = true
+			if !slices.Contains(iplist, addr.Address) {
 				iplist = append(iplist, addr.Address)
 			}
 		}


pkg/controller/services/svcaddress.go
```diff
@@ -261,46 +261,37 @@ func (s *svcAddress) readCurrentLB(ctx context.Context) (lb []networking.Ingress
...
+				continue
 			}
-			return ""
-		}()
-		exists := func() bool {
-			for _, ip := range iplist {
-				if ip == ipnode {
-					return true
-				}
+			if !seen[addr.Address] {
```
jcmoraisjr:
```suggestion
			if !slices.Contains(iplist, addr.Address) {
```

jcmoraisjr (commented):
Thank you for the PR! Your refactor looks great.

Just added a minor suggestion that avoids allocating a hashmap. The list should always be pretty small so a full scan shouldn't be a problem. What do you think.

mia-mouret:
Good call — applied! Much cleaner without the map. Thanks for the quick review!

jcmoraisjr:
Thank you, merging now. I'll take of backporting it as far as v0.15.

PR #1385, opened 2026-02-24
fix: move logic to capture request origin header earlier

### LINKED ISSUE #1384: Wildcard allow origin in custom error page

I have set a global `proxy-body-size` limit in the configmap, and I have one specific API service for which I want to override the 413 error document to be JSON rather than HTML with an ingress annotation:

```
    haproxy-ingress.github.io/http-response-413: |
      content-type: application/json
      
      {"message":"Request body too large"}
```

This works correctly (with 0.16 beta2), however I also need to handle CORS with `Access-Control-Allow-Credentials: true`, so I add annotations

```
    haproxy-ingress.github.io/cors-enable: "true"
    haproxy-ingress.github.io/cors-allow-credentials: "true"
    haproxy-ingress.github.io/cors-allow-origin-regex: ".*"
    haproxy-ingress.github.io/cors-expose-headers: "Content-Length,X-My-Custom-Header"
```

The allow-origin-regex correctly reflects back the caller's origin for successful requests

```
access-control-allow-origin: https://example-origin.com
access-control-allow-methods: GET, PUT, POST, DELETE, PATCH, OPTIONS
access-control-allow-headers: DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization
access-control-allow-credentials: true
access-control-expose-headers: Content-Length,X-My-Custom-Header
vary: Origin,Origin

{"response":{"hello":"world"}}
```

but when I try to send a too-large request I get the following with the 413 response:

```
access-control-allow-origin: *
access-control-allow-methods: GET, PUT, POST, DELETE, PATCH, OPTIONS
access-control-allow-headers: DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization
access-control-allow-credentials: true
access-control-expose-headers: Content-Length,X-My-Custom-Header
vary: Origin

{"message":"Request body too large"}
```

The other `access-control-*` headers are still added, but the `access-control-allow-origin` is `*` rather than a reflection of the caller's submitted `Origin` header (which is the same in both examples), which is not permitted when `access-control-allow-credentials` is true.

My guess is that the `lua.send-413` in the backend config is kicking in before the `set-var(txn.hdr_origin0)` has a chance to capture the incoming origin value; is there a way to reverse that ordering or is it simply impossible to capture request headers if we've broken the flow with a 413?

**Expected behavior**

When access-control-allow-credentials is set, error responses should reflect the client origin in exactly the same way as successful responses do.

**Environment information**

HAProxy Ingress version: `v0.16.0-beta2`

Command-line options:

```
"default-ssl-certificate": "{my-ns}/{my-wildcard-certificate}"
```

Global options:

```
proxy-body-size: "10m"
cross-namespace-secrets-ca: "allow"
cross-namespace-services: "allow"
```

Ingress objects:

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    haproxy-ingress.github.io/cors-enable: "true"
    haproxy-ingress.github.io/cors-allow-credentials: "true"
    haproxy-ingress.github.io/cors-allow-origin-regex: ".*"
    haproxy-ingress.github.io/cors-expose-headers: "Content-Length,X-My-Custom-Header"
    haproxy-ingress.github.io/http-response-413: |
      content-type: application/json
      
      {"message":"Request body too large"}
    haproxy-ingress.github.io/http-response-504: |
      content-type: application/json
      
      {"message":"No response from upstream service"}
    haproxy-ingress.github.io/timeout-connect: "180s"
    haproxy-ingress.github.io/timeout-server: "180s"
    haproxy-ingress.github.io/path-type: "regex"
  name: ingress-proxy-haproxy
  namespace: default
spec:
  ingressClassName: haproxy
  rules:
  - host: api-proxy.example.com
    http:
      paths:
      - backend:
          service:
            name: api-proxy
            port:
              number: 8081
        path: /process(?:-document)?/.*
        pathType: ImplementationSpecific
  tls:
  - hosts:
    - api-proxy.example.com
```

ianroberts:
Move the logic that captures the incoming request Origin header into the `txn.hdr_origin0` variable so that it comes before the directives that handle MaxBodySize.  The `http-request use-service` directive that applies if the body size exceeds the configured maximum has the effect of disregarding any other `http-request` directives that come after it in the configuration, which in turn meant that the generated 413 response would say `Access-Control-Allow-Origin: *` even in cases where a successful request would reflect back the client's `Origin` header.  Moving the variable capture before the `lua.send-413` means the CORS headers are consistent between successful and 413-error responses.

Fixes #1384

### DIFF
--- a/rootfs/etc/templates/haproxy/haproxy.tmpl
+++ b/rootfs/etc/templates/haproxy/haproxy.tmpl
@@ -649,6 +649,23 @@ backend {{ $backend.ID }}
 {{- end }}
 
 {{- /*------------------------------------*/}}
+{{- $corsCfg := $backend.PathConfig "Cors" }}
+{{- range $i, $cors := $corsCfg.Items }}
+{{- if and $cors.Enabled $cors.AllowOrigin }}
+{{- $corsNeedACL := or $cors.AllowOriginRegex (gt (len $cors.AllowOrigin) 1) (ne (index $cors.AllowOrigin 0) "*") }}
+{{- if $corsNeedACL }}
+    http-request set-var(txn.hdr_origin{{ $i }}) req.hdr(Origin)
+{{- end }}
+{{- range $pathIDs := $corsCfg.PathIDs $i }}
+    http-request set-var(txn.cors_max_age) str({{ $cors.MaxAge }}) if METH_OPTIONS
+        {{- if $pathIDs }} { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
+    http-request use-service lua.send-cors-preflight if METH_OPTIONS
+        {{- if $pathIDs }} { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{- /*------------------------------------*/}}
 {{- $maxbodyCfg := $backend.PathConfig "MaxBodySize" }}
 {{- if and $backend.CustomHTTPResponses.Lua $maxbodyCfg.Items }}
     http-request set-var(txn.lua_scope) str({{ $backend.ID }})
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
lgtm, thank you! merging now.

PR #1386, opened 2026-02-24
fix: move logic to capture request origin header earlier (backport 0.16)

ianroberts:
Move the logic that captures the incoming request Origin header into the txn.hdr_origin0 variable so that it comes before the directives that handle MaxBodySize.  The `http-request use-service` directive that applies if the body size exceeds the configured maximum has the effect of disregarding any other `http-request` directives that come after it in the configuration, which in turn meant that the generated 413 response would say `Access-Control-Allow-Origin: *` even in cases where a successful request would reflect back the client's `Origin` header.  Moving the variable capture before the `lua.send-413` means the CORS headers are consistent between successful and 413-error responses.

This is a backport of the same fix from #1385 into the 0.16 branch.

### DIFF
--- a/rootfs/etc/templates/haproxy/haproxy.tmpl
+++ b/rootfs/etc/templates/haproxy/haproxy.tmpl
@@ -625,6 +625,23 @@ backend {{ $backend.ID }}
 {{- end }}
 
 {{- /*------------------------------------*/}}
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
lgtm thanks!

PR #1389, opened 2026-02-26
Reflect back request origin when credentials enabled

### LINKED ISSUE #1388: CORS: reflect back request origin when allow-credentials is true
**What are you trying to do**

Currently when you enable CORS with the default wildcard allowed origin, and enable credentials:

```
haproxy-ingress.github.io/cors-enable: "true"
haproxy-ingress.github.io/cors-allow-credentials: "true"
```

the response returns a literal `Access-Control-Allow-Origin: *` alongside `Access-Control-Allow-Credentials: true`. This is invalid per the CORS spec, when allowing credentials the ACAO header must name the specific origin rather than using `*`.

The current workaround is to also specify a match-everything regex:

```
haproxy-ingress.github.io/cors-allow-origin-regex: ".+"
```

since a regex allow rule will trigger the logic that reflects back the request `Origin` instead of just using `*`.

**What HAProxy Ingress should do or how it should behave differently**

There should be a special case that when we allow any origin but also want to allow credentials, then we reflect back the supplied origin instead of returning a literal `*`.

ianroberts:
When CORS is configured with the default "allow any origin" but also with cors-allow-credentials set to true, we need to reflect back the `Origin` value from the request rather than returning a literal `Access-Control-Allow-Origin: *`, since the CORS spec does not permit `Access-Control-Allow-Credentials: true` to be combined with a wildcard origin.

In order to implement this reliably I've had to change the logic so that CORS headers are only added to the response if the request included an `Origin` header.  If the request has no `Origin` then no `Access-Control-*` headers are included in the response.  This will not break anything, since by definition a request without an `Origin` is not using CORS.

Fixes #1388

### DIFF
--- a/rootfs/etc/templates/haproxy/haproxy.tmpl
+++ b/rootfs/etc/templates/haproxy/haproxy.tmpl
@@ -818,10 +818,8 @@ backend {{ $backend.ID }}
     http-response set-header Access-Control-Allow-Origin
         {{- if or $dynamicOrigin $wildcardWithCreds }}  "%[var(txn.hdr_origin{{ $i }})]"{{ else }}  "{{ index $cors.AllowOrigin 0 }}"{{ end }}
         {{- "" }} if cors_has_origin{{ $i }}
-        {{- if or $dynamicOrigin $pathIDs }}
-            {{- if $pathIDs }} { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
-            {{- if $dynamicOrigin }} cors_allow_origin{{ $i }}{{ end }}
-        {{- end }}
+        {{- if $pathIDs }} { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
+        {{- if $dynamicOrigin }} cors_allow_origin{{ $i }}{{ end }}
 {{- if $dynamicOrigin }}
     http-response set-header Access-Control-Allow-Origin  "{{ index $cors.AllowOrigin 0 }}"
         {{- "" }} if cors_has_origin{{ $i }}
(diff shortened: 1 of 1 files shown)


rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -817,38 +816,44 @@ backend {{ $backend.ID }}
...
     http-response set-header Access-Control-Expose-Headers "{{ $cors.ExposeHeaders }}"
-        {{- if $pathIDs }} if { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
+        {{- "" }} if cors_has_origin{{ $i }}
+        {{- if $pathIDs }} { var(txn.pathID) -m str {{ $pathIDs }} }{{ end }}
 {{- end }}
-{{- if $dynamicOrigin }}
+{{- if or $dynamicOrigin $wildcardWithCreds }}
     http-response set-header Vary %[res.fhdr(Vary)],Origin
-        {{- "" }} if { res.hdr(Vary) -m found }
+        {{- "" }} if { res.hdr(Vary) -m found } ! { res.hdr(Vary) -m str "Origin" }
```
ianroberts: This just tidies up a case I spotted where I was getting `Vary: Origin,Origin` - now we only add Origin to the `Vary` header if it is not already there.

rootfs/etc/templates/haproxy/haproxy.tmpl
```diff
@@ -817,38 +816,44 @@ backend {{ $backend.ID }}
 {{- end }}
 {{- range $pathIDs := $corsCfg.PathIDs $i }}
     http-response set-header Access-Control-Allow-Origin
-        {{- if $dynamicOrigin }}  "%[var(txn.hdr_origin{{ $i }})]"{{ else }}  "{{ index $cors.AllowOrigin 0 }}"{{ end }}
-        {{- if or $dynamicOrigin $pathIDs }} if
+        {{- if or $dynamicOrigin $wildcardWithCreds }}  "%[var(txn.hdr_origin{{ $i }})]"{{ else }}  "{{ index $cors.AllowOrigin 0 }}"{{ end }}
+        {{- "" }} if cors_has_origin{{ $i }}
+        {{- if or $dynamicOrigin $pathIDs }}
```
jcmoraisjr: It seems you don't need this statement anymore, it was there to conditionally add an `if`, which is always added in the line above.

ianroberts: Good catch!

jcmoraisjr:
lgtm, merging now.

There's still a number of tests failing against gwapi conformance v1.5.0, to be handled on a new PR though. This PR otoh is fixing one of them.

Issue #1391, opened 2026-02-27
Use `http-after-response` to add CORS headers


ianroberts:
Re https://github.com/jcmoraisjr/haproxy-ingress/issues/1384#issuecomment-3967174185

Setting response headers with `http-response set-header` only affects "successful" responses, i.e. those generated by a successfully-proxied request to a backend, or by a lua script.  The headers are not applied to responses generated internally by haproxy itself from an `errorfile` or similar.

Switching to `http-after-response set-header` would mean these headers get applied to internal responses as well as proxied/lua ones.  This is particularly useful for CORS as it means that failures like 504 gateway timeout will still have the correct CORS headers and will make it back through to the calling `fetch` rather than getting blocked by the browser's same origin policy.

jcmoraisjr:
Fix just merged to v0.17, v0.16, v0.15 and v0.14, being part of their respective releases in a few weeks from now. Thanks for reporting and all the tests.

PR #1393, opened 2026-02-28
remove CORS headers if Origin is not allowed

jcmoraisjr:
Configures CORS to not include CORS headers in responses whose Origin request header is omitted or not authorized in the allow origin list. If the Origin header is allowed, it is copied to the Access-Control-Allow-Origin response header despite how the allow origin list was configured.

This update not only improves CORS compliance, but also makes the configuration simpler and easier to understand.

### DIFF
--- a/pkg/converters/gateway/gateway.go
+++ b/pkg/converters/gateway/gateway.go
@@ -412,9 +412,10 @@ func (c *converter) syncHTTPRoutesFilterCORS(cors *gatewayv1.HTTPCORSFilter) (ha
 	haCORS.Enabled = true
 	for _, origin := range cors.AllowOrigins {
 		originStr := string(origin)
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
@ianroberts wondering if you want to have a look, if agree with the approach, and have the chance to run some tests on your use cases.

ianroberts:
I don't personally have any use cases where it'll make a difference, since all my apps allow any origin, so I don't think I can do any useful testing for you, but in general I like this approach a lot, much more elegant this way.

While you're working on CORS would it be cheeky to ask if you can switch the directives to be `http-after-response` to fix #1391?

jcmoraisjr:
Thanks, appreciated, always good to have another pair of eyes.

> `http-after-response`

Sure, no need to postpone what we know how to fix and already know that works. I'll do it via another branch though.

PR #1396, opened 2026-03-02
remove default certificate for tlsroute api

jcmoraisjr:
The default certificate configured for HTTP were leaking to non matching TLS routes, on a TLS only listener. It should deny and close the connection instead, before the TLS handshake, since:

1) there is no certificate to be sent to the client, and
2) HTTP is not a protocol to be used for a non matching TLS


ozhankaraman:
Hi Joao, do we plan to add tlsroute gateway api resource definition support on upcoming v0.16?

jcmoraisjr:
Hi, there's some basic support on TLSRoute on v0.16, give it a try, maybe it works on your use case. It is going to be fully compliant only on v0.17.

Issue #1382, opened 2026-02-19
Cors-allow-origin with custom URL scheme (capacitor://) falls back to wildcard * in v0.14.11


VorkerJ:
## Bug Description

When using `cors-allow-origin` annotation with a list (comma-separated or space-separated) containing custom URL schemes like `capacitor://localhost`, HAProxy Ingress falls back to wildcard `*` in the `Access-Control-Allow-Origin` response header instead of returning the specific origin from the request.

This breaks CORS for Capacitor/Ionic mobile applications that require `cors-allow-credentials: true`.

## Environment

- **HAProxy Ingress Version:** `v0.14.11`
- **Image:** `quay.io/jcmoraisjr/haproxy-ingress:v0.14.11`
- **Kubernetes Version:** `1.31`
- **Deployment:** `DaemonSet`

### 1. Create Ingress with origins list

**Tested with both formats:**

```yaml
# Comma-separated (doesn't work)
ingress.kubernetes.io/cors-allow-origin: "capacitor://localhost,https://localhost,http://localhost"

# Space-separated (doesn't work)
ingress.kubernetes.io/cors-allow-origin: "capacitor://localhost https://localhost http://localhost"
```

**Full Ingress manifest:**

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-cors
  namespace: default
  annotations:
    ingress.kubernetes.io/cors-enable: "true"
    ingress.kubernetes.io/cors-allow-origin: "capacitor://localhost,https://localhost,http://localhost"
    ingress.kubernetes.io/cors-allow-methods: "GET,POST,OPTIONS,PUT,PATCH,DELETE"
    ingress.kubernetes.io/cors-allow-headers: "Authorization,Content-Type"
    ingress.kubernetes.io/cors-allow-credentials: "true"
spec:
  ingressClassName: haproxy
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: backend-service
                port:
                  number: 80
```

### 2. Apply the Ingress

```bash
kubectl apply -f ingress.yaml
```

### 3. Test with curl

```bash
curl -v -H 'Origin: capacitor://localhost' \
  -H 'Content-Type: application/json' \
  -d '{"test":"data"}' \
  https://api.example.com/endpoint
```

## Actual Result ❌

```
< HTTP/2 200
< access-control-allow-origin: *
[...]

PR #1401, opened 2026-03-04
make cors allow header rule more flexible

jcmoraisjr:
We have CORS validation for allowed origins, which expects http or https as the URL protocol. Although the majority of origins use http or https, some applications use distinct protocols like capacitor, chrome-extension, maybe other patterns in the future. This update makes the validation flexible, allowing any scheme allowed by the RFC.


jcmoraisjr:
#1382

PR #1402, opened 2026-03-04
add cors headers on proxy generated responses

jcmoraisjr:
CORS headers should be present on all response pages, despite of being generated by the backend or the proxy, otherwise browsers would incorrectly consider Origin as a not allowed one. This update changes the response keyword to http-after-response, which builds the response headers on top of proxy generated responses, either the default ones generated by HAProxy, or the custom ones generated by HAProxy Ingress.


jcmoraisjr:
#1391

Issue #1367, opened 2026-02-05
Redundant option http-server-close in haproxy.tmpl - overridden by option http-keep-alive


ppieczynski:
In rootfs/etc/templates/haproxy/haproxy.tmpl (lines 167-168), the defaults section contains:
   ```
    option http-server-close
    option http-keep-alive
```
According to [HAProxy source code](https://github.com/haproxy/haproxy/blob/v2.6.0/src/cfgparse-listen.c#L1989), these HTTP mode options are mutually exclusive and override each other:

```
/* HTTP options override each other. They can be cancelled using
 * "no option xxx" which only switches to default mode if the mode
 * was this one (useful for cancelling options set in defaults
 * sections).
 */
```

Each option clears PR_O_HTTP_MODE flags before setting its own:

```
curproxy->options &= ~PR_O_HTTP_MODE;
curproxy->options |= PR_O_HTTP_KAL;  // or PR_O_HTTP_SCL
```

Current behavior:

Since option http-keep-alive comes after option http-server-close, only http-keep-alive is active. The http-server-close line has no effect.

Problem:
The generated HAProxy configuration is misleading. Users inspecting their config see both options and may incorrectly assume http-server-close is active (connections to backends are closed after each response). In reality, only http-keep-alive is in effect (persistent connections to backends).

This can lead to:
Incorrect assumptions about load balancing behavior
Confusion when debugging connection issues
False documentation of actual HAProxy behavior

Question:
Is http-keep-alive the intended mode? If so, the option http-server-close line should be removed.
Or was http-server-close intended (common for reverse proxies)? If so, http-keep-alive should be removed.

Expected behavior:
Only one HTTP mode option should be present in the template to avoid confusion and accurately reflect the actual HAProxy behavior.

jcmoraisjr:
Hey thanks for finding and reporting this gem. This is history, let me show you.

> https://github.com/jcmoraisjr/haproxy-ingress/blob/v0.1/rootfs/haproxy.tmpl

This is there since the very beginning, 10 years ago. From where did it come from?

> https://github.com/kubernetes-retired/contrib/blob/master/service-loadbalancer/template.cfg

I contributed to service-loadbalancer, and it was the project that inspired HAProxy Ingress.

---

History aside, lets get rid of the former, since the later is the one active today and looks like the best default option.

PR #1403, opened 2026-03-04
drop misplaced and noop http-server-close option

jcmoraisjr:
`option http-server-close` is incorrectly configured, just before a `option http-keep-alive` that makes it noop. Although this is there since the very beginning, 10y ago, we have no attachment to it, so lets remove.


jcmoraisjr:
#1367

Issue #1328, opened 2025-11-26
load-server-state feature can produce corrupted state file


benedictweis:
**Description of the problem**

Dear haproxy-ingress team, when using the [`load-server-state`](https://haproxy-ingress.github.io/docs/configuration/keys/#load-server-state) feature, HAProxy Ingress controller produces a corrupted state file in `/var/lib/haproxy/state-global`. It is corrupted because HAProxy expects the state file to contain a new line at the end of the file. However, HAProxy Ingress controller truncates the file and removes the new line at the end.

**Expected behavior**

HAProxy is able to reload successfully when using  [`load-server-state`](https://haproxy-ingress.github.io/docs/configuration/keys/#load-server-state).

In the current configuration, the corrupt server state file produces a warning that causes the state file to be ignored.

**Steps to reproduce the problem**

The script below is a minimal example of the behavior described. It can also be reproduces in other modes of operation.

```
# Dependencies: kind, helm, kubectl

# Create a kind cluster
kind create cluster --name haproxy-ingress-test

# Deploy haproxy-ingress
helm repo add haproxy-ingress https://haproxy-ingress.github.io/charts
helm upgrade haproxy-ingress haproxy-ingress/haproxy-ingress\
  --install\
  --create-namespace --namespace ingress-controller\
  --version 0.15.0\
  --wait\
  --timeout 120s\
  -f values.yaml

# Enable the load-server-state feature via ConfigMap
kubectl patch configmap haproxy-ingress --namespace ingress-controller --type merge --patch '{
    "data": {
        "load-server-state": "true"
    }
}'

# Deploy some example ingress resources, this is not necessary to reproduce the issue
rm -rf ./setup.tmp.yaml

for i in $(seq 1 5); do
    cat <<EOF >> setup.tmp.yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
    name: echoserver-$i
    namespace: default
spec:
    replicas: 1
    selector:
        matchLabels:
            app: echoserver-$i
    template:
        metadata:
            labels:
                app: echoserver-$i
        spec:
            containers:
            - name: echoserver
              image: k8s.gcr.io/echoserver:1.3
              ports:
              - containerPort: 8080
---
apiVersion: v1
kind: Service
metadata:
    name: echoserver-$i
    namespace: default
spec:
    type: ClusterIP
    selector:
        app: echoserver-$i
    ports:
    - protocol: TCP
      port: 8080
      targetPort: 8080
[...]

PR #1404, opened 2026-03-05
fix server state file generation

jcmoraisjr:
`load-server-state` config creates a servers state file to be loaded on HAProxy reloads. On worker mode, this file is generated by our socket client, which removes trailing spaces from outputs, leading haproxy to interpret it as corrupted and ignoring it. Manually adding the trailing line break makes it to work again.

The state file generation is also being moved earlier in the reload process, this allows to create an empty file before the first run, so HAProxy does not complain that it cannot find the file.

### DIFF
--- a/docs/content/en/docs/configuration/keys.md
+++ b/docs/content/en/docs/configuration/keys.md
@@ -2103,8 +2103,14 @@ The following annotations are supported:
 Define if HAProxy should save and reload it's current state between server reloads, like
 uptime of backends, qty of requests and so on.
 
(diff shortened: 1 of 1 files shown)


jcmoraisjr:
#1328

jcmoraisjr:
`load-server-state` [is not well supported in HAProxy](https://www.mail-archive.com/haproxy@formilux.org/msg46626.html), so skipping backport on GA branches to avoid deployments that makes it "suddenly start to work".

Issue #1360, opened 2026-01-23
Certificate reading fails for certificates with trailing whitespace


DCkQ6:
## Description

HAProxy Ingress fails to read TLS certificates from Kubernetes secrets when the certificate data contains trailing whitespace after the last valid PEM block.

## Error Message

```
 INFO    converter    ingress/ingress.go:976    warning: using default certificate due to an error reading secret 'all-example-com' on Ingress 'my-ingress/example-com': no valid PEM formatted block found
```

## Steps to Reproduce

1. Create a Kubernetes TLS secret with a certificate that has trailing whitespace after the certificate block
2. Reference the certificate in an Ingress resource
3. Observe the error in controller logs

### Example

A certificate secret where `tls.crt` contains:
```
-----BEGIN CERTIFICATE-----
...certificate data...
-----END CERTIFICATE-----

```
(Note the trailing newline or whitespace)

## Expected Behavior

The controller should successfully parse certificates that have trailing whitespace after the last valid PEM block, as long as at least one valid PEM block was decoded. This is standard behavior for PEM parsers and matches the pattern used elsewhere in the codebase (e.g., `pkg/acme/x/acme/acme.go`).

## Actual Behavior

The controller fails with "no valid PEM formatted block found" error and falls back to the default certificate, even when a valid certificate was present in the data.

## Environment

- HAProxy Ingress version: v0.15.0
- Kubernetes version: v1.33.5

## Proposed Fix

The fix should allow trailing non-PEM data after at least one valid PEM block has been decoded. This is a common requirement for robust PEM parsing.

### Suggested Code Change

In `checkValidCertPEM()`:
```go
if block == nil {
    // If we've already decoded at least one valid certificate, allow trailing non-PEM data
    if x509crt != nil {
        break
    }
    return nil, fmt.Errorf("no valid PEM formatted block found")
}
```

Similar change should be applied to `checkValidPEM()` function:
```go
if block == nil {
    // If we've already decoded at least one valid block, allow trailing non-PEM data
    if der != nil {
        break
    }
    return nil, fmt.Errorf("no valid PEM formatted block found")
}
```

## Workaround

Manually clean the certificate data in the Kubernetes secret to remove any trailing whitespace or non-PEM data, or apply the fix manually.

jcmoraisjr:
Merged to v0.17, v0.16 and v0.15, new v0.17 later this week, v0.16 and v0.15 one week or so later.

PR #1405, opened 2026-03-05
improve validation of PEM data

jcmoraisjr:
Certificates and keys provided by the user are validated to avoid misconfigurations added to haproxy, leading it not to properly reload, and failing in a way hard to find the root cause.

Our validation however is failing to identify the end of a PEM encoded file: trailing spaces and line breaks are decoded, no new PEM block is found, and the whole validation fails.

This update is changing the way we finish the validation: it succeeds in case no more PEM data is found, but we already found what we were looking for.

### DIFF
--- a/pkg/controller/services/ssl.go
+++ b/pkg/controller/services/ssl.go
@@ -31,6 +31,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
#1360

Issue #786, opened 2021-05-25
mTLS support without using sni


ndbemmanuel:
Is it possible to add support to have mutual TLS authentication without having SNI support.  We have deployed devices that need to have mutual TLS authentication and doesn’t support SNI.  We did get a somewhat working solution by having the following config :
```
bind-https: ":443 ssl crt /var/lib/haproxy/crt/sw0081-dev_sw0081-srv.pem.pem ca-file /var/lib/haproxy/cacerts/ca_sw0081-dev_sw0081-ca.pem.pem verify optional"

config-frontend
        http-request set-var(req.hostbackend) var(req.base),map_dir(/etc/haproxy/maps/_front_https_sni__prefix.map) if !{ var(req.hostbackend) -m found }
        acl tls-has-invalid-crt ssl_c_ca_err gt 0
        acl tls-has-invalid-crt ssl_c_err gt 0
        acl tls-need-crt hdr(host) -i -m str -f /etc/haproxy/maps/_front_tls_needcrt__exact.list
        acl tls-check-crt hdr(host)  -i -m str -f /etc/haproxy/maps/_front_tls_auth__exact.list
        http-request set-var(req.tls_invalidcrt_redir) str(_internal) if tls-has-invalid-crt tls-check-crt
```
In sort, it’s a copy the SNI config but applied to the host header field

But that solution has some problems:
- The container will crash before the required ingress rules are created with a file not found
- Modification of the ingress rules like adding a path prefix rule will break the fronted
- It is applied both to HTTP and HTTPS front end
- Depends of the internal working of the load balancer

Our suggestion:
- Have an option to ingress annotation would tell that the endpoint needs to be accessible using only the host header field.
  - Set the HTTPS binding to have verify optional
  - Configure the front end to validate both with and without the SNI
  - Configure the front end to find the correct back end with and without SNI

jcmoraisjr:
Iirc the details of this functionality, the SNI extension has three roles:

- choose a proper server certificate to the client;
- defines if mTLS should be used and use a proper CA bundle;
- if a client certificate is provided, guarantee that the client was properly authenticated before forward the request.

The first two options is a matter of configuring mTLS in the default backend, but HAProxy Ingress currently doesn't do this and I'll take care. The third and last option compares SNI and Host header if a certificate was provided, this behavior can be changed on behalf of a less secure deployment, which isn't a problem at all if all of your clients use the same CA bundle.

Is this approach useful in your use case?

[...]

PR #1408, opened 2026-03-06
configure CA for requests missing SNI

jcmoraisjr:
TLS handshake happens around the SNI extension provided by the client. It is used to choose an appropriate server certificate, defines if the server will accept a client certificate or even enforce that one is provided. None of this can happen if client does not provide SNI.

Older clients however don't support SNI at all, so their requests cannot have the benefit of a dedicated certificate and dedicated mTLS behavior. We already have default certificate for missing or unknown SNI, and this update includes the support for mTLS as well. In case a CA and optional CRL is provided, HAProxy asks an optional client certificate during the TLS handshake. If provided, all the SSL headers are populated just like in the SNI matching counterpart.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -238,7 +238,6 @@ intm
 invalidcrt
 IPHTTP
 iplist
-ipnode
 IPTCP
 IPVS
 ironashram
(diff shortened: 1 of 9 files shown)


jcmoraisjr:
#786

Issue #1361, opened 2026-01-26
v0.15.1: Buggy behaviour when using regex & rewrite target


thorsten-frlncr:
# Description of the problem

Using `path-type: regex` in combination with `rewrite-target` cannot be used if your regex makes use of capture groups.

# Expected behavior

As haproxy applies `(.*)$` to the regex, it must reference it with the correct capture group id.

# Steps to reproduce the problem

With the following ingress manifest (snippets):
```
...
  annotations:
    haproxy-ingress.github.io/path-type: regex
    haproxy-ingress.github.io/rewrite-target: /\1\2
...
          - path: /test/(me/)(.*)$
            pathType: ImplementationSpecific
...
```
I get a backend configuration in haproxy.cfg:
```	
    http-request replace-path ^/test/(me/)(.*)$(.*)$       /\1\2\1
```

changing the ingress to 
```
...
  annotations:
    haproxy-ingress.github.io/path-type: regex
    haproxy-ingress.github.io/rewrite-target: /\1
...
          - path: /test/(me/)
            pathType: ImplementationSpecific
...
```
I consequently end up with
```
    http-request replace-path ^/test/(me/)(.*)$       /\1\1
```

# Environment information

HAProxy Ingress version: `v0.15.1`

# Note

This seems to be a bug? I am wondering if my use cases is that specific that it didn't come up before or if I am doing something wrong here?

heinrichJan:
Dear @thorsten-frlncr - thanks for opening up this issue! Im currently at the very same point with our work on ingress, so we ran into the same problem here. It would be great if we find a solution here :)

fleroy-isagri:
Hi,

I had the same problem

As a workaround, i use a configuration snippet : https://haproxy-ingress.github.io/docs/configuration/keys/#configuration-snippet

With the annotation `haproxy-ingress.github.io/config-backend`, it fixed the configuration to use.

For your example, it will be : 

```
    haproxy-ingress.github.io/config-backend: |
      http-request replace-path ^/test/(me/)(.*)$       /\1\2
```

heinrichJan:
Hi @fleroy-isagri  - unfortunately this doesnt work for me, any update in this topic ?

jcmoraisjr:
[...]

PR #1411, opened 2026-03-09
fix rewrite path handling for regex paths

jcmoraisjr:
Rewrite path were never revisited after implementing distinct path types. It is handling path match using `begin` match, although `prefix` differs slightly, `exact` has a big difference, and `regex` is a completely different approach.

This is the first of two distinct improvements. In this first step the classic behavior is maintained without any change for backward compatibility for `begin`, `prefix` and `exact`. However if path type is `regex`, we are completely skipping the rewrite handling, giving the user the ability to configure it the way they need.

Should be merged as far as v0.14.


jcmoraisjr:
#1361

PR #1419, opened 2026-03-09
fix(gateway): prefer v1 resources when both v1 and v1beta1 are served

### LINKED ISSUE #1387: GatewayClass and Gateway status conditions are never updated despite successful reconciliation

Note reporting the correct update/status/condition on reconciliation. But the reconciliation works.

```
k describe gatewayclasses.gateway.networking.k8s.io haproxy
Name:         haproxy
Namespace:
Labels:       <none>
Annotations:  argocd.argoproj.io/tracking-id: argoapps:gateway.networking.k8s.io/GatewayClass:argocd/haproxy
API Version:  gateway.networking.k8s.io/v1
Kind:         GatewayClass
Metadata:
  Creation Timestamp:  2026-02-24T17:35:52Z
  Generation:          1
  Resource Version:    594558348
  UID:                 3f134ebb-c20c-4bdd-ba1b-a193e399541e
Spec:
  Controller Name:  haproxy-ingress.github.io/controller
Status:
  Conditions:
    Last Transition Time:  1970-01-01T00:00:00Z
    Message:               Waiting for controller
    Reason:                Pending
    Status:                Unknown
    Type:                  Accepted
Events:                    <none>
```

```
k describe gateway test
Name:         test
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  gateway.networking.k8s.io/v1
Kind:         Gateway
Metadata:
  Creation Timestamp:  2026-02-24T23:30:30Z
  Generation:          1
  Resource Version:    594774623
  UID:                 a4eed684-5f1a-4f72-982a-63c6bc81c719
Spec:
  Gateway Class Name:  haproxy
  Listeners:
    Allowed Routes:
      Namespaces:
        From:  Same
    Name:      http
    Port:      80
    Protocol:  HTTP
Status:
  Conditions:
    Last Transition Time:  1970-01-01T00:00:00Z
    Message:               Waiting for controller
    Reason:                Pending
    Status:                Unknown
    Type:                  Accepted
    Last Transition Time:  1970-01-01T00:00:00Z
    Message:               Waiting for controller
    Reason:                Pending
    Status:                Unknown
    Type:                  Programmed
Events:                    <none>
```

```
k logs -n haproxy-ingress deploy/haproxy-ingress -f | grep gateway
Defaulted container "haproxy-ingress" out of: haproxy-ingress, haproxy, haproxy-ingress-init (init)
I0224 23:25:06.622874       7 config.go:213] "watching for Gateway API resources - --watch-gateway is true" logger="config"
I0224 23:25:06.629768       7 config.go:223] "found custom resource definition for gateway API v1" logger="config"
I0224 23:25:06.631896       7 config.go:227] "found custom resource definition for gateway API v1beta1" logger="config"
I0224 23:25:07.117341       7 services.go:274] "finish haproxy update" logger="services" id=1 parse-gateway-ms=0.09824 parse-ingress-ms=0.109721 write-maps-ms=0.167361 write-config-ms=1.121889 reload-haproxy-ms=77.865295 total-ms=79.36251
I0224 23:25:20.785573       7 services.go:274] "finish haproxy update" logger="services" id=2 parse-gateway-ms=1.621172 parse-ingress-ms=0.088441 write-maps-ms=0.00484 total-ms=1.714453
I0224 23:25:29.410954       7 services.go:274] "finish haproxy update" logger="services" id=3 parse-gateway-ms=0.05052 parse-ingress-ms=0.087961 write-maps-ms=0.197041 write-config-ms=6.390528 reload-haproxy-ms=124.764244 address-status-update-ms=101.31263 total-ms=232.80293
I0224 23:25:47.128441       7 services.go:274] "finish haproxy update" logger="services" id=4 parse-gateway-ms=0.04704 parse-ingress-ms=0.025201 write-maps-ms=0.0046 address-status-update-ms=0.220601 total-ms=0.297442
I0224 23:30:30.971109       7 services.go:274] "finish haproxy update" logger="services" id=5 parse-gateway-ms=0.064281 parse-ingress-ms=0.089601 write-maps-ms=1.40605 write-config-ms=3.757428 reload-haproxy-ms=83.01806 address-status-update-ms=0.275642 total-ms=88.61106
I0224 23:32:12.061387       7 services.go:274] "finish haproxy update" logger="services" id=6 parse-gateway-ms=0.04644 parse-ingress-ms=0.0198 write-maps-ms=0.0038 address-status-update-ms=0.163882 total-ms=0.233922
I0224 23:32:16.916190       7 services.go:274] "finish haproxy update" logger="services" id=7 parse-gateway-ms=0.188761 parse-ingress-ms=0.0382 write-maps-ms=0.004281 address-status-update-ms=0.223361 total-ms=0.454603
I0224 23:32:40.381762       7 services.go:274] "finish haproxy update" logger="services" id=8 parse-gateway-ms=0.206601 parse-ingress-ms=0.069401 write-maps-ms=0.585764 write-config-ms=9.014747 reload-haproxy-ms=102.83573 address-status-update-ms=1.616212 total-ms=114.32845
I0224 23:32:49.167636       7 services.go:274] "finish haproxy update" logger="services" id=9 parse-gateway-ms=0.061081 parse-ingress-ms=0.02136 write-maps-ms=0.00396 address-status-update-ms=3.020302 total-ms=3.106703
```

**Expected behavior**

GatewayClass.status.conditions[type=Accepted]=True
Gateway.status.conditions[type=Programmed]=True

1. Install Gateway API CRDs (standard).
2. Install haproxy-ingress with watch-gateway.
3. Create GatewayClass + Gateway.
4. Create HTTPRoute.
5. Send request.

**Environment information**

HAProxy Ingress version: `v0.16.0-beta.2`
Command-line options:

```
--watch-gateway=true
```

Global options:

```
healthz-port: "10253"
stats-port: "1936"
```

Gateway objects:

```
---
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: haproxy
spec:
  controllerName: haproxy-ingress.github.io/controller
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: test
  namespace: default
spec:
  gatewayClassName: haproxy
  listeners:
    - name: http
      port: 80
      protocol: HTTP
```

SergioChan:
## Summary
- prefer Gateway API `v1` objects whenever both `v1` and `v1beta1` are served by the apiserver
- keep `v1beta1` as a fallback path only when `v1` is not available
- apply the same preference in status update conversion logic so status writes target the same API version used for reads/lists

## Testing
- `go test ./pkg/controller/services ./pkg/converters/gateway`

## Related
Fixes #1387

### DIFF
--- a/pkg/controller/services/cache_gateway_prefer_v1_test.go
+++ b/pkg/controller/services/cache_gateway_prefer_v1_test.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/stretchr/testify/require"
-	apiext "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -24,13 +24,13 @@ func TestGatewayAccessorsPreferV1WhenBothVersionsAreEnabled(t *testing.T) {
 
 		cache := newGatewayCacheForTest(t,
 			&gatewayv1.GatewayClass{
-				ObjectMeta: apiext.ObjectMeta{Name: "v1-class"},
+				ObjectMeta: metav1.ObjectMeta{Name: "v1-class"},
 				Spec: gatewayv1.GatewayClassSpec{
 					ControllerName: gatewayv1.GatewayController("example.com/controller"),
 				},
 			},
 			&gatewayv1beta1.GatewayClass{
-				ObjectMeta: apiext.ObjectMeta{Name: "beta-class"},
+				ObjectMeta: metav1.ObjectMeta{Name: "beta-class"},
 				Spec: gatewayv1beta1.GatewayClassSpec{
 					ControllerName: gatewayv1beta1.GatewayController("example.com/controller"),
 				},
@@ -48,10 +48,10 @@ func TestGatewayAccessorsPreferV1WhenBothVersionsAreEnabled(t *testing.T) {
 
 		cache := newGatewayCacheForTest(t,
 			&gatewayv1.HTTPRoute{
-				ObjectMeta: apiext.ObjectMeta{Name: "v1-route", Namespace: "default"},
+				ObjectMeta: metav1.ObjectMeta{Name: "v1-route", Namespace: "default"},
 			},
 			&gatewayv1beta1.HTTPRoute{
-				ObjectMeta: apiext.ObjectMeta{Name: "beta-route", Namespace: "default"},
+				ObjectMeta: metav1.ObjectMeta{Name: "beta-route", Namespace: "default"},
 			},
 		)
 


jcmoraisjr:
Hi thanks for the contribution!

Do you have any evidence that beta versions are being chosen when both beta and v1 are deployed? As far as I can tell just one `Has<API><Version>` is flagged true per API in runtime, so the same API, the newer one available, should be chosen despite of the order of the checks.

This one-per-api check should be happening here:

https://github.com/jcmoraisjr/haproxy-ingress/blob/6241fc86ac9321ecf7bc48c14f683d6d5cfab60d/pkg/controller/config/config.go#L252-L263

SergioChan:
Good question — I haven’t yet captured a clean repro case that demonstrates both versions being selected in practice under the same API surface.

I’ll reproduce against a cluster serving both Gateway API versions and post concrete discovery logs/steps. If I can’t demonstrate the behavior from this PR, I’ll close this PR rather than keep it speculative.

SergioChan:
Good question — yes, this can happen when both API versions are served.

In the current selection path, resources are grouped by kind and then sorted lexicographically by `group/version/resource`, which can place `v1beta1` ahead of `v1` for the same kind depending on discovered names. The PR changes selection to explicitly prefer stable `v1` when both are present, while keeping existing behavior for kinds that only expose beta.

I’ll add a focused unit test in a follow-up commit to make this preference explicit in coverage.

SergioChan:
Follow-up as promised: I pushed eb212ce adding focused coverage for the dual-version case in pkg/controller/services/cache_gateway_prefer_v1_test.go.

It verifies that when both HasGatewayV1 and HasGatewayB1 are enabled:
- GetGatewayClassMap prefers v1 objects
- GetHTTPRouteList prefers v1 objects

This makes the stable-v1 preference explicit in tests rather than only in implementation order.

SergioChan:
Pushed a follow-up commit (6bb1f3f) to fix the failing `check-spelling` gate.

### What changed
- Renamed the `apiext` import alias to `metav1` in `cache_gateway_prefer_v1_test.go`
- Updated corresponding `ObjectMeta` references

No functional behavior changed; this is only to satisfy spelling checks on the newly added test file.

jcmoraisjr:
Thanks for the follow up on this. I really appreciate the effort.

What I observed however is that the cache mock does not reflect the reality on how `Has<Kind><Version>` flags work. They are mutually exclusive as already stated on my previous message, otherwise the watcher would issue events for all of them, in case both APIs are found in the same apiserver, duplicating the noise in the event queue:

https://github.com/jcmoraisjr/haproxy-ingress/blob/09856616818357277f36f241e85802f3bb2d3d8b/pkg/controller/reconciler/watchers.go#L70-L75

This is the main, maybe the only reason IIRC they work like this, which makes the change noop regarding the functionality. This was the reason I asked you if you have observed this behavior yourself, or you inferred based on how the cache is implemented. If the former, I am really interested on a reproducer, which can be simple steps like deploying CRDs x, y, z, and running some custom integration test or some manual inspections. In case you can produce it, I can help spotting the root cause and we can make the fix the right way, for all the duplicated APIs, not only Gateway, GatewayClass and HTTPRoute.

SergioChan:
Thanks for the detailed explanation — I re-checked this and agree with your point.

My previous validation relied on cache-level mocking and does **not** demonstrate a real runtime path where both `Has<Kind><Version>` flags are simultaneously true. With the watcher behavior you linked, this change is effectively a no-op without a concrete reproducer.

I don’t currently have a reliable real-cluster repro for the duplicated-version selection problem, so I’m closing this PR to avoid speculative churn.

If I can produce a concrete reproducer later (or if you share one), I’ll open a focused fix that matches the real watcher/runtime flow.

Issue #1420, opened 2026-03-12
ingress' status loadbalancer ip is not updated on migration of ingressClass to haproxy


stearz:
**Description of the problem**

I want to migrate from ingress-nginx to haproxy-ingress. haproxy-ingress and ingress-nginx are both installed and are able to serve ingresses with their respective ingressClass.

When I change the ingressClass of an existing ingress from `nginx` to `haproxy`, haproxy-ingress is serving the ingress over its LoadBalancer but it does not update the `status.loadBalancer.ingress[0].ip` to its own LoadBalancer IP address leading to external-dns not updating the DNS records. This only happens when the ip field has already been set. If I re-create the ingress or empty the ip field haproxy-ingress inserts its IP without a problem.

**Expected behavior**

haproxy-ingress should keep the ingress status ip up to date for all ingresses with its ingressClass.

**Steps to reproduce the problem**

1. Create an ingress with a different ingressClass and make sure the loadbalancer ip is being set:
   
    ```yaml
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      name: my-ingress
      namespace: my-namespace
      annotations:
        cert-manager.io/cluster-issuer: letsencrypt
        kubernetes.io/tls-acme: "true"
    status:
      loadBalancer:
        ingress:
        - ip: 123.45.67.89
    spec:
      ingressClassName: nginx
      tls:
      - hosts:
        - my-app.example.com
        secretName: my-secret
      rules:
      - host: my-app.example.com
        http:
          paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: my-service
                port:
                  number: 8080
    ```
2. Update the ingressClass of this ingress to `haproxy`
3. The IP address should not have been changed by haproxy-ingress

**Environment information**

HAProxy Ingress version: `v0.15.1`
`"version info" logger="config" controller-publicname="HAProxy Ingress" controller-release="v0.15.1" controller-build="git-68590ac" controller-repo="https://github.com/jcmoraisjr/haproxy-ingress"`

haproxy-ingress has been installed via helm with this values.yaml:
```
controller:
  ingressClass: haproxy
  ingressClassResource:
    enabled: true
    default: false
  service:
    httpPorts:
      - port: 80
        targetPort: http
    httpsPorts:
      - port: 443
        targetPort: https
  publishService:
    enabled: true
  kind: Deployment
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
[...]

PR #1421, opened 2026-03-13
check ingress status on add and update events

jcmoraisjr:
Ingress status is checked for updates when the ingress resource is created, or when the source of the IP addresses changes its IP. There is however an ignored scenario, when an ingress resource receives a matching ingress class. This is considered like an add operation in the partial parsing, but the generated event is an update, making this existing ingress resource to be ignored, potentially leading to an outdated status. This update iterates over all reported changes, checking ingress status and updating if needed.


jcmoraisjr:
#1420

Issue #1430, opened 2026-03-18
Use numeric USER in Dockerfile


ianroberts:
The `Dockerfile` needs to switch to `root` to install some extra packages, and then switch back to the unprivileged `haproxy` user for runtime.  Currently it does this with

```
USER haproxy
```

However this does not interact nicely when you try to run the ingress controller pod with a `securityContext` that specifies `runAsNonRoot` (e.g. the "Restricted" pod security standard) - if a pod specifies `runAsNonRoot: true` but does _not_ explicitly specify a `runAsUser` UID, then the pod will fail to schedule if the image configuration uses a symbolic username for the `USER` directive, since the container runtime cannot be _certain_ that this symbolic name will not resolve to UID 0 once the container is up and running.  Specifying the user numerically

```
USER 99
```

provides this guarantee; the runtime can tell from the config alone that the image will not run as root.

PR #1431, opened 2026-03-18
Use numeric USER in Dockerfile

ianroberts:
When switching back to the unprivileged user in the image build, we should specify this via the numeric UID (99) rather than the symbolic username (haproxy).  This is good practice generally, but it is essential for compatibility with the `runAsNonRoot` container security context setting in Kubernetes.

When a container is configured to `runAsNonRoot` but does _not_ have an explicit `runAsUser` setting, Kubernetes will refuse to run it if the image specifies a symbolic `USER` name in its config, since it is impossible for the container runtime to know for certain that this name will not map to UID 0 once the container is up and running.  Specifying the `USER` numerically provides this guarantee.

Fixes jcmoraisjr/haproxy-ingress#1430

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -43,7 +43,8 @@ RUN mkdir -p /var/empty /etc/haproxy /var/lib/haproxy /var/run/haproxy\
  && setcap 'cap_net_bind_service=+ep' /usr/local/sbin/haproxy
 
 STOPSIGNAL SIGTERM
-USER haproxy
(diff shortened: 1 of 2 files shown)


ianroberts:
Sorry, force-push was just because I forgot that there are two copies of the Dockerfile in the repo and needed to update the second one to stay in sync.

jcmoraisjr:
Looks good, thanks for spotting and addressing this.

PR #1434, opened 2026-03-20
Always run `go build` on the BUILDPLATFORM architecture

ianroberts:
Refactor the docker build to use Go's ability to cross-compile.  This means that all of the `go build` commands are running on the native architecture of the build machine, even though the output binaries target different platforms.  This should substantially speed up multi-arch builds as the `apk upgrade` step is the only significant item that needs to run under emulation.

This makes use of the handy [xx](https://github.com/tonistiigi/xx) tool to set all the right cross-compilation flags.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM --platform=$BUILDPLATFORM tonistiigi/xx:latest AS xx
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.9.0 AS xx
 
 FROM --platform=$BUILDPLATFORM golang:1.26.1-alpine AS builder-common
 COPY --from=xx / /


builder/Dockerfile
```diff
@@ -12,21 +12,32 @@
...
+
+FROM --platform=$BUILDPLATFORM golang:1.26.1-alpine AS builder-common
+COPY --from=xx / /
 
 ARG GIT_TAG
 WORKDIR /src
 RUN apk --no-cache add git
 
+COPY go.mod go.sum ./
+RUN go mod download
```
ianroberts: Another minor optimisation - only download the go modules once, before the build splits off by `TARGETPLATFORM`, rather than repeatedly in each platform-specific builder.

builder/Dockerfile
```diff
@@ -12,21 +12,32 @@
...
 COPY . .
 
 RUN GIT_TAG=${GIT_TAG:-latest}\
  && GIT_COMMIT=git-$(git rev-parse --short HEAD)\
  && GIT_REPO=$(git config --get remote.origin.url)\
  && VERSION_PKG=github.com/jcmoraisjr/haproxy-ingress/pkg/version\
- && CGO_ENABLED=0 go build\
-    -ldflags "-s -w -X ${VERSION_PKG}.RELEASE=${GIT_TAG} -X ${VERSION_PKG}.COMMIT=${GIT_COMMIT} -X ${VERSION_PKG}.REPO=${GIT_REPO}"\
-    -o haproxy-ingress pkg/main.go
+ && echo "-X ${VERSION_PKG}.RELEASE=${GIT_TAG} -X ${VERSION_PKG}.COMMIT=${GIT_COMMIT} -X ${VERSION_PKG}.REPO=${GIT_REPO}" > /x-args
```
ianroberts: Because I've split the git parsing and the actual `go build` into separate `RUN` instructions I can't just pass shell variables between the two, so I'm writing the dynamic part of the `ldflags` to a temporary file and then reading them back from that file in the actual build command.

ianroberts:
I think I've addressed your concerns now.  I considered rolling #1431 into this branch but I think they're far enough apart in the Dockerfile that they'll cleanly merge as-is.

jcmoraisjr:
Thank you! LGTM, merging it now.

> I considered rolling https://github.com/jcmoraisjr/haproxy-ingress/pull/1431 into this branch but I think they're far enough apart in the Dockerfile that they'll cleanly merge as-is.

Yep, better to have distinct PRs addressing distinct issues. Thank you again for all the help, really appreciated.

ianroberts:
Should have cherry-picked this before you tagged v0.16 😂

Issue #1437, opened 2026-03-23
cert-manager certificates not read correctly in v0.15.2


StopMotionCuber:
**Description of the problem**
We did have auto-updates for haproxy ingress patch releases enabled, and today the update to 0.15.2 was performed.
This lead to an issue that certificates issued by cert-manager weren't recognized anymore (I'm using HARICA certificates, and I'm not sure whether the issue is related to their certificate format, and does not exist on LetsEncrypt).

Afterwards, traffic was using a self-signed certificate, leading (expectedly) to a lot of SSL handshake errors.

I saw the following logline:
```
I0323 13:43:24.829181       2 ingress.go:976] "warning: using default certificate due to an error reading secret '<secret_name>' on Ingress '<namespace>/<ingress>': secret '<namespace>/<secret_name>' does not have keys 'tls.crt' and 'tls.key'" logger="converter"
```
I did verify that this claim is false and it does have both of these values. I also tried to dive deeper, and noticed that there's the following line, which I'd expect if the last logline was actually true (but it didn't appear in logs):
```go
	return nil, fmt.Errorf("secret '%s/%s' have neither ca.crt nor tls.crt/tls.key pair", ns, name)
```

There was also a reconciliaton error, but that is due to the fact it tried to reference the secret, which couldn't be gathered. It points to `ingress.go:466`
```
E0323 13:43:24.829288       2 signal_unix.go:925] "Observed a panic" logger="reconciler" reconcileID="668d0498-b86a-4c25-a4e3-0fd1c6d0fb70" panic="runtime error: invalid memory address or nil pointer dereference" panicGoValue="\"invalid memory address or nil pointer dereference\"" stacktrace=...
```

I've upgraded to `0.16.0` now, the issue doesn't exist there.

**Expected behavior**
Well, the certificate provided via secret should continue to work TLS.

**Steps to reproduce the problem**

1. Create an ingress with certificate issued via cert-manager, issuer in this case is HARICA
2. Install HAProxy v0.15.1 with a working TLS configuration
3. Update to HAProxy v0.15.2

I'm not quite happy with the steps to reproduce here, but this is dealing with secrets I would rather like to not expose.

**Environment information**

HAProxy Ingress version: `v0.15.2`

jcmoraisjr:
Hi, thanks for reporting this issue. I can clearly see the panic you reported, we're missing either a validation or a properly configured defaultCrt and we should fix this.

[...]

PR #1441, opened 2026-03-25
enable leaf reference on go1.19 compatibility mode

jcmoraisjr:
A fix on x509 certificate validation for v0.15.1 changed the func used to parse the crt and key pair. The method creates a `tls.Certificate`, whose `Leaf` attribute is used internally.

The `Leaf` attribute is populated since go1.23, which is older than the version used to compile the controller. The legacy controller engine code used up to v0.14 have compatibility issues with go version newer than 1.19. v0.15 added controller-runtime, but still has the older engine available that can be enabled in legacy mode. Because of the legacy engine, v0.15 code runs in go1.19 compatibility mode which leads the Leaf attribute of the certificate to become unassigned.

The integration tests cover this functionality and would caught this error, but the compatibility mode was being applied only in the main package. This update not only enables the Leaf attribute, but also adds the same compatibility mode in the integration tests package as well.

This is a v0.15 failure only, v0.16 had the compatibility mode removed along with the legacy controller engine, and v0.14 uses only the legacy code to validate certificates.

### DIFF
--- a/pkg/main.go
+++ b/pkg/main.go
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 //go:debug default=go1.19
+//go:debug x509keypairleaf=1
 package main
 
 // TODO remove go1.19 compatibility after dropping legacy mode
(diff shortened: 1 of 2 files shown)


jcmoraisjr:
#1437

PR #1442, opened 2026-03-26
add build image step in build workflow

jcmoraisjr:
Since the refactor of the Dockerfile builder, which cross-compile in the build platform instead of compile in the target, emulated one, the build step should be (almost) fast enough to be part of the regular build.

Adding the build image step can anticipate issues during cross-compile and local image generation for the target platform.


jcmoraisjr:
cc @ianroberts

Issue #1452, opened 2026-04-20
Panic: counter cannot decrease in value


wiegandf:
**Description of the problem**

Haproxy-Ingress-Controller crashed with a panic while updating a prometheus counter.

```bash
2026-04-16 01:13:10.409 	github.com/jcmoraisjr/haproxy-ingress/pkg/controller/controller.go:189 +0x1f1 

2026-04-16 01:13:10.409 created by github.com/jcmoraisjr/haproxy-ingress/pkg/controller.(*HAProxyController).startServices in goroutine 1 

2026-04-16 01:13:10.409 	k8s.io/apimachinery@v0.24.17/pkg/util/wait/wait.go:92 

2026-04-16 01:13:10.409 k8s.io/apimachinery/pkg/util/wait.Until(...) 

2026-04-16 01:13:10.409 	k8s.io/apimachinery@v0.24.17/pkg/util/wait/wait.go:135 +0x7f 

2026-04-16 01:13:10.409 k8s.io/apimachinery/pkg/util/wait.JitterUntil(0x3d5c4c533a30, 0x1dcd6500, 0x0, 0x1, 0x3d5c2cf16700) 

2026-04-16 01:13:10.409 	k8s.io/apimachinery@v0.24.17/pkg/util/wait/wait.go:158 +0xaf 

2026-04-16 01:13:10.409 k8s.io/apimachinery/pkg/util/wait.BackoffUntil(0x3d5c4c533a30, {0x555c9f12eac0, 0x3d5c4b75f2c0}, 0x1, 0x3d5c2cf16700) 

2026-04-16 01:13:10.409 	k8s.io/apimachinery@v0.24.17/pkg/util/wait/wait.go:157 +0x33 

2026-04-16 01:13:10.409 k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1(0x3d5c2dc09f40?) 

2026-04-16 01:13:10.409 	github.com/jcmoraisjr/haproxy-ingress/pkg/controller/controller.go:190 +0x1f 

2026-04-16 01:13:10.409 github.com/jcmoraisjr/haproxy-ingress/pkg/controller.(*HAProxyController).startServices.func1() 

2026-04-16 01:13:10.409 	github.com/jcmoraisjr/haproxy-ingress/pkg/haproxy/instance.go:268 +0x2ae 

[...]

PR #1456, opened 2026-04-21
adding boundary in the idle_pct metric

jcmoraisjr:
HAProxy's Idle_pct metric should go between 0 and 100, which generates a valid counter for controller's idle metric. However the counter calculation has already reported a negative value which led to Prometheus panicking. The only way to the value being negative is a Idle_pct as 101 or greater. This update adds boundary check as a safeguard, in order to protect the controller process against inaccurate data coming from HAProxy.

Also, we're skipping metric reading in case of an unexpected failure reading the Idle_pct metric.

Should be merged to all the supported branches.


jcmoraisjr:
#1452

Issue #1448, opened 2026-04-10
Multiple auth backends with stale addresses when service serving auth-url has HPA


fgbreel:
**Description of the problem**

Hey there,

I'm in the process of migrating away from `ingress-nginx` to `haproxy-ingress` and it has been working pretty well for me. Thank you!

I noticed a strange behavior that leads to HTTP 5XX after some time running `haproxy-ingress` + using `auth-url` pointing to a service whose horizontal pod autoscaler is enabled.

Over time `/etc/haproxy/haproxy.cfg` configuration file accumulates several entries for `_auth_backend0XX_YYYY` containing IPs from pods that no longer exists.

Then, when requesting a service that has an Ingress annotated with `auth-url` this error messages shows up in the logs:

```
[warning] 099/172825 (103) : Failure in auth-request backend '_auth_14427': http.get: Receive error (headers): connection closed.
[warning] 099/172819 (235) : Failure in auth-request backend '_auth_14446': http.get: Receive error (headers): connection closed.
```

Here is a fragment of `haproxy.cfg` when the issue occurs: [parts-of-haproxy-cfg.txt](https://github.com/user-attachments/files/26640550/parts-of-haproxy-cfg.txt)

**Expected behavior**

No HTTP 5XX when service serving `auth-url` is autoscaled up/down. 

**Steps to reproduce the problem**

1. Enable `auth-url` pointing to a service with horizontal pod autoscaler enabled.
2. Trigger autoscaling for auth service.
3. Check `/etc/haproxy/haproxy.cfg` for duplicated `_auth_backend0XX_8080`.

**Workaround**

Set `service-upstream: true`.

**Environment information**

EKS 1.33+
HAProxy Ingress version: `v0.15.3` (via helm chart)
I'm using the same configuration mentioned [here](https://github.com/jcmoraisjr/haproxy-ingress/issues/1445#issuecomment-4217192366), but without the `service-upstream: true`.

```yaml
---
controller:
  kind: DaemonSet
  ingressClassResource:
    # watch for Ingress objects
    enabled: true
  publishService:
    # when 'true' the .status field of Ingress receives the LB address.
    enabled: false
  service:
    type: NodePort
    extraPorts:
    - port: 10253
      targetPort: 10253
  podAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "9101"
  metrics:
    enabled: true
    embedded: false
  stats:
    enabled: true
  resources:
    requests:
[...]

PR #1462, opened 2026-04-25
add tracking for http based auth backends

jcmoraisjr:
External authentication needs a backend to reach and load balance the external service. It uses a fully managed one on svc proto, and a manually created one on http and https protos. The later was missing a proper tracking, so changes in the ingress or the underlying service of that backend will properly remove the unused ones.

The manually created authentication backends need to use deterministic names, so new reconciliations also generates the same name and avoids misbehavior during partial parsing and dynamic configuration check.

Also, the tracked backend removal was missing to remove the reference to the manually created backend, which is fixed now.


jcmoraisjr:
#1448

Issue #1451, opened 2026-04-19
v0.17, IPv6 and Gateway issue


ctr49:
**Description of the problem**

I'm trying to setup haproxy as gateway (only, no ingress). The cluster is dual-stack with IPv6 first. However, the frontends only bind to IPv4.
Despite
```
bind-http: :80,:::80
bind-https: :443,:::443
```
in the configmap - probably due to the shift from Global to Frotend scopes, I can't achieve an IPv6 listener. However, I also can't configure this in Frontend context as gateway objects don't support annotations-based configuration (as far as I understand).

**Expected behavior**

Global listeners on IPv6 address.

**Steps to reproduce the problem**

1. Deploy dual-stack Kubernetes cluster
2. Deploy haproxy-ingress (i.e. using Helm chart)
3. Configure IPv6 listeners:
``` bash
$ kubectl -n haproxy get configmap prod-haproxy-haproxy-ingress -o yaml |   grep -nE 'bind-http|bind-https|bind-ip-addr-http|bind-ip-addr-healthz'
3:  bind-http: :80,:::80
4:  bind-https: :443,:::443
5:  bind-ip-addr-healthz: '::'
```
4. Oberserve listeners on IPv4 only:
``` bash
$ kubectl -n haproxy exec -it prod-haproxy-haproxy-ingress-778b6f4c79-9w4bf --   sh -c "grep -nE '^[[:space:]]*bind ' /etc/haproxy/haproxy.cfg"
84:    bind :443 ssl alpn h2,http/1.1 crt-list /etc/haproxy/maps/_front_https_bind_crt.list ca-ignore-err all crt-ignore-err all
112:    bind :80
146:    bind :1936
159:    bind :::10253
```

**Environment information**

HAProxy Ingress version: `v0.17-alpha.1`

ConfigMap:
``` yaml
apiVersion: v1
data:
  bind-http: :80,:::80
  bind-https: :443,:::443
  bind-ip-addr-healthz: '::'
  config-frontend-late: |
    http-request redirect scheme https code 308 if !{ ssl_fc } !{ path_beg /.well-known/ }
    http-request set-var(req.hostbackend) var(req.base),map_dir(/etc/haproxy/maps/_front_http_host__prefix.map) if { ssl_fc }
  create-default-frontends: "true"
  healthz-port: "10253"
  stats-port: "1936"
kind: ConfigMap
metadata:
  annotations:
[...]

PR #1466, opened 2026-04-29
add ip mode command-line option

jcmoraisjr:
Add option to customize the IP stack HAProxy should use when configuring loopback addresses (used in auth external and empty slots for dynamic update), as well as the frontend listeners.

By default the IP stack is read from the node HAProxy Ingress is deployed - the IP formats used on node's addresses makes HAProxy Ingress bind ports on those same stacks. For the loopback addresses, IPv4 is always used except if the v4 stack is not found.

This adds a backward compatibility change: if controller can reach the node status, IPv6 should be enabled if the node has an IPv6 address. Use --ip-mode=v4 to preserve the legacy behavior. IPv4 only is used by default if there is an error reading node status.


jcmoraisjr:
#1451

PR #1480, opened 2026-05-18
Fix not rendered markdown

killermoehre:
https://haproxy-ingress.github.io/ doesn't render the markdown proper, so switching to plain HTML it is.


jcmoraisjr:
Thanks! lgtm, merging now.

PR #1481, opened 2026-05-20
Make server rename part of dynamic updates

alexanderstephan:


alexanderstephan:
Keeping this as a draft for now as we don't have the server rename feature yet.

pkg/haproxy/dynupdate.go
```diff
@@ -287,9 +300,17 @@ func (d *dynUpdater) dynamicallySyncSlots(pair *backendPair) bool {
...
 			pair.cur = added[0]
 			pair.cur.Name = pair.old.Name
 			added = added[1:]
+			if wantRename && desiredName != pair.old.Name {
+				renames = append(renames, pendingRename{
+					ep:          pair.cur,
+					oldName:     pair.old.Name,
+					desiredName: desiredName,
+				})
+			}
```
jcmoraisjr: Do you predict some use on this part of the code? From the best of my knowledge you cannot have epPair joining EPs taking advantage of a rename because this is done based on ep.target, which should change both on IP:port changes and on targetRef changes.

pkg/haproxy/dynupdate.go
```diff
@@ -314,7 +343,35 @@ func (d *dynUpdater) dynamicallySyncSlots(pair *backendPair) bool {
...
+			d.execEnableServer(curBack.ID, r.ep)
+			updated = false
+			continue
+		}
+		r.ep.Name = r.desiredName
+		if !d.execEnableServer(curBack.ID, r.ep) {
+			updated = false
+			continue
+		}
+		d.logger.InfoV(2, "renamed server on backend '%s' from '%s' to '%s'", curBack.ID, r.oldName, r.desiredName)
```
jcmoraisjr: It seems that this orchestration fits better on the `exec<...>Endpoint` family of methods.

pkg/haproxy/dynupdate.go
```diff
@@ -268,6 +268,19 @@ func (d *dynUpdater) dynamicallySyncSlots(pair *backendPair) bool {
...
+	// rename afterwards so that the server state file stays consistent.
+	// Only meaningful with EpIPPort or EpTargetRef naming where the name
+	// carries semantic value (IP:port or pod name).
+	type pendingRename struct {
+		ep          *hatypes.Endpoint
+		oldName     string
+		desiredName string
+	}
+	var renames []pendingRename
+	wantRename := curBack.EpNaming == hatypes.EpIPPort || curBack.EpNaming == hatypes.EpTargetRef
```
jcmoraisjr: wantRename should also consider the HAProxy version, which we don't have yet but it's simple to add in the model. Meanwhile it worth a TODO here so we don't forget to handle this when we're going to revisit the code after moving out from draft.

jcmoraisjr (commented):
Looks good! Nice approach and good tests. Just added some minor comments and doubt.

Wondering if you can rebase it so we can fully run the tests and make it green.

alexanderstephan:
Thanks for the feedback @jcmoraisjr 👍 I am currently working on it again. Sadly, we still have to get the server rename feature into the haproxy itself. This could still take some more time, but I will keep you updated!

alexanderstephan:
@jcmoraisjr I have seen that dynamic addition and deletion of servers is coming in the 17.x release. Wouldn't this make this change obsolete?

jcmoraisjr:
It depends, and a way to define it is e.g. your use case working fine with the dynamic addition/deletion. I have no objection on merging it, since it should continue as an option even after migrating the default approach on a future release.

alexanderstephan:
Submitted the server-rename to mailing list: https://www.mail-archive.com/haproxy@formilux.org/msg47211.html
Let's see what they say.

jcmoraisjr:
I can see the proposal on upstream [between these two commits](https://github.com/haproxy/haproxy/compare/7aec1e6d9ab2ccc11ce3c66610a8117cda2be094..af9bb8507dde45829f88e8878dcb204834a12094), congrats!

alexanderstephan:
@jcmoraisjr Thanks! I am also happy that my clear server counters PR got merged: https://github.com/haproxy/haproxy/commits/master/?author=alexanderstephan&since=2026-07-30&until=2026-07-30

It should get released within a few weeks from now. I will look into your feedback again now. Feel free to do another round of review then!

pkg/haproxy/dynupdate.go
alexanderstephan: True, I've dropped the rename from the join branch; that path now only re-enables the substituted endpoint as before.

pkg/haproxy/dynupdate.go
alexanderstephan: Agreed. I've moved the full rename sequence into execRenameEndpoint so all the orchestration lives in the `exec<...>Endpoint `family.

pkg/haproxy/dynupdate.go
alexanderstephan: Added a TODO at wantRename. 👍 I guess we cannot update the model for a dev release?

alexanderstephan:
Hi @jcmoraisjr, Could you take another look? The feature is released now. 🎉

pkg/converters/ingress/ingress.go
```diff
@@ -963,8 +963,10 @@ func (c *converter) addBackend(source *annotations.Source, pathLink *hatypes.Pat
 		switch mapper.Get(ingtypes.BackBackendServerNaming).Value {
 		case "ip":
 			backend.EpNaming = hatypes.EpIPPort
+			backend.ServerRename = true
 		case "pod":
 			backend.EpNaming = hatypes.EpTargetRef
+			backend.ServerRename = true
 		default:
 			backend.EpNaming = hatypes.EpSequence
 		}
```
jcmoraisjr, line 972: The functionality is being enabled unconditionally, leading to warnings for users not having a compatible HAProxy version, and not having an opt-out in case of an unexpected misbehavior.

Wondering if you can gate it behind a new `backend-server-rename` configuration key defaulting to `false`: it allows to enable the functionality without impacting current users.

```suggestion
		switch mapper.Get(ingtypes.BackBackendServerNaming).Value {
		case "ip":
			backend.EpNaming = hatypes.EpIPPort
			backend.ServerRename = mapper.Get(ingtypes.BackBackendServerRename).Bool()
		case "pod":
			backend.EpNaming = hatypes.EpTargetRef
			backend.ServerRename = mapper.Get(ingtypes.BackBackendServerRename).Bool()
		default:
			backend.EpNaming = hatypes.EpSequence
		}
```

pkg/haproxy/dynupdate.go
```diff
@@ -601,6 +618,97 @@ func (d *dynUpdater) execEnableServer(backname string, ep *hatypes.Endpoint) boo
 	return d.execCommandBackendServer(d.metrics.HAProxySetServerResponseTime, backname, ep, cmd, cmdSetServerState)
 }
 
+func (d *dynUpdater) execSetNameServer(backname, oldName, newName string) bool {
+	cmd := fmt.Sprintf("set server %s/%s name %s", backname, oldName, newName)
+	d.logger.InfoV(2, "api call: %s", cmd)
+	msgs, err := d.execCommand(d.metrics.HAProxySetServerResponseTime, []string{cmd})
```
jcmoraisjr: nit: any reason why `execCommandBackendServer()` wasn't used here? It seems to cover all the response pattern on a common way.

pkg/haproxy/dynupdate.go
```diff
@@ -601,6 +618,97 @@ func (d *dynUpdater) execEnableServer(backname string, ep *hatypes.Endpoint) boo
...
+// socket error, or an HAProxy without the command). The command first shipped
+// in HAProxy 3.5-dev4 (dev snapshot; not yet in a stable release), so on the
+// stable versions most users run this returns false and logs a warning;
+// callers treat that as a soft failure: the rename has already succeeded; the
+// only consequence is that counters remain accumulated from the previous slot
+// occupant until the next reload.
+func (d *dynUpdater) execClearCountersServer(backname, name string) bool {
+	cmd := fmt.Sprintf("clear counters server %s/%s force", backname, name)
+	d.logger.InfoV(2, "api call: %s", cmd)
+	msgs, err := d.execCommand(d.metrics.HAProxySetServerResponseTime, []string{cmd})
```
jcmoraisjr: nit: same as above, wondering if `execCommandBackendServer()` would be a better fit.

pkg/haproxy/dynupdate.go
```diff
@@ -601,6 +618,97 @@ func (d *dynUpdater) execEnableServer(backname string, ep *hatypes.Endpoint) boo
...
+	}
+	ep.Name = newName
+	d.logger.InfoV(2, "renamed server on backend '%s' from '%s' to '%s'", backname, oldName, newName)
+
+	// Reset counters while the server is still in maintenance mode so
+	// per-pod counter attribution starts cleanly for the new occupant.
+	// Failure is non-fatal by design (the helper logs the warning
+	// internally): the rename has already committed and dirty counters are
+	// preferable to failing the update entirely.
+	d.execClearCountersServer(backname, newName)
```
jcmoraisjr: It seems our linter is misconfigured, this should have failed in CI.


```suggestion
	_ = d.execClearCountersServer(backname, newName)
```

jcmoraisjr (commented):
The update looks sound, thanks again for the contribution.

I've just one ask about gate the functionality behind a new configuration key, what do you think about it? All the others are minor/nit.

I'm going to tag `alpha.3` next weekend, but take your time since we'll have another alpha just after it and we can wait for your feedback.

pkg/haproxy/dynupdate.go
alexanderstephan: Good catch, it was just an oversight on my end

pkg/converters/ingress/ingress.go
alexanderstephan: I also was annoyed by this warning, but I was not sure how to handle it. Will add the gate. 👍

alexanderstephan:
Hi @jcmoraisjr, Thanks for the review. I would be glad to make it to the next alpha. Agreed for the gating, I added it. I also addressed the nits and improved comments/docs a bit. Feel free to another look!

jcmoraisjr:
Thanks for the quick reply, sorry about taking so much time in the review queue. Looks good, merging now, should be part of the upcoming `alpha.3`.

alexanderstephan:
No worries. Thanks a lot! 👍  Do you have a rough estimate when v0.17 could become available for usage in production?

jcmoraisjr:
tl;dr: a few weeks from now, should be more than a month and probably less than two.

We're tagging `alpha.3` this week, probably having everything we want for v0.17. If so we should skip .4 and jump to beta releases. The plan is tagging once per other week, or even every week depending on the sort of the reports. About three of them should be enough to consider v0.17 GA.

We've however a number of tricky changes, the highlights are branding and the frontend refactor, so the pace of the pre-GA releases depends mostly on how stable the code continues to be after these refactors. We're currently missing lots of automated e2e, my own usage covers a fraction of the resources, so we need help from the community with tests on real scenarios. In case you can help us deploying v0.17 on staging and reporting issues, we should probably reach GA sooner than later. Btw the more you test on your own use case, the better GA tends to be for you.

alexanderstephan:
Okay, great to hear! I will do some testing with ingress controller in the next week and get back to you in case I have any findings. At least, I will run the server renaming for a few weeks in a test cluster.

PR #1482, opened 2026-05-20
Dynamic health check port updates

andrejbaran:
Hey @jcmoraisjr,

I came across a "bug" when `dynamic-scaling` is enabled and you switch `service-upstream` to true and the `service` port isn't the same as `targetPort` of the endpoint/pod.

The health-check would retain the old endpoint's port while the backend server entry would correctly be changed.

Let me know what you think about the fix, it's been a while since a digged through the the code so this might be rusty :)

### DIFF
--- a/pkg/haproxy/dynupdate.go
+++ b/pkg/haproxy/dynupdate.go
@@ -307,7 +307,7 @@ func (d *dynUpdater) dynamicallySyncSlots(pair *backendPair) bool {
 			// if cookie doesn't match here and preserving the value is
 			// important, don't even enable the endpoint before reloading
 			updated = false
-		} else if !d.execEnableEndpoint(curBack.ID, nil, added[i]) || added[i].Label != "" {
+		} else if !d.execEnableEndpoint(curBack, nil, added[i]) || added[i].Label != "" {
 			updated = false
 		}
 	}
@@ -392,7 +392,7 @@ func (d *dynUpdater) checkEndpointPair(backend *hatypes.Backend, pair *epPair) b
 	}
 	if reflect.DeepEqual(&oldEPCopy, pair.cur) {
 		// TODO revisit Label check, this is related with blue/green deployment
-		return d.execEnableEndpoint(backend.ID, pair.old, pair.cur) && pair.old.Label == "" && pair.cur.Label == ""
+		return d.execEnableEndpoint(backend, pair.old, pair.cur) && pair.old.Label == "" && pair.cur.Label == ""
 	}
 
 	// cannot handle via enablement, need to either delete+add or reload,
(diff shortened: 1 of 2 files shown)


pkg/haproxy/dynupdate.go
```diff
@@ -664,6 +690,8 @@ func cmdResponseOK(cmdcls cmdClass, response string) bool {
 		return response == "New server registered."
 	case cmdSetServerAddr:
 		return response == "nothing changed" || strings.HasPrefix(response, "IP changed from ") || strings.HasPrefix(response, "port changed from ") || strings.HasPrefix(response, "no need to change ")
+	case cmdSetServerCheckPort:
+		return response == "" || response == "nothing changed" || strings.HasPrefix(response, "check port changed from ") || strings.HasPrefix(response, "no need to change ")
```
jcmoraisjr, line 694: Does the API return empty response as well? If not my suggestion is to make it more strict by removing the empty check and configure the unit test instead to issue a custom and valid response.

pkg/haproxy/dynupdate_test.go
```diff
@@ -80,6 +80,43 @@ INFO-V(2) empty response from server
...
+			doconfig2: func(c *testConfig) {
+				b := c.config.Backends().AcquireBackend("default", "app", "8080")
+				b.Dynamic.DynScaling = types.DynScalingSlots
+				b.HealthCheck.Interval = "2s"
+				b.AcquireEndpoint("10.233.18.242", 80, "")
+			},
+			expected: []string{
+				"srv001:10.233.18.242:80:1",
+			},
+			dynamic: true,
```
jcmoraisjr, line 100: Nice test, wondering if you can add another one exploring an invalid response for the `set server check-port` API call

pkg/haproxy/dynupdate.go
```diff
@@ -509,6 +516,18 @@ func (d *dynUpdater) execEnableEndpoint(backname string, oldEP, curEP *hatypes.E
...
+	return oldEP != nil &&
+		oldEP.Port != curEP.Port &&
+		backend.HealthCheck.Port == 0 &&
+		backendHasHealthCheck(backend)
+}
+
+func backendHasHealthCheck(backend *hatypes.Backend) bool {
+	hc := backend.HealthCheck
+	return hc.Addr != "" || hc.FallCount != 0 || hc.Interval != "" || hc.RiseCount != 0
+}
```
jcmoraisjr, line 529: This is not a proper idiomatic way to test health check enabled, but I understand that it is the proper one for now since this is the way the template currently does.

https://github.com/jcmoraisjr/haproxy-ingress/blob/bc62f480405f4a225735505a7c84d1bc340b468c/rootfs/etc/templates/haproxy/haproxy.tmpl#L940

That said please add a quick comment about the missing of a flag and being the current template approach as a way to explain why this is being done this way. Adding `port!=0` shouldn't hurt as well, this better follows what the template does, although it is ignored when called from `shouldUpdateCheckPort()`.

jcmoraisjr (commented):
It looks good, adding just some changing suggestions. It followed a good approach, this is not rusty at all.

jcmoraisjr:
Hi @andrejbaran it seems I don't have permission to update your branch. Wondering if you can push [this commit](https://github.com/jcmoraisjr/haproxy-ingress/commit/a015c7fe69da967f7124fd7b4fae4a05bb871d14), I'll merge the changes just after that. Thanks!

PR #1486, opened 2026-05-28
fix: repopulate ingress status after leader re-acquisition

### LINKED ISSUE #1485: status.loadBalancer.ingress is permanently emptied after a leader-election loss and never repopulated when leadership is re-acquired

After a transient leader-election loss (e.g. a brief Kubernetes API / etcd hiccup that exceeds `RenewDeadline`), `status.loadBalancer.ingress` is cleared from every Ingress managed by the controller, and is **not** repopulated when the controller re-acquires the lease — even though leader election then runs continuously for days.

The cleared status persists across resync and only recovers when the controller pod is restarted. There is no error logged after the lease is re-acquired; from the operator's point of view the controller looks healthy but Ingress status is silently empty.

This breaks any downstream that consumes `status.loadBalancer.ingress` (external-dns, dashboards, etc.). In my case, ArgoCD sync is broken (continuously showing `Progressing`).

**Here is my first analysis:**

Root cause can be in `pkg/controller/services/svcaddress.go`. On lose-leadership, `svcAddress.Start` exits and calls `shutdown()`, which
calls `updateAllResources(nil)` and writes an empty status to every Ingress/Gateway. But it does **not** reset the in-memory cache `s.curr`.

`s.curr` still holds the LB list from the previous leadership, which equals readCurrentLB() (the node IP / publish address / publish service
hasn't actually changed). So the full-resync branch is skipped and only Ingresses currently in an event burst get their status restored. Anything stays empty until pod restart.

A single transient API-server slowness on a single-replica deployment / DaemonSet (one pod per node — the common topology) turns into a persistent Ingress-status outage.

**Expected behavior**

When the controller re-acquires the leader lease after a transient loss, `status.loadBalancer.ingress` should be repopulated on all managed Ingresses within one reconcile, without requiring a pod restart.

1. Run a single-replica controller writing ingress status. Confirm status.loadBalancer.ingress is populated
on Ingresses.
2. Force a brief leader-election loss while this is the sole replica 
3. Allow the controller to re-acquire the lease (happens automatically
within seconds).
4. every ingress statuses are empty and remains empty after the sync period
5. kubectl rollout restart the controller. Statuses repopulate within seconds.

**Environment information**

HAProxy Ingress version: v0.16.1

Command-line options (only flags relevant to the bug; the rest of our
flags don't affect it):

--controller-class=prod
--ingress-class=haproxy-prod
--report-node-internal-ip-address=true
--sync-period=10m
--reload-interval=10s
--track-old-instances=true
--update-status-on-shutdown=true

**Proposed patch**

Reset `s.curr` during shutdown, so the controller recovers on its own after leadership re-acquisition.

```go
func (s *svcAddress) shutdown() {
    s.log.Info("no other controller running, removing address from status")
    if err := s.updateAllResources(nil); err != nil {
            s.log.Error(err, "error updating resources")
    }
   s.curr = nil
 }
```

Until this is fixed, I `kubectl rollout restart` the controller as a workaround.

Happy to open the PR if the approach is acceptable.

arthlr:
On lose-leadership, shutdown() writes nil to every managed Ingress/Gateway status but does not reset `s.curr`. If leadership is re-acquired later, `s.curr` is equal to the freshly-read LB list and status stays empty until a pod restart.

My idea is to reset `s.curr`  so Ingress/Gateway statuses recover automatically on the next reconcile.

Fixes #1485

### DIFF
--- a/pkg/controller/services/svcaddress.go
+++ b/pkg/controller/services/svcaddress.go
@@ -216,6 +216,8 @@ func (s *svcAddress) shutdown() {
 	if err := s.updateAllResources(nil); err != nil {
 		s.log.Error(err, "error updating resources")
 	}
+	// Mirror the wipe so a later leadership re-acquisition repopulates status via checkChanged.
+	s.curr = nil
 }
 
 func (s *svcAddress) readCurrentLB(ctx context.Context) (lb []networking.IngressLoadBalancerIngress, err error) {


jcmoraisjr:
LGTM, thanks! Merging now.

PR #1525, opened 2026-09-02
Fix data race in TestHAProxyProcsLoop and enable -race flag for tests

Garfield96:
Add `-race` to the test target in the Makefile, and fix the data race it exposed in TestHAProxyProcsLoop by protecting `clientMock.cmdError` and `callCnt` with a mutex.


jcmoraisjr:
Looks good, thanks! Merging now.

PR #1526, opened 2026-09-02
Strip managed fields from cached objects

Garfield96:
To reduce memory consumption, managed fields (managed by API server) get removed from cached objects. Documentation: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#TransformStripManagedFields


jcmoraisjr:
We're committed in reducing memory consumption in the next cycle, this is a great first step on that direction. Thanks for suggesting it. Merging now.

PR #1535, opened 2026-09-16
Do not shutdown the controller's own connection when shedding sessions

### LINKED ISSUE #1534: Graceful session shedding aborts early — shutdownSessionsSync shuts down its own control socket

When `close-sessions-duration` is configured, `shutdownSessionsSync()` builds its shutdown list from every line of `show sess` output. That output includes the controller's *own* connection to the admin socket (`proto=unix_stream`), because the command is issued over the very socket being enumerated.

When the loop reaches that entry it shuts down its own control connection. The next `shutdown session` write fails, and the function returns — silently abandoning every session still left in the list.

Those abandoned sessions are not shed gracefully. They stay open until `timeout-stop`, at which point HAProxy logs `soft-stop running for too long, performing a hard-stop` and closes all of them in the same millisecond. So a feature designed to spread session closure over `close-sessions-duration` instead produces a partial paced phase followed by an abrupt mass close — the exact behaviour it exists to prevent.

The connection cannot recover, by design:

- [`Unlistening()`](https://github.com/n42-gateway/n42-gateway/blob/master/pkg/haproxy/socket/socket.go#L127-L134) dials the socket (pinning it to the instance being tracked) and then sets `s.listening = false`
- the write-retry at [socket.go:162](https://github.com/n42-gateway/n42-gateway/blob/master/pkg/haproxy/socket/socket.go#L162) is gated on `s.keepalive && s.listening`, so it is skipped
- [`acquireConn()`](https://github.com/n42-gateway/n42-gateway/blob/master/pkg/haproxy/socket/socket.go#L204) then refuses to re-dial: `cannot connect to '%s': listening is down`

The relevant loop is [connections.go:116-122](https://github.com/n42-gateway/n42-gateway/blob/master/pkg/haproxy/connections.go#L116-L122):

```go
var sessionList []string
for _, s := range strings.Split(sess[0], "\n") {
	i := strings.Index(s, ":")
	if i > 0 {
		sessionList = append(sessionList, s[:i])
	}
}
```

Note the doc comment immediately above it shows `proto=unix_stream` lines as the example output being parsed.

The error return at [connections.go:129-133](https://github.com/n42-gateway/n42-gateway/blob/master/pkg/haproxy/connections.go#L129-L133) carries the comment *"maybe the connection or the instance is gone, haproxy takes care of the remaining sessions if any"*. That is true in the sense that HAProxy does close them — but it closes them all at once at `hard-stop-after`, which defeats the pacing.

**Expected behavior**

All eligible sessions are shed at the interval implied by `close-sessions-duration`, and the controller does not include its own control connection in the list it shuts down.

1. Configure `close-sessions-duration` smaller than `timeout-stop` (e.g. `8m` / `10m`) with `--track-old-instances=true`
2. Establish a number of long-lived connections through the proxy (TCP mode / tunnelled connections make this easiest to observe)
3. Trigger a config reload
4. Observe session closures: a paced run at `close-sessions-duration / (N+1)` intervals starting at `timeout-stop - close-sessions-duration`, which stops partway; the remainder all close together at `timeout-stop`, with `soft-stop running for too long, performing a hard-stop` in the log

**Observed behaviour**

The fraction shed gracefully is determined by where HAProxy happens to list the controller's own connection in `show sess`, which is not deterministic and is not configurable. Measured on a single reload across four identically configured instances — same config, same reload, comparable session counts — the paced fraction ranged from **30.8% to 54.5%**. Across many reloads the observed range was **18.5% to 99.8%**.

The controller's own socket is reliably present, not an edge case. Sampling `show sess` on every worker of one instance ten times over three minutes, **every sample contained between 3 and 5 such sockets** — one per tracked old instance still inside its shed window. Each is identifiable as `proto=unix_stream src=unix:1 fe=GLOBAL be=<NONE> calls=1`.

This also means the paced interval is computed from an inflated `len(sessionList)`, though that error is negligible.

**Suggested fix**

Skip `unix_stream` sessions when building the list:

```go
for _, s := range strings.Split(sess[0], "\n") {
	i := strings.Index(s, ":")
	if i > 0 && !strings.Contains(s, "proto=unix_stream") {
		sessionList = append(sessionList, s[:i])
	}
}
```

Happy to open a PR for this if you'd like.

**Environment information**

Observed on `v0.14.7`. The code in `shutdownSessionsSync()` and the `listening` handling in `socket.go` are unchanged on `master` as of `c8fc12d`, so this appears to affect all versions where `close-sessions-duration` is available.

Command-line options:

```
--track-old-instances=true
```

Global options:

```
close-sessions-duration: 8m
timeout-stop: 10m
```

Ingress objects: not relevant — reproduces with any Ingress set. The trigger is a config reload with established sessions present.

pranaybm:
Fixes #1534.

`shutdownSessionsSync()` collects every line of `show sess` into its shutdown list, including the controller's own connection to the admin socket. When the loop reaches that entry it closes its own control connection; the next write fails and the function returns, leaving the remaining sessions to be closed together at `timeout-stop` instead of paced over `close-sessions-duration`.

This skips `proto=unix_stream` entries when building the list.

On backports — the affected branches are `release-0.14`, `release-0.15` and `release-0.16`. `pkg/haproxy/connections.go` does not exist on `release-0.13`, so v0.13 is unaffected and needs nothing.

`go build`, `go vet`, `gofmt` and `go test ./pkg/haproxy/...` all pass.


pranaybm:
@jcmoraisjr Gentle ping — the workflows on this need approval to run, since it's my first contribution here. Happy to add anything you'd like before then; otherwise it's just the one-line filter from #1534.

pkg/haproxy/connections.go
```diff
@@ -116,7 +116,9 @@ func shutdownSessionsSync(sock socket.HAProxySocket, duration time.Duration) {
 	var sessionList []string
 	for _, s := range strings.Split(sess[0], "\n") {
 		i := strings.Index(s, ":")
-		if i > 0 {
+		// skip our own connection to the admin socket, otherwise the loop
+		// below shuts it down and cannot reach the remaining sessions
+		if i > 0 && !strings.Contains(s, "proto=unix_stream") {
```
jcmoraisjr, line 121: The connection cannot be compared that way, it'd ignore connections when ssl-passthrough is enabled. Check if the following change works for you:


```suggestion
		if i > 0 && !strings.Contains(s, "fe=GLOBAL") {
```

jcmoraisjr (commented):
Good catch, thanks for sharing. One changing suggestion, this should make your update work when ssl-passthrough is enabled.

pkg/haproxy/connections.go
```diff
@@ -116,7 +116,9 @@ func shutdownSessionsSync(sock socket.HAProxySocket, duration time.Duration) {
 	var sessionList []string
 	for _, s := range strings.Split(sess[0], "\n") {
 		i := strings.Index(s, ":")
-		if i > 0 {
+		// skip our own connection to the admin socket, otherwise the loop
```
jcmoraisjr, line 119: quick update, check-spelling should also complain here.


```suggestion
		// skip our own connection to the admin socket; otherwise the loop
```

